PR #1101, opened 2014-11-16
Rule engine v 0.1 - a basic backward chainer

misgeatgit:
running it on scheme shell example
guile> (load-scm-from-file "$OC_DIR/opencog/python/pln/examples/backward_chaining/criminal.scm")
gulie> (cog-bc  (InheritanceLink ( VariableNode "$isCrimina") (ConceptNode "criminal")))

(ListLink
   (ListLink
      (VariableNode "$isCriminal")
      (ConceptNode "West")
   )
)

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/InferenceSCM.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/InferenceSCM.cc
@@ -25,6 +25,7 @@
 #include <opencog/guile/SchemePrimitive.h>
 #include <opencog/guile/SchemeSmob.h>
 #include <opencog/reasoning/RuleEngine/rule-engine-src/pln/ForwardChainer.h>
+#include <opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.h>
 #include <opencog/atomspace/AtomSpace.h>
 
 using namespace opencog;
@@ -45,8 +46,11 @@ InferenceSCM::~InferenceSCM() {
 void InferenceSCM::init(void) {
 	_inst = new InferenceSCM();
 #ifdef HAVE_GUILE
-    //all commands for invoking the rule engine from scm shell should be declared here
-	define_scheme_primitive("cog-fc", &InferenceSCM::do_forward_chaining, _inst); //eg. from scm shell (cog-fc (InheritanceLink (ConceptNode "cat")(ConceptNode "animal"))
+	//all commands for invoking the rule engine from scm shell should be declared here
+	define_scheme_primitive("cog-fc", &InferenceSCM::do_forward_chaining,
+			_inst); //eg. from scm shell (cog-fc (InheritanceLink (ConceptNode "cat")(ConceptNode "animal"))
+	define_scheme_primitive("cog-bc", &InferenceSCM::do_backward_chaining,
+			_inst); //backward chaining
 #endif
 }
 
@@ -55,8 +59,28 @@ Handle InferenceSCM::do_forward_chaining(Handle h) {
 	AtomSpace *as = SchemeSmob::ss_get_env_as("cog-fc");
 	ForwardChainer fc(as);
 	fc.do_chain(h); //START FORWARD CHAINING
-    HandleSeq result = fc.get_chaining_result();
-    return as->addLink(LIST_LINK,result,TruthValue::DEFAULT_TV());
+	HandleSeq result = fc.get_chaining_result();
+	return as->addLink(LIST_LINK, result, TruthValue::DEFAULT_TV());
+#else
+	return Handle::UNDEFINED;
+#endif
+}
+
+Handle InferenceSCM::do_backward_chaining(Handle h) {
+#ifdef HAVE_GUILE
+	AtomSpace *as = SchemeSmob::ss_get_env_as("cog-bc");
+	BackwardChainer bc(as);
+	bc.do_chain(h);
+	map<Handle, HandleSeq> soln = bc.get_chaining_result();
+	HandleSeq soln_list_link;
+	for (auto it = soln.begin(); it != soln.end(); ++it) {
+		HandleSeq hs;
+		hs.push_back(it->first);
+		hs.insert(hs.end(), it->second.begin(), it->second.end());
+		soln_list_link.push_back(
+				as->addLink(LIST_LINK, hs, TruthValue::DEFAULT_TV()));
+	}
+	return as->addLink(LIST_LINK, soln_list_link, TruthValue::DEFAULT_TV());
 #else
 	return Handle::UNDEFINED;
 #endif
--- a/opencog/reasoning/RuleEngine/rule-engine-src/README.md
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/README.md
@@ -1,0 +1,169 @@
+Unified rule engine
+-------------------
+
+**Introduction**
+
+   The unified rule engine project aims at building a generic opencog rule engine on top of the pattern matcher with a c++ PLN
+ implementation where rules are put in a scheme representation. This will enable the reuse PLN for any kind of Backward
+ and Forward chaining inferences as long as rules are represented as an ImplicationLinks or BindLinks which are decoupled  from 
+ the PLN inference engine and are loaded dynamically from a scheme file.In the new design of the rule engine, PLN will use the Pattern matcher API
+ for querying the atomspace which will also guarantees code reuse and no reinvention of the wheel.The pattern matcher can be invoked implicitly with the default
+ callbacks or explicitly using custom callback handlers for specialized cases.All the above criteria and other issues has required a new implementation of PLN. 
+
+**Overall requirement/objectives**
+  
+  1.Rules should be specified as ImplicationLinks or BindLinks in a scheme file.  
+  2.Mechanisms will be provided for backward chaining from a given target, or forward chaining from a set of premises.  
+  3.The control mechanism, i.e. the policy for choosing which rules to apply in a given case,
+   must be pluggable. That is, when you invoke the rule engine, you get to choose which of the
+   available control policies will be used.   
+  4.There should be a standard way to express exclusivity between rules,and priority levels of rules --- but both of
+   these things should be relative to a given control policy (i.e rule X might have higher priority than rule Y with 
+   respect to policy A, but not policy B)   
+  5.The rule engine should be associated with some way of keeping track of which rules have been applied to which Atoms.
+    This information may be used by some control policies.   
+  6.Use pattern matcher for finding groundings and implement the callbacks if the need arises.
+    
+  further reading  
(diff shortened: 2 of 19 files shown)


AmeBel:
@linas or @cosmoharrigan  it would be great if you would review this changes. Thanks

opencog/reasoning/RuleEngine/rule-engine-src/InferenceSCM.cc
```diff
@@ -55,8 +59,28 @@ Handle InferenceSCM::do_forward_chaining(Handle h) {
 	AtomSpace *as = SchemeSmob::ss_get_env_as("cog-fc");
 	ForwardChainer fc(as);
 	fc.do_chain(h); //START FORWARD CHAINING
-    HandleSeq result = fc.get_chaining_result();
-    return as->addLink(LIST_LINK,result,TruthValue::DEFAULT_TV());
+	HandleSeq result = fc.get_chaining_result();
+	return as->addLink(LIST_LINK, result, TruthValue::DEFAULT_TV());
```
linas: trivial comment: you don;t have to say "TruthValue::DEFAULT_TV()" because if you don't specify it, it will become the default anyway...

opencog/reasoning/RuleEngine/rule-engine-src/README.md
```diff
@@ -0,0 +1,169 @@
+Unified rule engine
+-------------------
+
+**Introduction**
+
+   The unified rule engine project aims at building a generic opencog rule engine on top of the pattern matcher with a c++ PLN
```
linas: another minor comment: wrapping long lines to be 72 chars wide make this more readable in raw-text form.

opencog/reasoning/RuleEngine/rule-engine-src/README.md
```diff
@@ -0,0 +1,169 @@
...
+					hcurrent_target = choose_target_from_atomspace(main_atom_space); //start FC on a random target
+				else
+					hcurrent_target = htarget;
+	    	else 
+				if (!target_list_.empty())
+					hcurrent_target = choose_target_from_list(target_list_)
+			choose_input(hcurrent_target); //add more premise via pattern matching of related atoms to hcurrent_target
+			choose_rule()
+			patternmatch(hcurrent_chosen_rule) // call the pattern matching with the chosen rule wrapped in a BindLink
+			steps++
```
linas: I think you are missing an important step here: how is the target_list updated, after  the pattern matcher has run?

linas: Oh, and another important case: what if the choosen rule yielded no results: i.e. it simply can't be applied to the current target?

opencog/reasoning/RuleEngine/rule-engine-src/README.md
```diff
@@ -0,0 +1,169 @@
...
+ 
+Below is a kind of pseudo code of the do_bc function
+
+	do_chain(htarget)
+		hcurrent_target	
+		steps = 0
+		while (steps <= ITERATION_SIZE /*or !terminate*/) 
+			if steps == 0
+				if htarget == Handle::UNDEFINED
+					hcurrent_target = choose_target_from_atomspace(main_atom_space); //start FC on a random target
```
linas: Don't do this. Just throw an error instead. Let some higher layer decide what to do here.

opencog/reasoning/RuleEngine/rule-engine-src/README.md
```diff
@@ -0,0 +1,169 @@
...
+			patternmatch(hcurrent_chosen_rule) // call the pattern matching with the chosen rule wrapped in a BindLink
+			steps++
+	
+ *Implementing fitness based rule choosing:For the sake of having a prototype implementation other parts of the algorithms, I have implemented the choose_rule as a random selector. 
+ 
+ *The termination criteria: right now it just does some constant iterations specified in the config file which should be changed to appropriate criteria. 
+ 
+ *The control policy: I did a basic rule loading from a config file as a prototype. But I am thinking of how to declare conditional rule applications. An example usage scenario is
+ 
+ the relex to logic SV and SVO rules. the detail is found [here](http://wiki.opencog.org/w/RelEx2Logic_Rules#Suggested_Rule_File_Format).
```
linas: Not sure if you are following the conversation in another thread, but I'm advocating a minor change to the file format, and a major change to how one thinks about the application of rules.  In the short-term, this mostly doesn't affect you, because the "context+forking" mutual-disjunction ideas there are equivalent to randomly choosing different rules.

opencog/reasoning/RuleEngine/rule-engine-src/pln/BCPatternMatch.cc
```diff
@@ -0,0 +1,65 @@
...
+
+}
+
+BCPatternMatch::~BCPatternMatch() {
+
+}
+
+bool BCPatternMatch::node_match(Handle& node1, Handle& node2) {
+	if (AttentionalFocusCB::node_match(node1, node2))
+		return false;
```
linas: Did you mean to reverse true and false here?? Just write this instead:
```
return  AttentionalFocusCB::node_match(node1, node2);
```

opencog/reasoning/RuleEngine/rule-engine-src/pln/BCPatternMatch.cc
```diff
@@ -0,0 +1,65 @@
...
+	else
+		return true;
+
+}
+bool BCPatternMatch::link_match(LinkPtr& lpat, LinkPtr& lsoln) {
```
linas: true and false are reverse here too .. Just write `return AttentionalFocusCB::link_match(lpat, lsoln));`

opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
```diff
@@ -0,0 +1,492 @@
...
+	for (Handle h : result) {
+		HandleSeq var_containing = commons_->get_nodes(h, vector<Type> {
+				VARIABLE_NODE });
+		if (var_containing.empty())
+			grounded.push_back(h);
+	}
+	return grounded;
+}
+
+HandleSeq BackwardChainer::filter_rules(HandleSeq result) {
```
linas: Each method should be documented ... explain what it does, why it's needed.

opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
```diff
@@ -0,0 +1,492 @@
...
+	HandleSeq rules;
+	for (Handle h : result) {
+		HandleSeq links = as_->getIncoming(h);
+		for (Handle l : links) {
+			Type t = as_->getType(l);
+			if (t == LIST_LINK) {
+				HandleSeq incoming = as_->getIncoming(l);
+				if (not incoming.empty()) {
+					if (as_->getType(incoming[0]) == EXECUTION_LINK) {
+						HandleSeq hs = as_->getIncoming(incoming[0]);
```
linas: There's a whole lot of if-then-else going on here ... it needs to be explained.

opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
```diff
@@ -0,0 +1,492 @@
...
+map<Handle, HandleSeq> BackwardChainer::unify_to_empty_set(Handle&htarget) {
+	HandleSeq vars = commons_->get_nodes(htarget,
+			vector<Type> { VARIABLE_NODE });
+	map<Handle, HandleSeq> result;
+	for (Handle h : vars)
+		result[h] = HandleSeq { Handle::UNDEFINED };
+	return result;
+}
+
+map<Handle, HandleSeq> BackwardChainer::unify(Handle& htarget, Handle& match,
```
linas: The name of  this method, and the last few methods, is confusing to me.  The pattern matcher is a unifier.  That is what it does: it unifies. Why do you need to create a new, different kind of unifier?

opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.h
```diff
@@ -50,42 +54,177 @@ using namespace opencog;
...
+	Handle get_root_logical_link(Handle hrule)
+			throw (opencog::InvalidParamException);
+	/**
+	 * find and return all Handles with no variables
+	 * @param handles
+	 * @return vectors of handles with no variables in them
+	 */
+	HandleSeq get_grounded(HandleSeq handles);
+	/**
+	 * Find and return all handles containing variable and are within an implicationLink
```
linas: FYI, there are handy tools for finding variables and nodes and things like that, here:  opencog/query/PatternUtils.h

opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
```diff
@@ -0,0 +1,194 @@
...
+Handle PLNCommons::create_bindLink(Handle himplicant, bool vnode_is_typedv)
+		throw (opencog::InvalidParamException) {
+	if (!LinkCast(himplicant)) {
+		throw InvalidParamException(TRACE_INFO, "Input must be a link type ");
+	} //xxx why?
+
+	//if(vnode_is_typedv)
+	himplicant = create_with_unique_var(himplicant);
+
+	HandleSeq variable_nodes = get_nodes(himplicant, vector<Type> {
```
linas: I almost hate to mention this, as perhaps its low priority, but .. the pattern matcher does have lower-level API's than bind-link.   If you already know what the variable are, and what the clauses are, you can pass them directly to the pattern matcher.

When you give the pattern matcher a bind link, the very first thing it does is to extract the variables, and reorganize everything into C++ HandleSeq's. So, if you already have handleSeq's, there is not much point to pack them into a bind-link, which the pattern matcher must then unpack. 

This is low priority, though, because there is no harm to the pack/unpack step, and it does give the pattern matcher a chance to look for badly-formed patterns, and throw errors in such cases ...

linas:
I merged, because I saw no point of waiting any longer.  I still have not thought deeply about what the backward-chainer code is doing.  The pseudocode for the backwards-chainer is still to dense, complex for me to really understand it.  A lot of the C++ code is still missing an explanation.

opencog/reasoning/RuleEngine/rule-engine-src/pln/BCPatternMatch.cc
williampma: Caution: most of these changes pre-date the recent reversal of Pattern Matcher's callback value, so likely it should be re-reversed?

misgeatgit: @williampma thanks, yes we should reverse it.

PR #1136, opened 2014-12-05
stage one of adding r2l preposition rules

anitzkin:
Tested these. They seem to work.  Of course not until the commits I just made to the sem-algs and r2l rules are merged also . . .

### DIFF
--- a/opencog/nlp/relex2logic/rule-helpers.scm
+++ b/opencog/nlp/relex2logic/rule-helpers.scm
@@ -414,30 +414,31 @@
 )
 ; -----------------------------------------------------------------------
 ; Adjective and adverb rules
+; (advmod is also called for adverbials, such as some prepositional phrases 
 ; -----------------------------------------------------------------------
 (define (amod-rule concept instance adj adj_instance)
 	(list (InheritanceLink  (ConceptNode adj_instance df-node-stv) (ConceptNode adj df-node-stv) df-link-stv)
-	(InheritanceLink  (ConceptNode instance df-node-stv) (ConceptNode adj_instance df-node-stv) df-link-stv)
 	(InheritanceLink  (ConceptNode instance df-node-stv) (ConceptNode concept df-node-stv) df-link-stv)
-	)
+	(InheritanceLink  (ConceptNode instance df-node-stv) (ConceptNode adj_instance df-node-stv) df-link-stv))
 )
 
 (define (advmod-rule verb instance adv adv_instance)
 	(list (InheritanceLink  (ConceptNode adv_instance df-node-stv) (ConceptNode adv df-node-stv) df-link-stv)
-	(InheritanceLink  (SatisfyingSetLink (PredicateNode instance df-node-stv) df-link-stv) (ConceptNode adv_instance df-node-stv) df-link-stv)
 	(ImplicationLink  (PredicateNode instance df-node-stv) (PredicateNode verb df-node-stv) df-link-stv)
+	(InheritanceLink  (SatisfyingSetLink (PredicateNode instance df-node-stv) df-link-stv) (ConceptNode adv_instance df-node-stv) df-link-stv)
 	)
 )
-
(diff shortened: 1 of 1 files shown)


opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -1056,6 +1110,74 @@
...
+		)
+	)
+)
+
+(define (attime-rule comp_concept comp_instance pred_concept pred_instance)
+	(list 
+		(ImplicationLink (PredicateNode comp_instance df-node-stv) (PredicateNode comp_concept df-node-stv) df-link-stv)
+		(ImplicationLink (PredicateNode pred_instance df-node-stv) (PredicateNode pred_concept df-node-stv) df-link-stv)
+		(AtTimeLink df-link-stv)
+			(PredicateNode pred_instance df-node-stv)
```
AmeBel: shouldn't there be a TimeNode? http://wiki.opencog.org/w/TimeServer#Temporal_association_in_AtomSpace_hypergraph_and_TimeServer

opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -414,30 +414,35 @@
...
+; prepositional phrase rule
+;-----------------------------------------------------------------------
+(define (pp-rule prep_concept prep_instance noun_concept noun_instance)
+	(list (ImplicationLink (PredicateNode prep_instance df-node-stv) (PredicateNode prep_concept df-node-stv) df-link-stv)
+	(InheritanceLink (ConceptNode noun_instance df-node-stv) (ConceptNode noun_concept df-node-stv) df-link-stv)
+	(EvaluationLink df-link-stv
+		(PredicateNode prep_instance df-node-stv)
+		(ListLink (ConceptNode noun_instance df-node-stv))
+	))
+)
```
williampma: The output is this pp-rule seems incomplete.  For example, it will generate
```
(EvaluationLink
  (PredicateNode "on@11")
  (ListLink
    (ConceptNode "table@22")
  )
)
```
without saying what is "on-table".  However, I noticed you are using the "SVO-rule" to generate the full link (it-on-table) as well, so maybe the pp-rule is not needed.

Probably should rename SVO-rule now that it is used for so many purpose...

opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -1056,6 +1110,74 @@
 		)
 	))])
 )
+
+;-----------------------------------------------------------------------
+; complement clauses
+;-----------------------------------------------------------------------
```
AmeBel: some example here

opencog/nlp/relex2logic/rule-helpers.scm
anitzkin: Yes, in general, a lot of the changes I have made make the construction of the logical form more modular.  This seems to me to make sense, because for example, a prepositional phrase may serve the role of an adjective (an adjectival), as in "Look at the book on the table," an adverb (an adverbial), as in "Sit on the table, please," or an intensional complement, as in "the book is on the table."  And likewise, many things other than prepositional phrases can play each of these roles, so rather than having a separate rule for each kind of phrase in each role, it makes more sense to just have a rule for each kind of phrase, and a rule for each role, and then put those together -- fewer rules, greater flexibility.

As for SVO, need to look at that again.  Treating "The book is on the table" as SVO is not technically corect, from the point of view of linguistics, but I may have instantiated it that way in order to adapt to the limitations of relex output at the time i did those revisions. We can probably do better.

opencog/nlp/relex2logic/rule-helpers.scm
anitzkin: Ya think? Yeah, I would think so.

AmeBel:
Merged because the AtTimeLink issue @ https://github.com/opencog/opencog/pull/1136#discussion_r24643058 requires more thought. Works for now.

Issue #1147, opened 2014-12-17
REST module can't get an updated content of the atomspace


misgeatgit:
If one starts the REST API and does http://localhost:5000/api/v1.1/atoms request,  The REST server returns the  content of the atomspace. However, any later modification made to the atomspace after starting the REST server is not being reflected in the subsequent http://localhost:5000/api/v1.1/atoms requests.

Note that the issue is generic to any other python modules and not particular to the REST server.Here https://github.com/misgeatgit/opencog/blob/restapi-bug-reproduce/opencog/python/web/api/restapi_bug_sscce.py is a simple python module with an http server to reproduce the bug without involving the REST API.You can use it by adding restapi_bug_sscce in your PYTHON_PRELOAD param in opencog.conf. or pulling https://github.com/misgeatgit/opencog/tree/restapi-bug-reproduce.

misgeatgit:
Sorry, the problem was actually particular to the REST implementation. process forking seems to copy the atomspace. I have made a pull request to fix this problem but It recreates the problem addressed by commit 36d9c895353f71e9e50b9ad6ebae9e1c2f8eda9e.

cosmoharrigan:
@misgeatgit What do you think happens during creation of a daemonic child process that causes that symptom, which doesn't occur using a thread? Is there anything about how the atomspace reference is passed in atomspace_details.pyx that you mention above that could be modified so that the daemonic child process would receive a reference to the correct atomspace?

misgeatgit:
@cosmoharrigan  According to the discussion http://stackoverflow.com/questions/10721915/shared-memory-objects-in-python-multiprocessing supports my guess. The solution would be making the atomspace a shared object between the python processes(https://docs.python.org/2/library/multiprocessing.html#sharing-state-between-processes). I will do that.

bgoertzel:
Ah... that sounds promising ...

On Sun, Jan 4, 2015 at 10:39 AM, Misgana Bayetta <notifications@github.com>
wrote:

> @cosmoharrigan <https://github.com/cosmoharrigan> According to the
> discussion
> http://stackoverflow.com/questions/10721915/shared-memory-objects-in-python-multiprocessing
> supports my guess. The solution would be making the atomspace a shared
> object between the python processes(
> https://docs.python.org/2/library/multiprocessing.html#sharing-state-between-processes).
> I will do that.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #1176, opened 2015-01-03
A  fix to issue 1147

misgeatgit:
The fix applied here used to exisit before commit
36d9c895353f71e9e50b9ad6ebae9e1c2f8eda9e
Unfortunately this commit will resurrect the problem solved by the
above commit.

### DIFF
--- a/opencog/python/web/api/restapi.py
+++ b/opencog/python/web/api/restapi.py
@@ -2,7 +2,7 @@ __author__ = 'Cosmo Harrigan'
 
 import opencog.cogserver
 from web.api.apimain import RESTAPI
-from multiprocessing import Process
+from threading import Thread
 
 # Endpoint configuration
 # To allow public access, set to 0.0.0.0; for local access, set to 127.0.0.1
@@ -40,7 +40,6 @@ class Start(opencog.cogserver.Request):
                   "http://127.0.0.1:5000/api/v1.1/atoms?type=ConceptNode"
 
     def __init__(self):
-        self.process = Process(target=self.invoke)
         self.atomspace = None  # Will be passed as argument in run method
 
     def run(self, args, atomspace):
@@ -57,10 +56,10 @@ class Start(opencog.cogserver.Request):
         parent process exits, it will attempt to terminate the daemonic
         child process (https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Process.daemon)
         '''
-        self.process.daemon = True
-        self.process.start()
-
-        print "REST API is now running in a separate process."
+        thread = Thread(target=self.invoke)
+        thread.start()
+        print "REST API is now running in a separate thread."
+        # @todo: detect Control-C to end the thread
 
     def invoke(self):
         self.api = RESTAPI(self.atomspace)


cosmoharrigan:
See comment https://github.com/opencog/opencog/issues/1147#issuecomment-68608410

cosmoharrigan:
@misgeatgit It's great that you have identified the use of multiprocessing as being involved here. But, as you alluded to above, I don't think reverting to this prior usage of threading is a good solution, because using threading instead of multiprocessing causes other very unpleasant issues. Instead I think it would be preferable to use the knowledge you have discovered, to try to figure out how to access the shared atomspace, while using multiprocessing.

linas:
Cosmo:  What are the threading issues?  I believe that it is fundamentally impossible to share the atomspace via shared memory.  It is making use of a large and sophisticated infrastructure inside of glibc that supports stack unwinding, local storage, stack variables, signal handling. You are not going to be able to replicate that without becoming a glibc expert, and copying large parts of that code. (I'm expounding on this, because, in one if my previous jobs, I added support for a new CPU to glibc. There is a huge wealth of complex machinery in there, that most people are quite oblivious to.  That machinery works with threads. It doe not work at all with IPC)  

Think of it this way: just think of what it would take to create a shared mem allocator!  You'd have to reproduce all of malloc!  The pool management, the COW pages, the linux kernel system calls.  And debug all of that. People have tried this in the past, and you should be able to find some package somewhere, even on github, that does this, but it will be shaky and buggy.  IPC is hard.

In addition, the atomspace also uses boost reference counting for pointers, and so you will have to figure out how to share that state across processes as well. This stuff is deep in the bowels of boost.  So you'd have to reverse-engineer boost, and figure out how to do that across IPC.   

So that's just for C+. But you are using python...   I don't think even the most serious python maintainers would know how to do that. Just read the comments in the stack-exchange articles quoted in the original bug report to get a flavor f it.

Finally: go look at the history of threading: it was invented to solve these serious and fundamental IPC issues.   Threading is just IPC without the bugs of IPC.  That's the whole point of threading. To try to go back to IPC, and try to wrestle with these deep and fundamental comp-sci issues is  ... this is not the project to do this.

bgoertzel:
Well if the issue is just respect of the CogServer "shutdown" command,
there must be some way to solve that consistent with use of threading

Or perhaps there is a miscommunication here and Cosmo means something
different by "multiprocessing" than Linas thinks he does?

ben

On Mon, Jan 5, 2015 at 8:13 AM, Linas Vepštas <notifications@github.com>
wrote:

> Cosmo: What are the threading issues? I believe that it is fundamentally
> impossible to share the atomspace via shared memory. It is making use of a
> large and sophisticated infrastructure inside of glibc that supports stack
> unwinding, local storage, stack variables, signal handling. You are not
> going to be able to replicate that without becoming a glibc expert, and
> copying large parts of that code. (I'm expounding on this, because, in one
> if my previous jobs, I added support for a new CPU to glibc. There is a
> huge wealth of complex machinery in there, that most people are quite
> oblivious to. That machinery works with threads. It doe not work at all
> with IPC)
>
> Think of it this way: just think of what it would take to create a shared
> mem allocator! You'd have to reproduce all of malloc! The pool management,
> the COW pages, the linux kernel system calls. And debug all of that. People
> have tried this in the past, and you should be able to find some package
> somewhere, even on github, that does this, but it will be shaky and buggy.
> IPC is hard.
>
> In addition, the atomspace also uses boost reference counting for
> pointers, and so you will have to figure out how to share that state across
> processes as well. This stuff is deep in the bowels of boost. So you'd have
> to reverse-engineer boost, and figure out how to do that across IPC.
>
> So that's just for C+. But you are using python... I don't think even the
> most serious python maintainers would know how to do that. Just read the
> comments in the stack-exchange articles quoted in the original bug report
> to get a flavor f it.
>
> Finally: go look at the history of threading: it was invented to solve
> these serious and fundamental IPC issues. Threading is just IPC without the
> bugs of IPC. That's the whole point of threading. To try to go back to IPC,
> and try to wrestle with these deep and fundamental comp-sci issues is ...
> this is not the project to do this.
>
> —
> <https://github.com/opencog/opencog/pull/1176#issuecomment-68656279>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

cosmoharrigan:
By "multiprocessing" I am referring to https://docs.python.org/2/library/multiprocessing.html.

> What are the threading issues?

The issue was that the thread would not exit when the cogserver exited; as a result, it would still be running in the background, and if one tried to restart the cogserver in the same shell window, one would receive an error unless the process was manually terminated. This became a particularly important issue when using the [python-client](https://github.com/opencog/python-client) external library to run experiments that required repeatedly starting and stopping the cogserver.

@misgeatgit Maybe you can try making the REST API thread into a daemon thread?
http://stackoverflow.com/questions/190010/daemon-threads-explanation

@linas Alright. I'll merge this in the meantime.

Issue #309, opened 2013-09-24
MathUTest fails with gcc-4.7


linas:
MathUTest  pass/fail seems to depend on some very subtle float-point rounding issues; a minor change in rounding causes the test to blow up.  I debugged a little and here's what I found:

rectangles are added and removed from SuperEntity, and then the overall geometry is recomputed in SuperEntity::rebuild()  This uses the function rectangle.isInside() repeatedly to test if geometries intersect. 

The problem seems to be that the geometries neither intersect, nor don't intersect: they just touch. So the tiniest shift in geometry, a change in the last bit of a floating-pint number, will cause the Rectangle::isInside() to go from true to false.  This changes how SuperEntity::rebuild() responds, and thus, causes MathUTest to bomb out.

Figuring out how to fix this will require understanding what SuperEntity is and what it does ...

brett79:
SuperEntity provides behaviour to gather the outside 2D perimeter of a number of overlapping entities. Any internal edges are 'hidden'.This could be useful if a lot of objects are overlapping and you wish to navigate around them. The MathUTest was checking for the edges to be in a specific order. Order is not enforced within the existing code, and entities are stored in a non-ordered hash map. Suggested fix is to check for the correct edges, and ignore the order they are presented.
While reviewing the code, I noticed an extra edge being deleted due to some midpoint checking, fix should be to check either end of the edge.

linas:
Hi, Thanks for the patch; I merged because it looked reasonable. If you understand, even partly, what this code does, could you added an explanation, either to the .cc or the .h file, or possibly create a wiki page for it?  My general impression is that everything in the 'spatial' directory is under-documented, and no one has a very clear idea of what it is or what it does.  (Well, I don't)

AmeBel:
The test is passing,
buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/70/steps/test/logs/stdio

PR #1177, opened 2015-01-04
Changes for SuperEntity Testing. https://github.com/opencog/opencog/issues/309

brett79:
Edges of the SuperEntity are currently subject to an unsorted list. Updated test_SuperEntity() to pre-sort edges prior to testing, added edge count assertions. Updated splitEdges() to only delete edges wholely within another entity.
https://github.com/opencog/opencog/issues/309

### DIFF
--- a/opencog/spatial/SuperEntity.cc
+++ b/opencog/spatial/SuperEntity.cc
@@ -74,7 +74,7 @@ bool SuperEntity::splitEdges( const SubEntityPtr& subEntity1, const SubEntityPtr
                     double distance = it1->distanceTo( *it2, &pointInA, &pointInB);
 
                     if ( distance <= math::LineSegment::TOLERANCE_DISTANCE ) {
-                        // distance lesser than treshold, so two edges intersects each other
+                        // distance lesser than threshold, so two edges intersects each other
 
                         // check if the intersection is just the begin or end points intersection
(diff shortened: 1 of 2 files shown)


linas:
Wow. Does this mean that the unit test now passes?  Do you understand what this code does now? Have you seen  any good documentation for it?  If not, care to add it to the wiki?

brett79:
Yes, the unit test passes. I understand what the code does, there's enough comments in there to make it out. I haven't looked outside the code to see where it has been used. I'm guessing it's for embodiement 2D navigation. There's some references to 3D, but I think it would need some work before it could use the z-axis. No, I haven't seen any good documentation, but I will throw something on the wiki. I should do an illustration, it's a lot simpler when you plot the 2D results.

linas:
Oh, OK.  I was thinking of an overview -- maybe 2-3 sentences giving the overview. But perhaps there's enough.  

The entire spatial directory has a big mass of code in it, and one very slim README.

brett79:
So I've added to the Wiki. Hopefully it's in a relevant location, hopefully the definition is clear :) ? http://wiki.opencog.org/w/SpaceServer_(Embodiment)#Notes_on_opencog.5Cspatial.5CSuperEntity.cc

linas:
Wow! Thanks!

PR #1180, opened 2015-01-06
make the rest api thread a daemon

misgeatgit:
Now both REST API server and cogserver can be shutdown from the
opencog shell when the shutdown command is prompted.


cosmoharrigan:
This worked properly for me. Thanks @misgeatgit.

PR #1187, opened 2015-01-08
SuReal: scheme primitives for running the Pattern Matcher

williampma:
Implemented two scheme primitives for the next version of SuReal: `sureal-match` and `sureal-get-mapping`.  These primitives are for SuReal's internal usage in the next stage, so probably will not be used by others.  Here's some information to help code review.

* `(sureal-match ...)` accepts a `SetLink` as input and uses the Pattern Matcher to find matching sentences.  Two hypergraph matches iff the structure is the same, and for each pair of matched nodes (ConceptNode/PredicateNode/etc), the corresponding `WordNode` has the same LG grammar dictionary entry.  The WordNode is found by extracting the name from the ConceptNode/PredicateNode and finding a WordNode with the same name.

  Two words' LG entries matched iff the structure of `(LgWordCset (WordNode "A") ...)` `(LgWordCset (WordNode "B") ...)` link to same LG hypergraph at "..."

* `(sureal-get-mapping ...)` can be called after running `(sureal-match ...)`.  It accepts an `InterpretationNode` returned by `sureal-match` as input, and returns a list of node-to-node mapping.

## Example

Here's a basic example:
```
(lg-get-dict-entry (WordNode "he"))
(lg-get-dict-entry (WordNode "she"))
(lg-get-dict-entry (WordNode "runs"))
(lg-get-dict-entry (WordNode "walks"))
(r2l "He runs.")
(r2l "He runs quickly.")
(sureal-match (SetLink (EvaluationLink (PredicateNode "walks") (ListLink (ConceptNode "she")))))
```
This will returns two `InterpretationNode` like
```
((InterpretationNode "sentence@d6f3a339-c2db-457a-b320-d3cf60aef6ce_parse_0_interpretation_$X")
 (InterpretationNode "sentence@eb8cb479-c2d6-4bcb-bc64-9973d5d083c8_parse_0_interpretation_$X")
)
```
ranked.  In this case, the InterpretationNode for "He runs" rank higher than "He runs quickly" because there's less left-over unmatched links.

Running the following
```
(sureal-get-mapping (InterpretationNode "sentence@d6f3a339-c2db-457a-b320-d3cf60aef6ce_parse_0_interpretation_$X"))
```
returns
```
(((PredicateNode "walks")
 (ConceptNode "she")
) ((PredicateNode "runs@2ed03b71-2cdf-4cd7-bd3b-384e1ab8b16b" (stv 0.001 0.99000001))
 (ConceptNode "he@4a160823-df3e-447b-b6eb-085609e4960b" (stv 0.001 0.99000001))
))
```
which is a list mapping "walks" to "runs", and "she" to "he" of the R2L output.

## Longer Example
For testing `SetLink` with multiple clauses for PM query
```
(lg-get-dict-entry (WordNode "dog"))
(lg-get-dict-entry (WordNode "yellow"))
(lg-get-dict-entry (WordNode "green"))
(lg-get-dict-entry (WordNode "cat"))
(lg-get-dict-entry (WordNode "hates"))
(lg-get-dict-entry (WordNode "loves"))
(r2l "The yellow dog hates the cat.")
(sureal-match
	(SetLink
		(EvaluationLink
			(PredicateNode "loves")
			(ListLink (ConceptNode "cat") (ConceptNode "dog")))
		(InheritanceLink (ConceptNode "cat") (ConceptNode "green"))))
```

## Harder-stuff
Currently, the LG dictionary matching is very strict, ie, we can match
```
he <-> she
runs <-> walks
eats <-> drinks
green <-> yellow
```
but not
```
I <-> you
runs <-> eats
```
since the LG entry is not 100% the same.  In some future version partial LG dict entry match will be needed (for some of the cases... It is good to match I <-> you, but not runs <-> eats).

## TODO
Actually plug these primitives into the SuReal pipeline to produce sentences.


opencog/nlp/lg-dict/LGDictReader.cc
```diff
@@ -69,7 +71,36 @@ LGDictReader::~LGDictReader()
  * @param word   the input word string
  * @return       the handle to the newly created atom
  */
-Handle LGDictReader::getAtom(const std::string& word)
+Handle LGDictReader::getAtom(const std::string word)
```
linas: why are you removing the reference? Using the reference avoids a string copy, right?

opencog/nlp/lg-dict/LGDictReader.cc
```diff
@@ -23,6 +23,8 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <opencog/guile/SchemeSmob.h>
```
linas: I don't think you need this include...

opencog/nlp/sureal/SuRealModule.cc
```diff
@@ -0,0 +1,211 @@
...
+ * by "sureal-match" as input, and returns the complete node-to-node mapping.
+ *
+ * @param h   an InterpretationNode returned by "sureal-match"
+ * @return    a list of two lists; the first list will be the nodes extracted
+ *            from the original query, the second list will the the
+ *            corresponding mapping for each node.
+ */
+HandleSeqSeq SuRealModule::do_sureal_get_mapping(Handle h)
+{
+#ifdef HAVE_GUILE
```
linas: I don't see why  this ifdef is needed here...

opencog/nlp/sureal/SuRealModule.cc
```diff
@@ -0,0 +1,211 @@
...
+    // clear the old results on new "sureal-match" to sync with "sureal-get-mapping"
+    m_results.clear();
+
+    // only accept SetLink
+    if (h->getType() != SET_LINK)
+        return HandleSeq();
+
+    AtomSpace* pAS = SchemeSmob::ss_get_env_as("sureal-match");
```
linas: If you wanted to, you could avoid depending on guile here simply by getting the atomspace from the cogserver. (it would even be a little bit faster).

SchemeSmob::ss_get_env_as() was invented for that code that cannot get access to the cogserver, or equivalently, to avoid having everything to depend on the cogserver.  So you can continue to use it here, if you wish.  You have a choice.

linas:
wow.

linas:
I merged, because almost all of this code hits on bits and pieces that I'd written.   I like what you did, although I did not try to think about whether pattern matching really is the best & easiest solution to the problem you are trying to solve.  I assume that it must be...

linas:
Regarding the `I <-> you` matchup: the language learning pipeline will also need the ability to spot disjunct collections that are similar but not identical.  ...

Converting DNF (disjunct normal form, which is how the LG exp's are structured) into "ENF" ("elegnat normal form", which minimizes the number of connectives) and then comparing ENF trees for similarity might be a good (i.e. natural-seeming) way of comparing similarity, but is also probably one of the more complex ways of doing this ...

linas:
FYI:  The pattern matcher also has the ability to reject certain matches: e.g. match "she walks" but not "she walks with", or, more generally, to match and report optional sub-patterns.  i.e. there can be a match if the optional piece is missing, there can be a match if the optional piece is present, or if you wish, there can be a reject if the optional piece is present.

linas:
One last comment: a few explicit tests for VariableNode have snuck into the pattern matcher, even though they probably should not be there. These may or may not interfere with what you are doing here.

opencog/nlp/sureal/SuRealModule.cc
williampma: What would happen if there are multiple atomspaces?  Will that still work?

linas: Oh, well, if you look at it that way then .. no ...

PR #1188, opened 2015-01-09
Added NumberNode for scm functions that return WordNode

williampma:
Hopefully doesn't break anything.

### DIFF
--- a/opencog/nlp/relex2logic/utilities.scm
+++ b/opencog/nlp/relex2logic/utilities.scm
@@ -51,7 +51,10 @@
 ; Return #t or #f depends on whether the node as a WordNode.
 ;
 (define (has-word? node)
-	(not (null? (cog-node 'WordNode (cog-name node))))
+	(not (and
+		(null? (cog-node 'WordNode (cog-name node)))
+		(null? (cog-node 'NumberNode (cog-name node)))
+	))
 )
 
 ; -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


linas:
Well, I really really don't like this fix.  The core problem is that reelex should not be generating a NumberNode in the first place -- that was a design flaw, which should be reverted.  The reason that it is a design flaw is clearly demonstrated by your patch here: it takes simple code and makes it more complex, but adding support for non-standard, exceptional cases.  

There is another reason:  for morphology learning, I need to work with ascii strings and not NumberNodes. So the relex code breaks morphology learning.

Another reason: langauge learning looks for correlations between pairs of words.  With number node, I have 4x as many cases to deal with: word-word, word-number, number-word, and number-number.  

Disjuncts can connect 2,3,4,5,6 words, and I now have to handle any one of those, in any position, being a NumberNode. So there is a combinatoric explosion there.  So that make thigngs really out of control. ... 

and finally, the comments in the relex code itself:

```
         // XXX FIXME -- this is a bit of a crazy hack that attempts
         // to handle numbers, but does it wrong.  Why is this wrong?
         // 1) because Europeans use a comma for a decimal, and a period
         //    for a thousands separator
         // 2) Because americans use a comma for a thousands separator,
         // 3) This fails to handle number words, like "one two three"
         // 4) This fails to handle time, dates, etc.
         // In my opinion, the below whould be removed, and higher
         // mathematics should be handled in Opencog, and not here...
         if (word.matches("[-+]?[0-9]*?\\.?[0-9]+"))
```

I conclude that the only correct fix is to remove the NumberNode generation inn relex, and instead, handle it correctly in the opencog pipeline.

linas:
Based on the above comments, I'm rejecting this request.

Oh, and one more reason:  besides NumberNode, there is also a need to handle TimeNode, AtTimeNode, etc. which is another reason why the NumberNode stuff should not be there...

linas:
I suppose that someone will be unhappy about this .. but  its thee right thing to do.  I removed the code that generates NumberNode from relex, just now.

PR #1212, opened 2015-01-13
PM: Make get_connected_components uses "vars"

williampma:
`get_connected_components` was ignoring its `vars` argument.  Changed `FindVariables` to accept a vars list.  Exisiting usage should be intact.

### DIFF
--- a/opencog/query/PatternUtils.cc
+++ b/opencog/query/PatternUtils.cc
@@ -128,7 +128,7 @@ void get_connected_components(
 					components[i] = curr_component;
 
 					// Add to the varset cache for that component.
-					FindVariables fv;
+					FindVariables fv(vars);
 					fv.find_vars(cl);
 					for (Handle v : fv.varset) cur_vars.insert(v);
 					component_vars[i] = cur_vars;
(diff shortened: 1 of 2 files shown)


linas:
I don't think this is right (at all, its very very wrong), but it does tell me what the correct solution actually should be.  Let me explain.

When a graph has multiple disjoint components,  the variables in each component will be a *subset* of the total variables in the whole thing.  In fact, the subsets themselves will be disjoint: that's the defiintion of the disconnected graph.   So .. find_vars should not be lookinf for VariableNodes, it should be looking only for those atoms that were given to it, in vars ... 

Its easier for me to fix this than to explain in grater detail, so let me do that.  Thanks, however, for the try, because I had not previously realized that there is a better solution ....

linas:
oh, wait, may this is the right fix
hang on ...

linas:
Sorry ... in fact, that was exactly the right fix...  my bad.

linas:
This was a very good catch!   It must not have been easy to find, or to figure out how to fix it.  I just pushed a commit that cleaned it up just a bit, and added documentation.

PR #1216, opened 2015-01-13
comments out the part that doesn't compile

shujingke:


linas:
A much easier way to "comment things out" is just to put an `#if 0` at the start, and `#endif` at the end.  That way, its only two lines of code...

PR #1217, opened 2015-01-13
Pattern match: minor twerking of William Ma's fix.

linas:
Added documentation, changed some names.


anitzkin:
This was the only case of "minor twerking" I was able to find using google
. . . just to be clear about our language :)

https://www.youtube.com/watch?v=aKTXqmBegWw

linas:
Quite right. I sloppily think in terms of there being a twerker and twerkee, which is not exactly the case.

Issue #1107, opened 2014-11-18
SuReal: name matching bug


williampma:
The following
```
(r2l "The brown hotdog tastes bad.")
(sureal (SetLink (InheritanceLink (ConceptNode "dog") (ConceptNode "mad"))))
```
give me
```
#<values (((the mad dog tastes bad .)) (0))>
```
which I don't think is intended?

williampma:
Fixed by newest SuReal changes.

Issue #1106, opened 2014-11-17
SuReal: recursive mapping


williampma:
This one will be quite complicated to implement I imagine.

Right now
```
(r2l "The mad dog hates the tree.")
(sureal (SetLink (InheritanceLink (ConceptNode "cat") (ConceptNode "yellow")) (EvaluationLink (PredicateNode "likes") (ListLink (ConceptNode "cat") (ConceptNode "tree"))))
```
will just return
```
#<values (((the mad cat likes the tree .)) (-1))>
```
without using the "yellow cat" information (hence the -1 weight).  That's because `(InheritanceLink (ConceptNode "cat") (ConceptNode "yellow"))` cannot map to `(InheritanceLink (ConceptNode "dog") (ConceptNode "mad"))`.

However, after mapping `(EvaluationLink (PredicateNode "likes") (ListLink (ConceptNode "cat") (ConceptNode "tree")))` to `(EvaluationLink (PredicateNode "hates") (ListLink (ConceptNode "dog") (ConceptNode "tree")))`, it should be possible to see "yellow cat" can be mapped to "mad dog". 

That means some sort of recursive algorithm is needed so that when a new thing is mapped, all the previous unmapped stuff need to be rechecked again.

williampma:
Fixed by newest SuReal changes.

PR #1219, opened 2015-01-14
SuReal: works with the new scheme primitives

williampma:
Make SuReal uses the new Pattern Matcher solution.

* Handle disconnected clauses, which appear quite often with SuReal's query
* Allow multiple solutions per interpretation (it is possible for conjunction for example)
* Make `(r2l ...)` starts filling atomspace with LG dictionary entries.

Since `(r2l ...)` will start making these huge LG dictionary graph, it might inconvenience ppl who are using it while checking the atomspace with `(cog-prt-atomspace)`.  In that case, just comment out lines L123-128 on surface-realization.scm

The words used in the input `SetLink` need to have the corresponding `WordNode` or `WordInstanceNode` before calling `sureal`

Issue #1106 & #1107 is fixed by the new implementation.

With the new way SuReal matches words, the testing sentences used in microplanning's test-atomspace.scm will no longer work.  To fix, either conjunction and anaphora need to be handled better on the R2L output, or microplanner needs to be changed so that pronoun insertion happens while chunking, instead of after as it currently does.  I will think about fixing it on the microplanner side a bit more.


opencog/nlp/sureal/README.md
```diff
@@ -1,5 +1,9 @@
 # Surface Realization
 
 Functions for surface realization of output from micro-planner's output.
-The main function is sureal which takes in a SetLink and returns a
+The main function is sureal which takes in a `SetLin`k and returns a
```
linas: badly located backtic

linas:
Looks reasonable to me, but can't be merged bacuase Nil renamed _atom_space to _as and this causes a merge conflict in suRealPMCB.cc

opencog/nlp/sureal/README.md
williampma: oops.

williampma:
rebased.

opencog/nlp/sureal/SuRealModule.cc
```diff
@@ -129,13 +124,106 @@ HandleSeq SuRealModule::do_sureal_match(Handle h)
...
+    // separate the disconnected clauses (this will happen often with SuReal)
+    std::set<HandleSeq> connectedClauses;
+    get_connected_components(sVars, qClauses, connectedClauses);
+
+    logger().debug("[SuReal] Found %d disconnected components", connectedClauses.size());
+
+    std::map<Handle, std::vector<std::map<Handle, Handle> > > collector;
+
+    // call the pattern matcher on each set of disconnected commponents
+    for (auto& c : connectedClauses)
```
linas: I did not study this carefully enough to see what you are trying to do, but .. wanted to point something out, that looks suspicious to me.

Suppose, for example there are 4 connected components, and the first one has M groundings, and the second has N, and the 3rd hand 4th have P, Q, then there are  M x N x P x Q total possibilities.  That requires 4 total nested for-loops to explore all the possibilities.  In general, if there are K components, you need K nested loops, i.e. you'd have to design something recursive to explore the possibilities ...  Seems to me that would have to apply here, as well, no?

opencog/nlp/sureal/SuRealModule.cc
```diff
@@ -129,13 +124,106 @@ HandleSeq SuRealModule::do_sureal_match(Handle h)
...
+
+            auto& existingMaps = it->second;                // a vector of previous mappings
+            auto& appendMaps = pmcb.m_results[it->first];   // a vector of unmerged mappings
+            std::vector<std::map<Handle, Handle> > newMaps;
+
+            // check all combinations of all the old mappings to the new
+            for (auto& em : existingMaps)
+            {
+                for (auto& am: appendMaps)
+                {
```
linas: Oh never mind. You write good code, and I talk before I think. Yes, if you accumulate everything done so far into "existing", and then add on to each of those, it will work.  So yes, this looks right after all. Sorry...

PR #1221, opened 2015-01-14
Pattern matcher improvements or so it says

ngeiswei:
Contains a bunch of cleaning, add comments, optimizations, and a memory leak fix.

### DIFF
--- a/opencog/atomspace/Foreach.h
+++ b/opencog/atomspace/Foreach.h
@@ -18,16 +18,16 @@ namespace opencog
  */
 
 
-//! Invoke the callback on each atom in the outgoing set of handle h.
+//! Invoke the callback on each atom in the outgoing set of handle h
+//! till one of them return true, and returns true as a
+//! result. Otherwise the callback is called on all outgoings and
+//! false is returned.
 /// XXX TODO this should be moved to class Link
 template<class T>
 inline bool foreach_outgoing_handle(LinkPtr ll, bool (T::*cb)(Handle), T *data)
 {
-    const std::vector<Handle> &vh = ll->getOutgoingSet();
-    size_t sz = vh.size();
-
-    for (size_t i = 0; i < sz; i++) {
-        bool rc = (data->*cb)(vh[i]);
+    for (Handle out_h : ll->getOutgoingSet()) {
+        bool rc = (data->*cb)(out_h);
         if (rc) return rc;
     }
     return false;
@@ -35,7 +35,10 @@ inline bool foreach_outgoing_handle(LinkPtr ll, bool (T::*cb)(Handle), T *data)
 
 /* ----------------------------------------------------------- */
 
-//! Invoke the callback on each atom in the incoming set of handle h.
+//! Invoke the callback on each atom in the incoming set of handle h
+//! till one of them returns true, and returns true as a
+//! result. Otherwise the callback is called on all incomings and
+//! false is returned.
 /// XXX TODO this routine should be moved to class Atom
 template<class T>
 inline bool foreach_incoming_handle(Handle h, bool (T::*cb)(Handle), T *data)
--- a/opencog/query/PatternMatchEngine.cc
+++ b/opencog/query/PatternMatchEngine.cc
@@ -730,7 +730,7 @@ bool PatternMatchEngine::pred_up(Handle h)
 void PatternMatchEngine::get_next_untried_clause(void)
 {
 	// First, try to ground all the mandatory clauses, only.
-	// Optional clauses are grounded only after all teh mandatory
+	// Optional clauses are grounded only after all the mandatory
 	// ones are done.
 	if (get_next_untried_helper(false)) return;
 
@@ -1074,13 +1074,13 @@ void PatternMatchEngine::print_solution(
 		if (soln == Handle::UNDEFINED)
 		{
 			printf("ERROR: ungrounded variable %s\n",
-				var->toShortString().c_str());
+			       var->toShortString().c_str());
 			continue;
 		}
 
 		printf("\t%s maps to %s\n",
-			var->toShortString().c_str(),
-			soln->toShortString().c_str());
+		       var->toShortString().c_str(),
+		       soln->toShortString().c_str());
 	}
 
 	// Print out the full binding to all of the clauses.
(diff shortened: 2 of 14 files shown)


opencog/atomspace/Foreach.h
```diff
@@ -18,16 +18,16 @@ namespace opencog
  */
 
 
-//! Invoke the callback on each atom in the outgoing set of handle h.
+//! Invoke the callback on each atom in the outgoing set of handle h
+//! till one of them return true, and returns true as a
+//! result. Otherwise the callback is called on all outgoings and
+//! false is returned.
 /// XXX TODO this should be moved to class Link
```
linas: FYI, I've never been entirely thrilled with these foreach-callback tools, so if you ever glimpse a better way, and have the time to fiddle with it, it could be replaced...

opencog/atomspace/Foreach.h
```diff
@@ -18,16 +18,16 @@ namespace opencog
...
 inline bool foreach_outgoing_handle(LinkPtr ll, bool (T::*cb)(Handle), T *data)
 {
-    const std::vector<Handle> &vh = ll->getOutgoingSet();
-    size_t sz = vh.size();
-
-    for (size_t i = 0; i < sz; i++) {
-        bool rc = (data->*cb)(vh[i]);
+    for (Handle out_h : ll->getOutgoingSet()) {
+        bool rc = (data->*cb)(out_h);
         if (rc) return rc;
```
linas: One reason I often used the old style, of putting the size in an int, and looping on an int, is that I beleive that the compiler will turn this into efficient.(at least, w.r.t. the loop termination condition).  By contrast, your new code is not obviously as efficient. I can't tell what the compiler would actually do in this case.   Note that this is fairly performance-critical code: the entire pattern matcher is a recursive nesting of these for-loops.   So I agonize about this ...

opencog/query/PatternMatchEngine.cc
```diff
@@ -30,7 +30,7 @@
 
 using namespace opencog;
 
-// #define DEBUG 1
+#define DEBUG 1
 #ifdef WIN32
```
linas: whoops!

opencog/query/PatternUtils.cc
```diff
@@ -117,21 +116,18 @@ void get_connected_components(
 			size_t nc = components.size();
 			for (size_t i = 0; i<nc; i++)
 			{
-				std::set<Handle> cur_vars(component_vars[i]);
```
linas: Please double check this, and the next few changes.  When I originally wrote this code, I remember thinking that I needed to make explicit copies, and not references. for some reason or another.  I agonized over this, because making copies hurts performance.   So ... maybe your changes are correct, but that is not obviously the case just by glancing at this.

Re: my earlier remarks about how unit tests test everything? Its possible that the unit tests do not actually exercise this code enough.

ngeiswei: I did check very carefully at the time. Actually I think it would benefit the code readability and safety to consity all that code. For instance node_match arguments could be const, the fact that they are not const suggests to the reader (to me at least) that they might be changed while it's not the case. I'll probably create a branch for that as well.

ngeiswei: Oh, well I start to understand why constness isn't supported so well, turns out constness isn't very well supported in the atomspace code as well... It starts to look like a much bigger task...

ngeiswei: Oh, note that actually this optimization would not be facilitated with more const support (as I've just realized). I'm thinking perhaps the only think that would go wrong is something with multithread support, but if that were the case I don't see why.

linas: add const whereever you want. however, in the atomspace, realize this:  under the covers, each atom has a use_count in it, which is incremented whenever a a Handle or AtomPtr points at it, and is decremented when a Handle or AtomPtr is destroyed.  So Atoms are kind-of-like const, but not really.  No clue how the compiler handles this.

Next: Handles are almost const, but not quite: if a Handle only has a uuid in it, but doesn't have a pointer to the actual atom, that needs to be fixed up.  It all happens deep in the guts, and so it is kind-of-like almost const, but I imagine that there could be some really weird bugs if the compiler is fooled. I don't know. So far, no problem.

The only place where I would expect this would be in the postgress side, where the database only stores uuid's and pointers are created later.  There's also a few oddball cases in the scheme bindngs which use only uuids ...

ngeiswei: To "fool" the compiler into constness one can decorate the members with the mutable keyword. I don't think it creates mess on the compiler's end. From the user's perspective I guess it means that we consider those mutable members will have no external side effects.

Regarding Atom, the tricky thing is that it inherits std::enable_shared_from_this, and then getHandle calls shared_from_this(), now I checked and there are 2 versions of it (in shared_ptr.h)

```
      shared_ptr<_Tp>
      shared_from_this()
      { return shared_ptr<_Tp>(this->_M_weak_this); }

      shared_ptr<const _Tp>
      shared_from_this() const
      { return shared_ptr<const _Tp>(this->_M_weak_this); }
```
so apparently our code is calling the non const version. I'm not sure yet how to deal with that...

ngeiswei: Ahah, I found the problem! Handle is missing the constructor Handle(std::shared_ptr<const Atom> atom);

instead it only has
```
    explicit Handle(AtomPtr atom);
```

I don't understand enough the inner workings of the atomspace to know if that could be just added, or if it requires some big changes...

ngeiswei: It seems to me (without thinking too deep), what we would need is 2 types of Handles,
```
Handle
```
for handle pointing to atoms that may be modified
```
ConstHandle
```
for atoms that may not be modified. A bit like iterator and const_iterator in STL.

linas: I can add the constructor `Handle(std::shared_ptr<const Atom> atom);` but I don't understand the problem that its solving.  I really would strongly prefer not to introduce ConstHandle.   Can you show some assembly code generated by gcc-4.8 where some sort of register reload could actually be avoided by using ConstHandle?

ngeiswei: I don't know if the compiler is that smart (we'd have to try I guess). But I have to say I'm more attached to constness by the information it brings to the developer, and the additional safety to the type system. I wouldn't personally start fiddling with const in the atomspace API however, certainly not at my current level of understanding.

PR #1224, opened 2015-01-15
Constify pattern matcher

ngeiswei:


ngeiswei:
I only managed to constify the *_match methods, to push further I'd need to constify the atomspace API and it's a too big task, that I'm besides not extremely confident to carry.

Also, methods like match or perform_search could certainly be further constified, but it's not clear yet to me where pre-processing, like validate_clauses, that modifies the arguments will take place.

ngeiswei:
Turns out my changes break the utests, I didn't expect that which is why I merged before getting the results of the utests... Merge has been reverted...

ngeiswei:
Well, turns out it's really hard to constify anything surrounding atom manipulation. For instance Atom::getHandle() should be const but it can't because of shared_from_this(). So I'm basically giving up. I'll add some comment though...

linas:
Yes right, re: getHandle ... 

Also, in the original pattern-match design, I explicitly did NOT want the callbacks to have const args, because I envisioned a case where the callbacks might modify truth values or attention values of the atoms that they look at.   I kept thinking that maybe PLN would want to change the truth values, especially in the 'clause_match' or in the final grounding-found callback.  Its also plausible to tweak attention value, every time the atom is touched ...

ngeiswei:
I see... My comment should be corrected then.

ngeiswei:
Oh, but const Handle doesn't prevent from modifying the Atom!!! See
```
    inline Atom* operator->() const {
        Atom* ptr = _ptr.get();
        if (ptr) return ptr;
        if (ULONG_MAX == _uuid) return NULL;
        return cresolve();
    }
```
in Handle.h

This seems to be tied with my remark that we may want to have a ConstHandle (https://github.com/opencog/opencog/pull/1221/files#r23034340).

linas:
Oh, heh. Hah ha.  Right. I knew that :-) OK, sure, yes, we could pass const Handle& for almost everything.  This will work because Handle::_ptr will be non-null for almost all atoms that the pattern matcher will ever touch.  The only place where its null is in a few places in the sql-persistance code, and a few places in the guile code.

However, the only reason I am interested in const is so that the, in theory, the compiler can save a few instructions and run faster. However, without looking at assembly code, I don't know if this actually happens for opencog atomspace code, so am cautious about over-engineering solutions to problems tht might not exist ...

Issue #1220, opened 2015-01-14
node_match, link_match, etc have confusing name


ngeiswei:
The naming DefaultPatternMatchCB::node_match, etc, is really confusing as it returns false when it actually matches and true when it doesn't.

I do understand that unix functions do historically return 0 when it does well and != 0 otherwise as a code error, but here we're dealing with C++ bool type, it's really counterintuitive to return false when it matches.

I suggest to either

1. rename them node_mismatch, link_mismatch, ...

2. return true when it matches, false when it doesn't.

3. do nothing.

I do understand that other functions like Implicator::grounding works the same way, return false when the grounding succeeds. It's also confusing but at least it doesn't blatently contradict the method name.

1. is gonna make the method names 3 char longer, I don't like that.

2. is opening the risk to introduce bugs, I don't like that.

3. the confusion still remains, I don't like that either.

Help me!

williampma:
I agree.  I had to constantly remind myself of this fact when I was writing the callbacks in order to avoid making a mistake (and I still ended up making one mistake and returning the opposite value).

ngeiswei:
OK, this makes me want to go with 2. It's a change that requires more care but the unity tests seem good enough. What I might do is mess up those functions (i.e. return true when it should return false) and see if the unit test can detect it, if it can then I'll do ahead and make the change. If it can't then I'll write more unit test.

Anyway, I won't do that today, but if you have any arguments against that, please speak out.

linas:
Yes. Its a design flaw that seemed reasonable at the time ...

Issue #1174, opened 2014-12-31
SCMExecutionUTest sometimes pass, sometimes not


williampma:
It passes some of the times and doesn't on other times.

When it doesn't pass:

```
[INFO] Using config file found at: /home/william/opencog/build/./lib/opencog-test.conf
[2014-12-31 07:41:01:868] [DEBUG] [NetworkServer] constructor
[2014-12-31 07:41:01:868] [DEBUG] [SystemActivityTable] constructor
[2014-12-31 07:41:01:912] [DEBUG] [SystemActivityTable] init
[2014-12-31 07:41:01:913] [DEBUG] Searching path /home/william/opencog/build/opencog/atomspace/core_types.scm
Running cxxtest tests (2 tests)[2014-12-31 07:41:01:951] [INFO] Loaded /home/william/opencog/build/opencog/atomspace/core_types.scm
.[2014-12-31 07:41:01:959] [DEBUG] BEGIN TEST: test_multi_threads
atomspace size:4003
[2014-12-31 07:41:04:560] [DEBUG] END TEST: test_multi_threads
SCMExecutionUTest: hashtab.c:137: vacuum_weak_hash_table: Assertion `removed <= len' failed.
Aborted (core dumped)
```

linas:
Closing. Don't know what else to do right now, other than to engauge the guile community, or install guile-2.2 manually, or just wait.

PR #1226, opened 2015-01-15
Negate *_match semantics

ngeiswei:
Here's my attempt to flip the semantics of *_match and tree_compare (i.e. a fix for issue #1220).

It seems to work but I've noticed that MultiThreadUTest and SCMExecutionUTest randomly fail. However I did see them failing before. So I think it's kinda crucial that we fix them, before perhaps merging that branch (unless those 2 tests are obviously unrelated to the pattern matcher).

### DIFF
--- a/opencog/atomspace/ForeachTwo.h
+++ b/opencog/atomspace/ForeachTwo.h
@@ -21,6 +21,8 @@ namespace opencog
  * Invoke the callback on each pair of corresponding atoms in the
  * vectors va and vb. This iterator is typically useful for making
  * comparisons between atoms.
+ *
+ * Returns true if all calls return true, return false otherwise.
  */
 template<class T>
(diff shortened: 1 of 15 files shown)


linas:
I've never seen SCMExecutonUTest fail.

When using guile-2.0, the MultiThreadUTest will sometimes crash, *after* it has finished, and passed the test, during the shutdown sequence.  This is due to a guile bug and does not happen w/ guile-2.2 theres a bug for this, william ma opened it: #1172.  I don't know what to do about that.  Oh wait .. maybe I have these two tests mixed up....

linas:
Merged. The tests all pass for me. If there are lurking bugs, then they've always been there.

linas:
Sorry, its bug #1174 and its here: http://lists.gnu.org/archive/html/bug-guile/2014-12/msg00008.html

PR #1230, opened 2015-01-20
Add example showing how to obtain atom type names in Python

cosmoharrigan:


linas:
FWIW, all other examples are in the examples directory:  I propose that /opencog/python/examples should be moved to /examples/python

cosmoharrigan:
@linas Done.

PR #1234, opened 2015-01-21
guile: Added OpenCog scheme module support

williampma:
Wrap the OpenCog scheme primitives into modules so that they can be loaded & used when creating custom modules in new scm files.  Previously defining new modules lost the binding to the opencog primitives.

* The default `cog-new-node` style primitives are in `(opencog)`
* Custom primitives are in `(opencog extension)`
* Scheme atom types shortcut such as `(ConceptNode "abc")` are in `(opencog atomtypes)`

All are loaded by default to allow existing code to function.  Existing tests that passed before still pass.  **You need to do `make clean` so the changes to the _types.scm generation code will take place.**

You can see what functions are in the module with
```
(module-for-each (lambda (symbol value) (display symbol) (newline)) (resolve-interface '(opencog)))
(module-for-each (lambda (symbol value) (display symbol) (newline)) (resolve-interface '(opencog atomtypes)))
(module-for-each (lambda (symbol value) (display symbol) (newline)) (resolve-interface '(opencog extension)))
```

When defining your own scm module externally, you can load the opencog module like
```
(define-module (your mod name))
(use-modules (opencog))
(use-modules (opencog atomtypes))
(use-modules (opencog extension))
```

### DIFF
--- a/opencog/guile/README
+++ b/opencog/guile/README
@@ -100,6 +100,18 @@ utilities can be pulled up at the scheme shell prompt, by saying
 have to consult the individual scm files for more info.
 
 
+== Modules ==
+-------------
+Currently OpenCog's scheme functions are broken into 3 different scheme
+modules: (opencog), (opencog extension), and (opencog atomtypes).  Each
+one can be listed with the following code:
+
+(module-for-each (lambda (symbol value) (display symbol) (newline))
+(resolve-interface '(opencog)))
+
+Replace '(opencog) with the corresponding module name.
+
+
 == Core Functions ==
 --------------------
 The following are the "core" functions, implemented in the guile-to-C++
@@ -791,9 +803,3 @@ ToDo
    data corruption. However, guile needs to be fixed to change this 
    behaviour.
 
-
--- list all proceedures in a module:
-   (module-for-each (lambda (symbol value) (display symbol) (newline))
-   (resolve-interface '(guile)))
-
-   Make the above easier for users.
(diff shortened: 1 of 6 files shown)


opencog/guile/README
```diff
@@ -100,6 +100,18 @@ utilities can be pulled up at the scheme shell prompt, by saying
...
+== Modules ==
+-------------
+Currently OpenCog's scheme functions are broken into 3 different scheme
+modules: (opencog), (opencog extension), and (opencog atomtypes).  Each
+one can be listed with the following code:
+
+(module-for-each (lambda (symbol value) (display symbol) (newline))
+(resolve-interface '(opencog)))
+
+Replace '(opencog) with the corresponding module name.
```
linas: This needs to be on the wiki too. 

Once upon a time, the wiki and the README were copies of each other, but have gotten out of sync ..

linas:
Wow, Willliam! Thank you!

linas:
But what made you decide that this needed doing, now?

williampma:
I had always wanted to do it since you mentioned "modules" when I first checked in the microplanning code, just didn't know how to do it.  Recent experience writing scheme primitives made me understand the guile code a bit better, so I could finally do it.

linas:
OH, OK. Well, I was thinking more along the lines of turning utils.scm and many of the other xxx.scm files into modules.   The primary reason for that would be to allow the modules to have private, internal functions that don't have name-space clashes with identically named functions in other modules... as we create more and more scheme code, its the naming collisions that got me thinking.

williampma:
Yeah, those probably should be in modules, but they will need this fix to be able to access the opencog scheme primitives from within their own modules.

So this is step 1 for doing that.

linas:
Ah OK. I was just pleasantly surprised.

PR #1237, opened 2015-01-22
Add scheme API timing example

cosmoharrigan:


examples/python/scheme_timer.py
```diff
@@ -0,0 +1,35 @@
+"""
+Checks the execution time of repeated calls to the Scheme API from Python
+
+Runs an empty Scheme command NUMBER_OF_ITERATIONS times and displays the
```
linas: Ouch. I'm guessing you measured about 3K/second ?  This has got to hurt: you've got to leave the python context (this takes about 0.05 millisecs, according to my measurements, then enter the scheme context (about 0.1 millisecs) then leave the scheme context (0.1 millisecs) then reenter python (about 0.1 millsiecs) or very roughly about 0.35 millisecs per op, or about 3K/second.   This is a ballpark number I'm guessing .. is that what you get?

I would rather not have code like this in the examples directory. It belongs in the benchmarks directory, and should be accompanied by a diary of what was actually measured, on which version of py and guile, on what date, and what CPU.

PR #1246, opened 2015-01-27
lg-dict: construct DNF by default

williampma:
Construct DNF be default; side-effect is that the hypergraph is even larger than before (also the construction is slower... may need some optimization)

I noticed for some words with huge lg expression tree if I try to construct the scm string and build everything in one go using `_scm_eval->eval_h(...)`, it will fail and return an undefined handle.  Looks like there's a limit to how large the scm string be for atom construction.

So ended up using C++ construction code instead.

### DIFF
--- a/opencog/nlp/lg-dict/CMakeLists.txt
+++ b/opencog/nlp/lg-dict/CMakeLists.txt
@@ -6,6 +6,7 @@ INCLUDE_DIRECTORIES (
 
 ADD_LIBRARY (lg-dict SHARED
 	LGDictReader
+	LGDictExpContainer
 )
 
 ADD_DEPENDENCIES (lg-dict
--- a/opencog/nlp/lg-dict/LGDictExpContainer.cc
+++ b/opencog/nlp/lg-dict/LGDictExpContainer.cc
@@ -73,7 +73,7 @@ LGDictExpContainer::LGDictExpContainer(Exp_type t, std::vector<LGDictExpContaine
     // need flattening again after changing to dnf
     basic_flatten();
 
-    // convert to normal order (- before +) and also delete repeated stuff in AND
+    // convert to normal order (- before +)
     basic_normal_order();
 }
 
@@ -174,10 +174,8 @@ void LGDictExpContainer::basic_dnf()
 /**
  * Convert to normal order.
  *
- * Putting connectors with - direction before those with + direction.  Also
- * remove duplicates from within AND.
- *
- * Assume everything already in DNF.
+ * Putting connectors with - direction before those with + direction. Assume
+ * everything already in DNF.
  */
 void LGDictExpContainer::basic_normal_order()
 {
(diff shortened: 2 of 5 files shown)


linas:
I'll try to review tomorrow.  If there is a limit to the string size, we should report it to the guile people (assuming the limit is not in our code0.

Yes, converting to DNF can be explosive.

linas:
Does this still need to be merged?

Yes, currently some of the LG expressions flatten out 10K or 20K or more disjuncts, and unless you are careful to be tail recursive, then you'll blow out the scheme stack, which seems to be 10K as you pointed out, in   https://www.gnu.org/software/guile/manual/html_node/Debug-Options.html

Clearly, if your C++ stack was 10K or 20K call frames long, you would blow up the C++ stack too -- so when you re-wrote the algorithm in C++ you changed the algo so that  it does not use such a huge stack.  But if you can change the algo in C++, you can use exactly the same algo in scheme, and not blow up the stack either.  The point is that, when you coded in scheme, you somehow forgot to be tail recursive, which is why it blew up. When you coded in C++, you didn't make this mistake.

Anyway -- Such a large number of disjuncts is really a pathology in the LG dicts, and (in the long run) should be fixed in LG.

williampma:
Yes, this is still needed since SuReal will be working with the disjuncts.

And about the stack thing, all I did was just one call of:

```
eval->eval_h("
(SetLink
   (LgWordCset
       (WordNode "blah")
       (LgOr
           (LgAnd ...)
           (LgAnd ...)
           (LgAnd ...)
           + a godzillion more LgAnd...
      )
   )
   (LgWordCset
      + even more LgAnd
....
)")
```
I think it just mean large atoms like this cannot be constructed in Scheme (at least not in one go anyway; not sure if using Scheme variables will help or not).

linas:
HM, 

OK, but multiple things fly to mind.

Yes, I never anticipated that a link might ever have tens of thousands of atoms in it, and the current scheme api has the members of a link passed as arguments, which would chew up stack:   you are making a function call, and passing tens of thousands of arguments to that function.  I suppose the api could be adjusted if this was a common usage.   FWIW I think the current API will also accept a list (i.e. create a scheme list first, and then say `(LgOr the-list)`.

But this begs the question: why are you building large strings, anyway?  I mean, since you are running C++ code, why say `for (i=0; i<100000; i++) { my_str += "(LgAnd ...)";} eval_h(my_str);` when you could just as easily say `for (i=0; i<100000; i++) { atomspace.addLink(LG_AND, ...) }`

There's a huge overhead of passing around data as strings: not only did you need to create it, but then scheme has to convert the string into a sequence of symbols, look up the bytecode for each symbol, run it, etc. oh, and the bytecode for create-atom just ends up calling C++ again ...  eval_h is "inefficient by design", as Jim Rutt has been gleefully pointing out.

williampma:
Well, I **was** building large string, but not anymore when I converted to C++ atom construction.

PR #1249, opened 2015-01-30
SuReal: minor bug fix

williampma:


linas:
I merged this because that is what I think you wanted .. but feel free to merge changes like this on your own -- its your code.

PR #1256, opened 2015-02-04
README: change xercesc version

kim135797531:
Hello! I'm new to this project.
OpenCog seems to use xercesc library version 3, but README.md file still descript install library version 2.


linas:
thanks!

PR #1257, opened 2015-02-05
Pattern matcher disconnect support

ngeiswei:


opencog/query/AttentionalFocusCB.cc
```diff
@@ -74,3 +76,101 @@ IncomingSet AttentionalFocusCB::get_incoming_set(Handle h)
...
+                                          std::vector<Handle>& clauses)
+{
+}
+
+void AttentionalFocusCB::perform_search(PatternMatchEngine *pme,
+                                        std::set<Handle> &vars,
+                                        std::vector<Handle> &clauses,
+                                        std::vector<Handle> &negations)
+{
+	// In principle, we could start our search at some node, any node,
```
linas: this seems to be a cut-n-paste of the default perform_search() method. It deserves a description of how it differs.  Also, since the difference is minor, perhaps there should be some sort of common code to handle both?  Some flag, or some extra parameter?

opencog/query/PatternMatch.cc
```diff
@@ -35,8 +35,7 @@ PatternMatch::PatternMatch(void) : _used(false) {}
 /// This is just a convenience wrapper around do_match().
 void PatternMatch::match(PatternMatchCallback *cb,
                          Handle hvarbles,
-                         Handle hclauses,
-                         Handle hnegates)
```
linas: Ooof. The "negates" are critical for handling optional clauses.  There are cases where you want to find patterns of a certain shape, but only if they do not have a certain sub-shape.  Alternately, you want to do a different kind of handling, if the subshape is present. ...  I am hoping that some other changes elsewhere restore this functionality ...

opencog/query/VirtualMatch.cc
```diff
@@ -256,20 +266,12 @@ void PatternMatch::validate_clauses(std::set<Handle>& vars,
 		              __FUNCTION__);
 	}
 
-	bogus = remove_constants(vars, negations);
```
linas: this is a CPU waster, I know, but it appears that people sometimes misunderstand bindlink and write them with an invalid syntax.  What we really need is a syntax validator that runs only once, and checks the syntax. After that, the bindlink would not ever need to be checked again, before use ... 

The problem is that invalid syntax tends to cause these hard-to-spot, hard-to-debug crazy crashes ... one looses a lot of time crawling through C++ code, only to realize later that the bindlink syntax was bad ...

opencog/query/PatternMatch.cc
linas: OK, never mind. I see that its just different,  below.

opencog/query/PatternUtils.cc
```diff
@@ -164,6 +164,20 @@ void get_connected_components(const std::set<Handle> &vars,
 	compset.insert(components.begin(), components.end());
 }
 
+void split_clauses_pos_neg(const std::vector<Handle>& clauses,
+                           std::vector<Handle>& affirm,
+                           std::vector<Handle>& negate)
+{
+	for (Handle h : clauses) {
+		if (NOT_LINK == h->getType()) {
```
linas: OK .. so, please realize that the "NotLink" here is a "temporary hack" mostly because I could not think of a better solution.  What is really needed is maybe(?) an "OptionalLink" or something like that .... something that says "if this part of the pattern is found then call optional_clause_match() instead of clause_match()".

The problem with NotLink is philosophical/theoretical: it does not play nice with so-called "intuitionsitic logic" or "constructive logic"  and I'm trying hard to make sure that opencog  defaults to constructive logic instead of classical logic.  The problem with classical logic is that it makes certain types of reasoning (extremely, disaterously) inefficient by blocking the use of certain fast algos. 

Creating an OptionalLink or a MaybeLink is the best I can think of right now, ...

PR #1258, opened 2015-02-05
Rule engine

misgeatgit:
-Changed config file format since conf files were not good enough for grouping attributes
-Separate class for handling Rule
-Added class FCMemory for keeping track of forward chaining inference. Will be useful for final dumping of inference information and analysis.

Changes to forward chaining in progress in a private repo.


lib/default_cpolicy.json
```diff
@@ -0,0 +1,92 @@
+{
+	"rules":
```
linas: why are we writing new config files in json? One of teh important reasons for adding scheme was NOT to manipulate the atomspace, but was to get away from the config files opencog.conf  To add et another config file in a new syntax that is opaque and not machine-controllable .. seems like a mistake to me.  We should not be proliferating, we should be reducing.

linas: anyway, this does not belong in this folder, it belongs in the opencog/reasoning/pln folder

opencog/reasoning/RuleEngine/rule-engine-src/ControlPolicyParamLoader.cc
```diff
@@ -0,0 +1,119 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+#include "ControlPolicyParamLoader.h"
+
+#include <opencog/guile/load-file.h>
+#include <opencog/util/misc.h>
+#include <opencog/util/Config.h>
+#include <opencog/guile/SchemeEval.h>
+
+ControlPolicyParamLoader::ControlPolicyParamLoader(AtomSpace * as, string conf_path) :
```
linas: again -- wasted time writing C++ to read yet another config file format ...

lib/default_cpolicy.json
```diff
@@ -0,0 +1,92 @@
...
+			"name": "pln-rule-modus-ponens",
+			"file_path": "reasoning/RuleEngine/rules/pln/modus-ponens.scm",
+			"priority": 1,
+			"mutex": 
+			[
+				
+			],
+
+			"category": "PLN"
+		},
```
linas: I mean, just to be clear, instead of writing json, you could just have written this:
```
(EvaluationLink
   (PredicateNode "default_cpolicy")
   (ListLink
       (ConceptNode "pln-rule-modus-ponens")
       (ConceptNode "priority:1")
       (ConcpetNode "mutex")
       (ConceptNode "category:PLN")
   ))
```
that way, you would not need a json file, nor a json file reader.

linas: Another very different and completely unrelated issue: a few weeks ago, there was a lengthly discussion about the relex2logic rules, which currently use priorities and mutually exclusive rule-sets.  There was considerable grumbling over this design, and I proposed a similar, but slightly different design.   I am somewhat disheartened to see that priority+mutex is showing up again, in new code.  It is likely to cause the same pain and heartache that R2L is complaining about, so I think it would be better to avoid it now.

Soooo... the slightly-different design is this.  Instead of priority, use "cost". A cost of zero is same as priority 1.  A cost of 1 is the same as priority 2. A cost of 0.5 is the same as a priority 1.5.  Costs are floating point.  Not important now, but will be very important later: costs are additive.  You will be able to add them together.

Next: instead of mutexes, use disjuncts. The disjuncts are stored separates from the rules, they are not part of a rule.  Instead, a collection of disjuncts look like this:
```
apply rule "modus ponens"
or apply rule "modus fashion"
or apply (rule "modus ice-cream" and rule "modus-pie")
or apply rule "modus fugit"
```
This gives you the choice of rules to apply at each stage, and they are, by definition mutually exclusive of each-other. That is what the "or" above is: its actually "exclusive-or".   Because each rule has a cost, you know what order the rules should be applied in.

This is a minor design change in the short run, and is almost completely identical to priority+mutex. However, in the long run, cost+disjunct is much more flexible and powerful and easier to use and manage.  So, please avoid priority+mutex.

misgeatgit: >I mean, just to be clear, instead of writing json, you could just have written this:
>
>(EvaluationLink
>   (PredicateNode "default_cpolicy")
 >  (ListLink
 >      (ConceptNode "pln-rule-modus-ponens")
 >      (ConceptNode "priority:1")
 >      (ConcpetNode "mutex")
 >      (ConceptNode "category:PLN")
 >  ))
>that way, you would not need a json file, nor a json file reader.

I'm less familiar with scheme but why do we need to store such ephemeral things in the atomspace?. Actually we can delete them after the reading, but wouldn't that also make the atomspace unnecessarily busy?

linas: in what way are they ephemeral? They seem to be designed to strongly affect how PLN works.  That's not ephemeral at all.

misgeatgit: I mean in my design the values are used to create the c++ Rule Objects.

misgeatgit:
I have made some changes. Can you take a look at it?

linas:
I'm not sure what you want me to do. This is a huge change ... it would take me many days to study it; and I don't have that kind of time right now.   If you have a specific question, then ask it. Otherwise, if you need to merge this, then you probably should: I won't object to a merge.  I trust it does more-or-less the right thing.

I expressed two concerns: on important, and one cosmetic.  The important one concerned the use of priorities and mutexes, which I beleive should be replaced by the use of costs and disjuncts.  It does not appear that you change that.

The cosmetic concern was the invention of yet another file format (json).  I don't see why the configuration could not have been done in either python, or in scheme, or in the opencog.conf file, or just a c++ file containing the configuration.  Why a fifth method is needed is just not clear to me. 

I suppose we can live with the json indefinitely, its not obviously harmful.  However, the priorities+mutexes issue is, I think, very important, and should not be avoided for long.

linas:
Anyway -- to summarize, you should merge.

misgeatgit:
Sorry for the large commit, I hope it won't be that big in the future as it gets mature. I will try to avoid the 'cosmetic' json at later stage. 
I can't merge this PR because I don't have a write access.

williampma:
Just wondering... why is the same set of commits at the top got repeated?

misgeatgit:
@williampma  which commits? may be I can squash them.

williampma:
The first 6 commits on the list got repeated with different SHA for some reason.

PR #1265, opened 2015-02-10
Surprisingness ii

shujingke:
Small fixes.

### DIFF
--- a/opencog/embodiment/ugly_male_soda-drinker_corpus.scm
+++ b/opencog/embodiment/ugly_male_soda-drinker_corpus.scm
@@ -1,4 +1,4 @@
-
+(setlocale LC_CTYPE "")
 
 (InheritanceLink (stv 1.000000 1.000000)
   (ConceptNode "Allen")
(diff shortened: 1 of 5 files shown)


opencog/embodiment/ugly_male_soda-drinker_corpus.scm
```diff
@@ -1,360 +1,360 @@
+(setlocale LC_CTYPE "")
```
linas: I'm pretty sure that this will cause conflicts and bad behavior. You really want the locale to be utf8  The default locale should be fine, already.

opencog/learning/PatternMiner/PatternMiner.cc
```diff
@@ -1707,148 +1701,149 @@ void PatternMiner::calculateSurprisingness( HTreeNode* HNode, AtomSpace *_fromAt
...
+////                cout << "For Super pattern: -------extended from a variable----------------- " << std::endl;
+////                cout << unifiedPatternToKeyString(curSuperRelation.extendedHTreeNode->pattern, atomSpace);
+
+//                float _min_surprisingness_II = 99999999.0f;
+
+//                float p_Ap = (float)curSuperRelation.extendedHTreeNode->count / atomspaceSizeFloat;
+
+//                // only calculate the connected subcomponets that also connected with patternE
+//                for(unsigned int subgramNum = 1;  subgramNum < gram; ++ subgramNum)
+//                {
```
linas: you know -- a much easier way of commenting out code is just to put an #if 0 ... #endif around it.

PR #1272, opened 2015-02-13
SuReal: use original LG connectors usage for word replacement

williampma:
This will make SuReal easier to use.  For example, given
```
(r2l "He drinks.")
```
it is now possible to generate `I eat.`, `John danced.`, etc.

I realized LG entries are actually a bit too forgiving in some cases.   For example, given
```
(r2l "The bird flew upward.")
```
then giving SuReal
```
(WordNode "Sam")
(WordNode "laughed")
(sureal (SetLink (EvaluationLink (PredicateNode "laughed") (ListLink (ConceptNode "Sam")))))
```
will generate
```
((the Sam laughed upward .))
```
because according to LG this is a valid sentence.... maybe when there are multiple "Sam" and the laughing is toward the sky...

# TODO
* update unit tests (done, in upcoming pull request)
* change to use native C++ code instead of scheme code with eval_h, eval_q

# To think about
* is it possible to have the microplanner not depends on using SuReal to test whether a "chunk" of atoms is "sayable"?  Because right now with SuReal doing more works, the microplanner calling SuReal repeatedly is quite slow.  It will be even slower when there is a huge atomspace for SuReal to pattern match against.

### DIFF
--- a/opencog/guile/SchemeEval.cc
+++ b/opencog/guile/SchemeEval.cc
@@ -813,6 +813,90 @@ void * SchemeEval::c_wrap_eval_h(void * p)
 	return self;
 }
 
+/**
+ * Evaluate a string containing a scheme expression, returning a HandleSeq.
+ * If an evaluation error occurs, then the error is logged to the log
+ * file, and an empty HandleSeq is returned.  If the result of evaluation
+ * is not a list, no error is logged, but an empty HandleSeq is still
+ * returned. If any item in the list is not a Handle, still no error is
+ * returned, but an empty HandleSeq is still returned. Otherwise, if the
+ * result of evaluation is a list of Handle, a HandleSeq is returned.
+ */
+HandleSeq SchemeEval::eval_q(const std::string& expr)
+{
+	if (_in_eval) {
+		SCM expr_str = scm_from_utf8_string(expr.c_str());
+		SCM rc = do_scm_eval(expr_str, recast_scm_eval_string);
+
+		if (!scm_list_p(rc))
+			return HandleSeq();
+
+		HandleSeq result;
+
+		while (rc != SCM_EOL)
+		{
+			Handle h = SchemeSmob::scm_to_handle(scm_car(rc));
+
+			if (h == Handle::UNDEFINED)
+				return HandleSeq();
+
+			result.push_back(h);
+			rc = scm_cdr(rc);
+		}
+
+		return result;
+	}
+
+#ifdef WORK_AROUND_GUILE_THREADING_BUG
+	thread_lock();
+#endif /* WORK_AROUND_GUILE_THREADING_BUG */
+
+	pexpr = &expr;
+	_in_eval = true;
+	scm_with_guile(c_wrap_eval_q, this);
+	_in_eval = false;
+
+#ifdef WORK_AROUND_GUILE_THREADING_BUG
+	thread_unlock();
+#endif /* WORK_AROUND_GUILE_THREADING_BUG */
+
+	return qanswer;
+}
+
+void * SchemeEval::c_wrap_eval_q(void * p)
+{
+	SchemeEval *self = (SchemeEval *) p;
+	// scm_from_utf8_string is lots faster than scm_from_locale_string
+	SCM expr_str = scm_from_utf8_string(self->pexpr->c_str());
+	SCM rc = self->do_scm_eval(expr_str, recast_scm_eval_string);
+
+	if (!scm_list_p(rc))
+	{
+		self->qanswer = HandleSeq();
+		return self;
+	}
+
+	HandleSeq result;
+
+	while (rc != SCM_EOL)
+	{
+		Handle h = SchemeSmob::scm_to_handle(scm_car(rc));
+
+		if (h == Handle::UNDEFINED)
+		{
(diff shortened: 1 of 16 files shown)


opencog/guile/SchemeEval.cc
```diff
@@ -813,6 +813,90 @@ void * SchemeEval::c_wrap_eval_h(void * p)
...
+/**
+ * Evaluate a string containing a scheme expression, returning a HandleSeq.
+ * If an evaluation error occurs, then the error is logged to the log
+ * file, and an empty HandleSeq is returned.  If the result of evaluation
+ * is not a list, no error is logged, but an empty HandleSeq is still
+ * returned. If any item in the list is not a Handle, still no error is
+ * returned, but an empty HandleSeq is still returned. Otherwise, if the
+ * result of evaluation is a list of Handle, a HandleSeq is returned.
+ */
+HandleSeq SchemeEval::eval_q(const std::string& expr)
```
linas: I really really really don't like this ... can't you just return a ListLink with all the handles in it?  That requires no new C code.

opencog/nlp/lg-dict/LGDictExpContainer.cc
```diff
@@ -199,47 +199,6 @@ void LGDictExpContainer::basic_normal_order()
...
-{
-    if (m_type == CONNECTOR_type)
-    {
-        if (m_string == "OPTIONAL")
-            return "(LgConnector (LgConnectorNode \"0\"))\n";
-
-        std::stringstream ss;
-        ss << "(LgConnector (LgConnectorNode ";
-        ss << "\"" << m_string << "\") ";
-        ss << "(LgConnDirNode \"" << m_direction << "\") ";
```
linas: ah, well, as previously discussed, you could have just said `atomspace->newNode(LG_CON_DIOR_NDOE, m_direction);` here.

opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -73,32 +80,60 @@ bool SuRealPMCB::variable_match(Handle &hPat, Handle &hSoln)
...
+
+    // no WordInstanceNode? reject!
+    if (hSolnWordInst == Handle::UNDEFINED)
         return false;
 
-    // get the neighbor of each WordNode within LgWordCset
-    HandleSeq qPatSet =_as->getNeighbors(hPatWordNode, false, true, LG_WORD_CSET, false);
-    HandleSeq qSolnSet = _as->getNeighbors(hSolnWordNode, false, true, LG_WORD_CSET, false);
+    // get the source connectors for the solution
+    std::string scmCode = "(word-inst-get-source-conn " + SchemeSmob::to_string(hSolnWordInst) + ")";
```
linas: to avoid having to use eval_q here, you could just say
```
std::string scmCode = "(ListLink (word-inst-get-source-conn " ...
 HandleSeq qTargetConns = m_eval->eval_h(scmCode)->getOutgoingSet();
```
Its a little more verbose ...

linas:
I'll merge this because I don't want to hold up your work, but would be a lot happier if there was no eval_q in the C++ code.

anitzkin:
re: "The Sam laughed upward"

Both of these errors are artifacts of the fact that Surreal currently
doesn't know any grammar and doesn't act in synergy with PLN.  In the fully
integrated system, it would recognize that a proper name is already a
definite noun, which cannot be modified by a determiner (unless the
system's goal is to say something unusal or funny for some reason, or
because there is an unusual context in which "Sam" is being re-purposed as
an indefinite noun for some reason).  And of course one can "laugh upward"
but the point here is that the system, should not in the long run be
inserting wods merely because they are present in a matching sentence
structure, but only because they represent one of the conceptual
sub-hypergraphs that has been sent to Surreal for expression.

I.e. don't expect to get it right purely through this sentence-matching
heuristic; keep in mind, that is only one of several processes that need to
go into producing the right expression.

opencog/nlp/sureal/SuRealPMCB.cc
williampma: I am curious, what is the reasoning of preferring to create a new `ListLink` in the atomspace than C code?

linas: Maintainability and simplicity. You added about 100 lines of code to do something that could be done in two.  In general, less code == better code.

The added lines of code are a cut-n-paste of existing code, with only a tiny change. i.e. the function is not "orthogonal", or "not normalized" aka "not in (ortho)normal form". Failure to normalize data (eg. sql tables), proofs, and algorithms is always considered to be bad form.  Its discussed on-line and in textbooks; data strcture textbooks usually have several chapters on normalization.
 
So, for example: in this case: if there is a future change to eval_h(), whoever makes that change needs to also change eval_q() -- so the amount of code to be maintained is doubled.   The testing effort is doubled, because a bugfix in one may or may not affect the other.  Both must be tested.  Because there are two blocks of code, they will typically start to diverge, as other users make changes to one, but not the other.  Do you think this won't happen? It does: you are not the first person who has tried to make the scheme bindings better by introducing minor changes, which then decay because other bug fixes by other people (none of them me!) are not correctly propagated throughout.  Eventually something stops working, and then someone removes the duplicated code ... etc.

PR #1273, opened 2015-02-13
NLP: Unit test update

williampma:


williampma:
This can be merged.  Most of it are just changing the test scm files, and no need to review.

PR #1277, opened 2015-02-17
lg-dict: moved some scm code to scm primitivies

williampma:
Putting `lg-conn-type-match?` and `lg-conn-linkable?` in LGDictModule as scm primitive in `(opencog nlp lg-dict)`.

Not sure where to put a C++ implementation of `word-inst-get-source-conn` yet.


opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -130,14 +131,15 @@ bool SuRealPMCB::variable_match(Handle &hPat, Handle &hSoln)
         // loop thru all connectors on both list
         while (not sourceConns.empty() && not targetConns.empty())
         {
-            // if we are repeating a multi-connector from source list
+            bool bResult;
+
             if (hMultiConn != Handle::UNDEFINED)
-                scmCode = "(lg-conn-linkable? " + SchemeSmob::to_string(hMultiConn) + " " + SchemeSmob::to_string(targetConns.front()) + ")";
+                bResult = lg_conn_linkable(hMultiConn, targetConns.front());
```
linas: I assume that you are replacing code in order to make it run faster?   I guess that makes sense for the pattern matcher, since any callback is going to be a performance bottleneck.

Someday someone should have a long chat with the guile people to understand what could be done to get a very high-speed interface between scheme and C.  I see several possibilities for the above, but none as fast as a native C++. (the new guile-2.2 VM seems really pretty fast)  The same issues will come up w/haskell, too, I guess.

williampma: > I assume that you are replacing code in order to make it run faster?

Yes, that is why.  Since this code is called for each connector on each disjunct on each node match, there could be several thousands scm call per node matches.  This change did make a difference.

PR #1278, opened 2015-02-17
modernize code

misgeatgit:
-Used C++11 range based loops and the auto specifier
-Replaced vector <'Handle'> syntax with HandleSeq

### DIFF
--- a/opencog/dynamics/attention/HebbianUpdatingAgent.cc
+++ b/opencog/dynamics/attention/HebbianUpdatingAgent.cc
@@ -95,9 +95,6 @@ void HebbianUpdatingAgent::run()
 
 void HebbianUpdatingAgent::hebbianUpdatingUpdate()
 {
-    std::vector<Handle> links;
-    std::back_insert_iterator< std::vector<Handle> > link_output(links);
-    std::vector<Handle>::const_iterator current_l;
 
     // tc affects the truthvalue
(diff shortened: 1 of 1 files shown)


opencog/dynamics/attention/HebbianUpdatingAgent.cc
```diff
@@ -209,13 +204,12 @@ void HebbianUpdatingAgent::hebbianUpdatingUpdate()
...
 
-std::vector<Handle>& HebbianUpdatingAgent::moveSourceToFront(std::vector<Handle> &outgoing)
+HandleSeq& HebbianUpdatingAgent::moveSourceToFront(HandleSeq &outgoing)
 {
     // find source, atom with positive norm_sti, and make first in outgoing
-    std::vector<Handle>::iterator outgoing_i;
     Handle theSource;
     bool foundTheSource = false;
-    for (outgoing_i = outgoing.begin(); outgoing_i < outgoing.end();) {
+    for (auto outgoing_i = outgoing.begin();outgoing_i != outgoing.end();) {
```
linas: why not just `for(Handle oh: outgoing)` like before?

misgeatgit: Because of line 219 and 224. Needs the iterator for erasure and insertion.

PR #1285, opened 2015-02-19
Remove cause of warnings from unused functions

inflector:
The functions in helpers.hpp (which comes from the ZeroMQ examples and is used in a few of the tests) were declared "static" instead of "inline". This has the effect of causing warnings when they are not used. Static is used when you don't want to allow external linkage to a function, so this warning is to let the developer know that they can remove that code.

In this case, the static is in a header file. In general, there is hardly ever reason to put a static function declaration in a header file.

Changing "static" to "inline" here, which does not affect semantics because there are no local static variable decorations in these functions, makes the warnings go away.

### DIFF
--- a/lib/zmq/zhelpers.hpp
+++ b/lib/zmq/zhelpers.hpp
@@ -62,7 +62,7 @@
 #endif
 
 //  Receive 0MQ string from socket and convert into string
-static std::string
+inline std::string
 s_recv (zmq::socket_t & socket) {
 
     zmq::message_t message;
@@ -72,7 +72,7 @@ s_recv (zmq::socket_t & socket) {
 }
 
 //  Convert string to 0MQ string and send to socket
-static bool
+inline bool
 s_send (zmq::socket_t & socket, const std::string & string) {
 
     zmq::message_t message(string.size());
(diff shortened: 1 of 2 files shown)


inflector:
Also suppressing warnings from Cython generated static functions in headers.

linas:
Normally, I would expect to get unused-function warnings after your change, because any c file that included the h file would now have those functions declared.  In addition, there will now be duplicate copies of these functions, in multiple files.

Thus, the usual solution, when using the inline keyword, is to say "static inline'.  This way, if the function is unused, you don't get the warning, and you also don't get the multiple copies.  Not clear why you aren't seeing this.

inflector:
I did a bit of research on this to make sure I was up to date and "inline" by itself doesn't trigger warnings for "-Wunused-function"  or "-Wall" when the inline functions are not used. Those warnings only apply to static functions (whether inline or not). Or at least that seemed to be the consensus.

linas:
Huh. OK. Compilers change.   FWIW, I think both LLVM and GCC are working on cross-module inlining so that would have to be one of the things that would change.

Issue #1191, opened 2015-01-11
Using python in the cogserver segfaults


linas:
Started the cogserver, and did this:
```
telnet localhost 17001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
opencog> py
Entering python shell; use ^D or a single . on a line by itself to exit.
py> print("ola")
```
This causes the cogserver to crash with a null-pointer dereference -- somehow the atomspace is not being set for python.
```
#1  0x00007ffff63585f6 in opencog::PythonEval::getPyAtomspace (this=0x6751e0, 
    atomspace=0x0)
    at /home/linas/src/novamente/src/opencog-git/opencog/cython/PythonEval.cc:142
#2  0x00007ffff63586a9 in opencog::PythonEval::init (this=0x6751e0)
    at /home/linas/src/novamente/src/opencog-git/opencog/cython/PythonEval.cc:120
#3  0x00007ffff63589dc in opencog::PythonEval::eval_expr (this=0x6751e0, 
    partial_expr=...)
    at /home/linas/src/novamente/src/opencog-git/opencog/cython/PythonEval.cc:422
#4  0x00007fffec8a5456 in opencog::GenericShell::do_eval (this=0x7fffd8000c10, 
    expr=...)
    at /home/linas/src/novamente/src/opencog-git/opencog/shell/GenericShell.cc:275
#5  0x00007fffec8a4bf0 in opencog::GenericShell::eval (this=0x7fffd8000c10, 
    expr=..., s=<optimized out>)
    at /home/linas/src/novamente/src/opencog-git/opencog/shell/GenericShell.cc:159
```

linas:
It occurs to me that this might be a variant of bug #1131

linas:
Huh. pull request #1286 does fix the stack trace that  was originally reported above.  Now, it crashes in python-internal code:
```
[2015-02-20 05:29:13:904] [ERROR] Caught signal 11 (Segmentation fault) on thread 140657821738752
	Stack Trace:
	2: basic_string.h:539	~basic_string()
	3: CogServerMain.cc:78	_Z7sighandi()
	4: ??:0	killpg()
	5: ??:0	PySys_GetObject()
	6: PythonEval.cc:129	  opencog::PythonEval::init()
	7: basic_string.h:1131	  std::string::assign(char const*)
	8: basic_string.h:539	~basic_string()
[...]

Issue #1131, opened 2014-11-29
Executing Python GroundedSchemaNode causes segfault


cosmoharrigan:
Executing a Python GroundedSchemeNode as suggested in the [wiki](http://wiki.opencog.org/w/ExecutionLink#Python_example) causes a segfault.

**guile> (cog-execute! (ExecutionLink (GroundedSchemaNode "py:print \"Hello world\n\"") (ListLink (ConceptNode "1") (ConceptNode "2"))))**
```
[2014-11-29 00:05:30:568] [ERROR] Caught signal 11 (Segmentation fault) on thread 
140708337927936
	Stack Trace:
	2: Logger.h:172	~Base()
	3: CogServerMain.cc:78	_Z7sighandi()
	4: ??:0	killpg()

[2014-11-29 00:05:33:732] [INFO] [PythonModule] destructor
Exception KeyError: KeyError(140708337927936,) in <module 'threading' from 
'/usr/lib/python2.7/threading.pyc'> ignored
```

(In response to inquiry by @rodsol & @bgoertzel)

linas:
The crash seen here is probably identical to the one that I currently see in bug #1191 and that one might be easier to debug.

cosmoharrigan:
@misgeatgit do you know if Python GroundedSchemaNodes are working now?

misgeatgit:
@cosmoharrigan No, it doesn't work. If its an urgent case, I might be able to continue debugging it.

misgeatgit:
...As I have mentioned early, one thing I noticed is the above function call is invalid since what the python eval does is call the function with the argument passed where the function name is considered as the name of the GroundedSchemaNode and arguments extracted from the the ListLink . (https://github.com/misgeatgit/opencog/blob/master/opencog/execution/ExecutionLink.cc#L110)
I will try it with another valid call. may be like 
(cog-execute! (ExecutionLink (GroundedSchemaNode "py:makeListLink") (ListLink (ConceptNode "1") (ConceptNode "2"))))
and define the function in some python file where the path can be resolved by PythonEval.cc

cosmoharrigan:
So, are you saying that it might work using the other, valid syntax, of a call? Have you tried that one? Thanks.

misgeatgit:
[...]

Issue #684, opened 2014-04-14
PLN Examples cause CogServer to crash


Alex-van-der-Peet:
Running the PLN Examples according to the instructions at the top of 

https://github.com/opencog/opencog/blob/master/tests/python/test_pln/pln_examples.py

Gives:

[2014-04-14 08:56:57:761] [ERROR] Caught signal 11 (Segmentation fault) on thread 140427848054528
	Stack Trace:
	2: /lib/x86_64-linux-gnu/libc.so.6 (x +0x364a0) [0x7fb80404e4a0]

[2014-04-14 08:57:00:817] [INFO] [PythonModule] destructor
Exception KeyError: KeyError(140427848054528,) in <module 'threading' from '/usr/lib/python2.7/threading.pyc'> ignored
[2014-04-14 08:57:00:825] [INFO] Terminating BenchmarkModule.

And causes the CogServer to shut down.

inflector:
After #1292, the above PLN examples no longer cause a crash. The test cases listed by @cosmoharrigan do work:

```
opencog> py-eval from opencog.atomspace import *
opencog> py-eval print "Hello"
Hello
opencog> 
```
The example from the top of the pln_examples.py file, however now reports a runtime error in the Python interpreter:
```
opencog> loadpy pln_examples
No subclasses of opencog.cogserver.MindAgent found.
No subclasses of opencog.cogserver.Request found.
opencog> py
Entering python shell; use ^D or a single . on a line by itself to exit.
py> from pln_examples import *
py> pln_examples = PLNExamples(ATOMSPACE)
py> pln_examples.run_pln_example('../tests/python/test_pln/scm/specific_rules/DeductionRule.scm')
../tests/python/test_pln/scm/specific_rules/DeductionRule.scm
[(InheritanceLink (av 0 0 0) (stv 1.000000 0.000000)
  (ConceptNode "A" (av 0 0 0) (stv 0.800000 0.800000)) ; [58]
  (ConceptNode "C" (av 0 0 0) (stv 0.470000 0.820000)) ; [60]
) ; [63]
]
[(ConceptNode "DeductionRule<InheritanceLink>" (av 0 0 0) (stv 1.000000 0.000000)) ; [68]
]
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "../tests/python/test_pln/pln_examples.py", line 112, in run_pln_example
    all_rules = AllRules(self.atomspace, chainer)
[...]

Issue #535, opened 2014-02-10
REST API in combination with Python shell produces segfault


cosmoharrigan:
Start the REST API: ```restapi.Start```
Enter the Python shell in the cogserver: ```py```
Run this command: ```from opencog.atomspace import *``` and a segfault occurs:
```
[ERROR] Caught signal 11 (Segmentation fault) on thread 139776596203264
	Stack Trace:
	2: ??:0	killpg()
	3: ??:0	PyImport_GetModuleDict()
	4: ??:0	PyImport_AddModule()
	5: ??:0	PyRun_SimpleStringFlags()
	6: PythonEval.cc:291	  opencog::PythonEval::apply_script(std::string const&)
	7: basic_string.h:578	  std::string::operator=(std::string&&)
	8: basic_string.h:288	  std::string::_M_data() const
	9: GenericShell.cc:157	  opencog::GenericShell::eval(std::string const&, opencog::ConsoleSocket*)
	10: ConsoleSocket.cc:232	  opencog::ConsoleSocket::OnLine(std::string const&)
	11: basic_string.h:288	  std::string::_M_data() const
	12: ??:0	thread_proxy()
	13: ??:0	start_thread()
	14: ??:0	__clone()
```

inflector:
This was fixed by: https://github.com/opencog/opencog/pull/1292

inflector:
@cosmoharrigan This bug should be closed after you confirm the above fix. I no longer get this error following the above instructions after the Pull Request #1292.

cosmoharrigan:
@inflector I pulled the latest code, and rebuilt, and got an exception when I tried to reproduce. Please see the session log below.

Shell # 1:
```
$ cd ~/opencog/build
$ ./opencog/server/cogserver
Using default config at lib/opencog.conf
Using default config at ../lib/opencog.conf
link-grammar: Info: Dictionary found at /usr/local/share/link-grammar/en/4.0.dict
Listening on port 17001
```

Shell # 2:
```
rlwrap telnet localhost 17001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.

opencog> restapi.Start
```

Shell # 1:
```
REST API is now running in a separate daemon thread.
 * Running on http://0.0.0.0:5000/
```

Shell # 2:
```
opencog> py
Entering python shell; use ^D or a single . on a line by itself to exit.

py> 2+2
Connection closed by foreign host.
```

Back to Shell # 1:
```
[...]

PR #1286, opened 2015-02-19
Fixed SEGFAULT in PyEval

inflector:
Fixed SEGFAULT accessing py_atomspace in PyEval by adding the required call for C APIs generated by Cython.

The C++ function: import_opencog__atomspace();
needed to be called before calling any of the functions defined as "api" in the file: atomspace_details.pyx
which is accessed in C++ through the header: atomspace_api.h

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -74,6 +74,7 @@ void PythonEval::init(void)
     eval_already_inited = true;
 
     global_python_init();
+    import_opencog__atomspace();
 
     logger().info("PythonEval::%s Initialising python evaluator.",
         __FUNCTION__);


linas:
I think this might also fix #1191  and also #1131

linas:
and also possibly #684

linas:
and also possibly #1147 ????

linas:
and possibly #535 ???

inflector:
This particular bug would cause PyEval to not work anywhere from what I can tell. So you are likely correct.

I"ll check those out and let you know if they're still here.

linas:
I just checked #1131 and #1191  they still crash, but now in a different place: inside of PySys_GetObject()  at PythonEval.cc:129

inflector:
I'd check the above again, as I think my last pull request #1292 will likely fix them.

PR #1292, opened 2015-02-20
Fixed bug with GIL not being grabbed in PyEval

inflector:


inflector:
This is the fix for #1191

PR #1301, opened 2015-02-22
Fix Python references, add PythonEval unit test

inflector:
Fixed many problems with Python object reference counting in PythonEval.

Added PythonEval unit test.

Added a throw of RuntimeError for bad function names in PythonEval::apply
Added a throw of RuntimeError for errors in Python code in PythonEval::apply_script.

### DIFF
--- a/tests/cython/CMakeLists.txt
+++ b/tests/cython/CMakeLists.txt
@@ -22,6 +22,22 @@ IF (HAVE_SERVER)
 ENDIF (HAVE_SERVER)
 
 
+# The PythonEvalUTest is derived from the CogServer class, and so 
+# cannot be tested unless the cogserver is actually being built!
+IF (HAVE_SERVER)
+	ADD_CXXTEST(PythonEvalUTest)
+	CONFIGURE_FILE(${CMAKE_SOURCE_DIR}/tests/cython/pyeval.conf
+		${PROJECT_BINARY_DIR}/tests/cython/pyeval.conf)
+
+	TARGET_LINK_LIBRARIES(PythonEvalUTest
+		server
+		cogutil
+	)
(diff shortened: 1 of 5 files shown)


tests/cython/CMakeLists.txt
```diff
@@ -22,6 +22,22 @@ IF (HAVE_SERVER)
 ENDIF (HAVE_SERVER)
 
 
+# The PythonEvalUTest is derived from the CogServer class, and so 
+# cannot be tested unless the cogserver is actually being built!
+IF (HAVE_SERVER)
```
linas: this probably should be HAVE_CYTHON, I'm guessing

linas: Oh,. never mind, I see you are using the server.  Many of the test cases use the server when they don't really need to -- its enough to create a bare atomspace, the rest of the server is typically un-needed.

linas:
p.s. after your code gets commited, it shouls  up in github under the "unknown user Curtis Micheal Faith" with email address luz@nova.home  You may want to fix your .git email address  

here for example: https://github.com/opencog/opencog/commit/65e96e7ae7488e293ec044feef4f2a64c20ae22c  top left corner

tests/cython/CMakeLists.txt
inflector: In this case, I wanted to create test case as close to bug #1191 as possible. Since that bug uses the Python Shell via the Cogserver, I figured following the lead of PythonModuleUTest was as close to the server path as I could get.

inflector:
Ah, thanks for the notice. It was set properly in .git but not in my preferences for the GitHub app on my Mac which I usually use for my commits because it's easier to check my change diffs over before the commit. Should be fine now. Please let me know if you see the same problem again.

PR #1309, opened 2015-02-23
Code shuffling

williampma:
Putting related codes closer together for easier tracking.


opencog/nlp/scm/processing-utils.scm
```diff
@@ -127,3 +133,160 @@
...
+		; relex-to-logic pipeline.
+		(define (eval-list a-list)
+			(if (= 1 (length a-list))
+				(begin (eval-string (list-ref a-list 0)) '())
+				(begin
+					(eval-string (list-ref a-list 0))
+					(let* ((parse-name (parse-str (list-ref a-list 0)))
+					       (parse-node (ParseNode parse-name)))
+;
+; XXX not really necessary at this point since we will only need the LG
```
AmeBel: would you pls adjust the spacing below?

williampma: Humm... it was on purpose because it's not really a commenting the code but instead is commenting about the commenting out of an entire block of code.

opencog/nlp/scm/processing-utils.scm
```diff
@@ -127,3 +133,160 @@
...
+	)
+
+	; ---------------------------------------------------------------------
+	; returns the name of a ParseNode if the string has a ParseNode entry.
+	; returns "sentence@dbce9f0d-8b8a-4ea7-a8c8-d69abf05f810_parse_1" from the string
+	; "84590fb2e46c (ParseNode \"sentence@dbce9f0d-8b8a-4ea7-a8c8-d69abf05f810_parse_1\""
+	(define (parse-str a-string)
+		(substring (match:substring
+			(string-match "ParseNode \"sentence@[[:alnum:]_-]*" a-string)) 11)
+	)
```
linas: ??? Why? why would you need this?

If you want to test if an atom exists, without creating it, you can do this:  `(cog-node 'ParseNode "sentence@abc")`  and this will return #t if that atom exists, else it returns #f

Doing string compares and regexes seems to be extremely inefficient, and counter to the design goals for the atomspace.  If you have to do string compares to make something work, then the design of the atomspace structures are somehow wrong or incomplete ... it should be enough to ask for atom types, their names, their outoging sets, and that's all.  String compares should never be needed.  That actual string contents should almost never matter.

opencog/nlp/scm/processing-utils.scm
linas: I like William's style. Bad indentation is very eye-catching, and is a great way to quickly spot trouble areas in the code.  Whenever I write temporary code that I need to eventually fix/change/remove, I indent it badly, on purpose. That makes it easier to find later one.  (This is one reason why python sucks -- you cannot indent badly on purpose, which destroys readability and makes python hard to program in ...)

linas:
p.s. Can you now run the r2l code paths, without using the cogserver?  Not urgent, I'm just curious.

opencog/nlp/scm/processing-utils.scm
williampma: Well.. these codes are not really working on the atomspace, but is working on the string emitted by the RelEx server.

I believe something like this was discussed already when @AmeBel first pushed these codes at https://github.com/opencog/opencog/pull/920

On the other hand, unrelated to these codes, atom's name string comparison was always part of RelEx2Logic's incomplete design.  It is certainly necessary to fix, but needs well structured steps & goals.  R2L is still quite chaotic.

williampma:
About running without cogserver... if only running the language parsing part (without the language generation), then I think it should be possible, since all the codes are just plain scheme code.  Lots of files need to be loaded, however.

With SuReal, however, the C++ module will need to be loaded separately.

linas:
which c++ module?  Reminder: both pattern matcher and database require the C++ modules to be loaded: if you set the LD_LIBRARY_PATH correctly .. but you know this already.  (well, in case you forgot, /examples/guile and the actual loading done here   https://github.com/opencog/opencog/blob/master/opencog/scm/opencog/persist.scm for one c++ library -- note its an ordinary library, not an opencog module)

williampma:
I believe for SuReal, libSuRealModule needs to be loaded with `(load-extension ...)` just like what you did with query and persist.

Anything c++ library that add to `(opencog extension)` will need to do the same right?

linas:
Well, 

-- for stand-alone scheme, you don't load SuRealModule, you load libsureal.so  and you would have to say `(load-extension "libsureal" "some_sureal_init_func")`  in your scheme code.  That's how the C++ gets loaded (or should be getting loaded, for the stand-alone case).

-- Because the same code also has to work with cogserver, I think that (opencog extension) will "magically" work with the cogserver, as long as the cogserver loads SuRealModule -- but in the stand-alone cose, there is nothing in SchemePrimitive that tries to load a shared library  .. and it should stay that way.  You should crate a /opencog/scm/opencog/sureal.scm which has two lines of code in it: (define-module (opencog sureal)) and `(load-extension "libsureal" "sureal_init")`  and add sureal_init to libsureal.so -- its also just one line of code.

That is, I don't think that `(opencog extension)` can work automatically outside of the cogserver, because there is no way of loading the c++ libts without saying `(load-extension "libsureal" "sureal_init")` somwhere.

Issue #1295, opened 2015-02-21
Implement flexible merging mechanism for TruthValues


bgoertzel:
What do we do when merging a TV of type X into a TV of type Y?    This is hairier...

I think it should be fair to assume, for the time being, that two TV types being assigned to the same Atom type, should be mutually interconvertible (perhaps with some loss of information).   If we assume this, then when merging a TV tv1 of type X into a TV tv2 of type Y, we can assume there exist methods to cast tv1 into type Y, or to cast tv2 into type X

So, suppose we have an Atom A with a TV tv1 of type X.  Suppose we want to merge a new TV tv2 of type Y into A's truth value....  So we are calling some method like

tv1.merge(tv2)

One way to look at this is

tv1.merge(tv2, typeoption = old | new | default | stronger | weaker)

If typeoption=old, then the type of the resulting TV is tv1

If typeoption=new, then the type of the resulting TV is tv2

If typeoption=dominant, then the type of the resulting TV is whatever is specified in the merge code as the default for merging a TV of type X and a TV of type Y (this is going to be either X or Y)

For instance, for merging a SimpleTV with an IndefiniteTV, the default might be IndefiniteTV

If typeoption=stronger then the type of the resulting TV is whichever of X or Y is specified as stronger

If typeoption=weaker then the type of the resulting TV is whichever of X or Y is specified as weaker

What I mean by "stronger vs. weaker" here is e.g., SimpleProbabilisticTV is weaker than IndefiniteTV, because an IndefiniteTV can exactly specify a SimpleProbabilisticTV, but a Simple ProbabilisticTV only specified an IndefiniteTV with a lot of heuristic fudging.

...

This kind of approach would require specifying somewhere or other, the strength ordering and default ordering of each pair of commensurable TV types (where "commensurable" means that the two TV types are applicable to the same Atom types)...

It seems to me this would give enough flexibility for different cognitive processes to do what they wish with TV types....   And if someone doesn't want to worry about it, they can call

tv1.merge(tv2) =tv1.merge(tv2, typeoption=default)

and then magic will happen and hopefully the default choices won't cause any egregious stupidity ;p

I'm not sure we want to require that all TruthValues partake in this merge-option mechanism.  But all PLN truth values should, as should CountTruthValue.

linas:
This work was completed in pull request #1312

PR #1310, opened 2015-02-23
Preparatory work for flexible TV merge 

misgeatgit:
Addressing  https://github.com/opencog/opencog/issues/1295

### DIFF
--- a/opencog/atomspace/CountTruthValue.cc
+++ b/opencog/atomspace/CountTruthValue.cc
@@ -34,20 +34,20 @@
 
 using namespace opencog;
 
-CountTruthValue::CountTruthValue(strength_t m, confidence_t n, count_t c)
+CountTruthValue::CountTruthValue(strength_t m, confidence_t n, count_t c):TruthValue(COUNT_TRUTH_VALUE)
 {
     mean = m;
     confidence = n;
     count = c;
 }
 
-CountTruthValue::CountTruthValue(const TruthValue& source)
+CountTruthValue::CountTruthValue(const TruthValue& source):TruthValue(COUNT_TRUTH_VALUE)
 {
     mean = source.getMean();
     confidence = source.getConfidence();
     count = source.getCount();
 }
-CountTruthValue::CountTruthValue(CountTruthValue const& source)
+CountTruthValue::CountTruthValue(CountTruthValue const& source):TruthValue(COUNT_TRUTH_VALUE)
 {
     mean = source.mean;
     confidence = source.confidence;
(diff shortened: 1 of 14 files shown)


opencog/atomspace/CountTruthValue.cc
```diff
@@ -34,20 +34,20 @@
 
 using namespace opencog;
 
-CountTruthValue::CountTruthValue(strength_t m, confidence_t n, count_t c)
+CountTruthValue::CountTruthValue(strength_t m, confidence_t n, count_t c):TruthValue(COUNT_TRUTH_VALUE)
```
linas: Please don't do this. This makes the TV larger than 8 bytes. There is no reason to store a compile time constant as if it were some dynamic, run-time variable.

opencog/atomspace/TruthValue.cc
```diff
@@ -94,3 +94,13 @@ bool TruthValue::isDefaultTV() const
...
 }
+
+TruthValueType TruthValue::getType() const
+{
+	return tv_type_;
+}
+
+void TruthValue::setType(TruthValueType t)
+{
+	tv_type_ = t;
```
linas: This is deeply and fundmentally wrong.  The different C++ classes have different numbers or members (simpe has 2, count has 3) and different merge methods. The numbers all have different meanings in the different classes. Changing the type is just .. impossible.

linas:
I can't merge this as written.  I have nothing against MergeOption (although I wish it was called MergeStyle).  However the change to the type is deeply flawed and must be removed.

opencog/atomspace/TruthValue.cc
misgeatgit: Hmm, I don't have any opposition about the optimization issue, so this PR might be turned down for that reason.If I have understood you correctly, you are in favor of changing the signature of the merge method (besides the naming issue) and oppose every other changes introduced. Right?

linas: Yes!  (and change the name to MergeStyle or something like that)

linas: BTW, I added Fuzzy and Probabilistic just a few days ago, for exactly this reason. The merge funcs for fuzzy and probabilistic are completely wrong, and need to be replaced by the correct formulas from the pln book.  The guts of SimpleTV merge func is wrong, as well.   Best to leave CountTV alone for now; if you want to change CountTV, then use either Probabilistic, or create a new TV, say, "FrequencyTruthValue"

PR #1312, opened 2015-02-24
Change merge function signature in order to support alternative TV merge

misgeatgit:
A modification to https://github.com/opencog/opencog/pull/1310
Addressing issue https://github.com/opencog/opencog/issues/1295

### DIFF
--- a/opencog/atomspace/TruthValue.h
+++ b/opencog/atomspace/TruthValue.h
@@ -71,6 +71,17 @@ enum TruthValueType
     NUMBER_OF_TRUTH_VALUE_TYPES
 };
 
+//Flags for alternative ways of merging two different truth value types
+//as described in https://github.com/opencog/opencog/issues/1295
+enum TVMergeStyle
+{
+    USE_OLDER_TV_TYPE,
+    USE_NEWER_TV_TYPE,
+    USE_STRONGER_TV_TYPE,
+    USE_WEAKER_TV_TYPE,
+    DEFAULT
+};
+
 class TruthValue;
 typedef std::shared_ptr<TruthValue> TruthValuePtr;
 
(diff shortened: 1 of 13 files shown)


misgeatgit:
Now, we can support return different truth value types in the merge function based on the value of the second argument in the merge function.

opencog/atomspace/TruthValue.h
```diff
@@ -71,6 +71,17 @@ enum TruthValueType
     NUMBER_OF_TRUTH_VALUE_TYPES
 };
 
+//Flags for alternative ways of merging two different truth value types
+//as described in https://github.com/opencog/opencog/issues/1295
+enum TVMergeStyle
+{
+    USE_OLDER_TV_TYPE,
+    USE_NEWER_TV_TYPE,
```
linas: well, this is mis-named: merge does not change the TV type, it changes the TV.   The type of the TV will usually not change.

So really, it  should be this:
USE_OLDER_TV_TYPE  is KEEP_CURRENT_TV  // no-op, no change, do nothing return current TV
USE_NEWER_TV_TYPE is REPLACE_TV i // replace current TV with given TV  i.e. return given TV
USE_STRONGER_TV_TYPE, is STRONGER_TV // of the two, pick the stronger TV  which is what the current code does, and it is what Ben doesn't like

So none of these merge strategies do what PLN actually needs.

linas:
Also: this comment is confusing:  "Now, we can support return different truth value types in the merge function" -- this has NOTHING to do with types.  You could merge different types last week, last year, no API changes were needed for that.  Some of the code already knows how to merge differrent types.   That is not the problem.

The problem is that PLN needs to use different arithmetical formulas in different situations, and the flags indicate which *formulas* to use.  The handling of types is automatic,

opencog/atomspace/TruthValue.h
bgoertzel: > USE_STRONGER_TV_TYPE, is STRONGER_TV // of the two, pick the stronger TV
> which is what the current code does, and it is what Ben doesn't like
>


Actually my problem wasn't about picking the stronger TYPE, but rather
about the way the simpleTV merge function chose the TV with the highest
confidence rather than doing some sort of confidence-weighted
combination...

PR #1314, opened 2015-02-24
Baby steps for Haskell support

linas:
Attempt (and fail) to compile a hello-world  Haskell file with current CMake infrastructure.

Lessons: 
1) Getting Haskell and CMake to co-exist will be hard.
2) In the long run, will almost surely have to use Cabal for building Haskell 
3) Getting Cabal and CMake to peacefully coexist will be a major task.


bgoertzel:
Cabal supports a custom build mode, so maybe you can just run cabal and
then invoke Cmake at an appropriate point in the cabal build process?

But I'm no expert on this stuff... cabal kinda sucks, but is better than
not having it ;p ...

On Wed, Feb 25, 2015 at 4:24 AM, Linas Vepštas <notifications@github.com>
wrote:

> Merged #1314 <https://github.com/opencog/opencog/pull/1314>.
>
> —
> <https://github.com/opencog/opencog/pull/1314#event-241218950>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

PR #1318, opened 2015-02-25
New NLP pipeline entry point for node count incrementing

williampma:
The PTV count increment doesn't work yet since R2L rule-helpers are rewriting each node with a new STV everytime.  They need to be removed.

* Changed `(r2l ...)` to `(r2l-parse ...)` to notify ppl of new entry point
* `(nlp-parse ...)` is the new entry point for parsing text, it will return the sentence-nodes

### DIFF
--- a/opencog/nlp/learn/link-pipeline.scm
+++ b/opencog/nlp/learn/link-pipeline.scm
@@ -128,12 +128,7 @@
 ; it parsed, and then updates the counts for the observed words and word
 ; pairs.
 (define (observe-text plain-text)
-	(begin
-		(relex-parse plain-text)  ;; send plain-text to server
-		(update-link-counts (get-new-parsed-sentences))
-		(release-new-parsed-sents)
-		(delete-sentences)
-	)
+	(update-link-counts (nlp-parse plain-text))
 )
 
 ; ---------------------------------------------------------------------
--- a/opencog/nlp/scm/processing-utils.scm
+++ b/opencog/nlp/scm/processing-utils.scm
@@ -301,21 +301,6 @@
 	(r2l-parse plain-text)
 	
 	(let ((sent-nodes (get-new-parsed-sentences)))
-		; increment the WordNode's count value
-		(parallel-map-parses
-			(lambda (p)
-				(map-word-instances
-					(lambda (wi)
-						(map-word-node
-							(lambda (w) (cog-atom-incr w 1))
-							wi
-						)
-					)
-					p
-				)
-			)
-			sent-nodes
-		)
 		; increment the R2L's node count value
 		(parallel-map-parses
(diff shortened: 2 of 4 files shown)


opencog/nlp/scm/processing-utils.scm
```diff
@@ -21,26 +21,20 @@
    )
 )
 
-; -----------------------------------------------------------------------
-; global vars:
-; new-sent anchor points at the node to which all new sentences are connected
-;
-(define *new-parsed-sent-anchor* (AnchorNode "# New Parsed Sentence" (stv 1 1)))
```
linas: There are two reasons to not remove this:
1) if someone changes the text "# New Parsed Sentence" in one place, they have to remember to do a global search-n-replace everywhere.   So, this is the same reason one avoids literals in C++ code.
2) there is some very minor performance boost from not having to evaluate this expression every time it is used.

opencog/nlp/learn/link-pipeline.scm
```diff
@@ -128,12 +128,7 @@
 ; it parsed, and then updates the counts for the observed words and word
 ; pairs.
 (define (observe-text plain-text)
-	(begin
-		(relex-parse plain-text)  ;; send plain-text to server
-		(update-link-counts (get-new-parsed-sentences))
-		(release-new-parsed-sents)
-		(delete-sentences)
-	)
+	(update-link-counts (nlp-parse plain-text))
```
linas: Please revert this. This change almoost surely breaks langauge learning.  You are making teh faulty assumption that learning uses Englsih .. it doesn't, it uses the ANY and AMY langauges, which do not have either relex nor r2l rules in them.  There's no relex because .. its not english.

linas: Also, be aware that the language learning code is doing very non-standard counting; it has its own independent system based on entropy, which niether probabilistic nor fuzzy tv know about. Also, the merge function for these will break the entropy/mutial information mechanism, and so that cannot be used.

linas: never mind, I reverted myself

opencog/nlp/scm/processing-utils.scm
linas: BTW *name* is a scheme convention that indicates that 'name' is a global

opencog/nlp/learn/link-pipeline.scm
linas: Oh, and one more note:  `(update-link-counts` above is counting word-pairs, which is something that your code does not do or need to do.  It also updates the counts in the SQL database, which, again, your code does not do, and probably(?) does not need to do.  In short, these are very different processing paths.

PR #1322, opened 2015-02-27
SuReal & LGDict code splitting

williampma:
Split the scheme primitives code out of the OpencCog Module code, so libsureal and liblg-dict will contain the scheme primitive defintions.

Couldn't get `load-extension` to work with this yet... keep saying `file not found` even though LTDL_LIBRARY_PATH is set correctly.


linas:
Fixed in 1547f01 You needed to add nlp-types to TARGET_LINK_LIBRARIES  -- the file-not-found message is misleading, it really means there are unresolved symbols in the so, and the dynamic linker cannot find the file that resolves those symbols.

williampma:
Oh, I see!  Thanks!

Issue #1307, opened 2015-02-22
JSON Persist missing FUZZY_TRUTH_VALUE and PROBABILISTIC_TRUTH_VALUE


inflector:
From the buildbot log:

`..build/opencog/persist/zmq/events/AtomSpacePublisherModule.cc:336:12: warning: enumeration value ‘FUZZY_TRUTH_VALUE’ not handled in switch [-Wswitch]`
`..build/opencog/persist/zmq/events/AtomSpacePublisherModule.cc:336:12: warning: enumeration value ‘PROBABILISTIC_TRUTH_VALUE’ not handled in switch [-Wswitch]`

It looks like this warning exposes a genuine bug, from AtomSpacePublisherModule.cc:

```
Object AtomSpacePublisherModule::tvToJSON(TruthValuePtr tvp)
{
    Object json;
    Object jsonDetails;

    switch (tvp->getType()) {
        case SIMPLE_TRUTH_VALUE: {
            json.push_back(Pair("type", "simple"));
            jsonDetails.push_back(Pair("strength", tvp->getMean()));
            jsonDetails.push_back(Pair("count", tvp->getCount()));
            jsonDetails.push_back(Pair("confidence", tvp->getConfidence()));
            json.push_back(Pair("details", jsonDetails));
            break;
        }

        case COUNT_TRUTH_VALUE: {
            json.push_back(Pair("type", "count"));
            jsonDetails.push_back(Pair("strength", tvp->getMean()));
            jsonDetails.push_back(Pair("count", tvp->getCount()));
            jsonDetails.push_back(Pair("confidence", tvp->getConfidence()));
            json.push_back(Pair("details", jsonDetails));
            break;
        }

        case INDEFINITE_TRUTH_VALUE: {
            IndefiniteTruthValuePtr itv = IndefiniteTVCast(tvp);
            json.push_back(Pair("type", "indefinite"));
            jsonDetails.push_back(Pair("strength", itv->getMean()));
            jsonDetails.push_back(Pair("L", itv->getL()));
            jsonDetails.push_back(Pair("U", itv->getU()));
            jsonDetails.push_back(Pair("confidence", itv->getConfidenceLevel()));
            jsonDetails.push_back(Pair("diff", itv->getDiff()));
            jsonDetails.push_back(Pair("symmetric", itv->isSymmetric()));
            json.push_back(Pair("details", jsonDetails));
            break;
        }

        case NULL_TRUTH_VALUE:
[...]

PR #1323, opened 2015-02-27
Silent warning mentioned on #1307

williampma:


opencog/persist/zmq/events/AtomSpacePublisherModule.cc
```diff
@@ -365,6 +367,24 @@ Object AtomSpacePublisherModule::tvToJSON(TruthValuePtr tvp)
...
+            jsonDetails.push_back(Pair("confidence", tvp->getConfidence()));
+            json.push_back(Pair("details", jsonDetails));
+            break;
+        }
+
+        case FUZZY_TRUTH_VALUE: {
+            json.push_back(Pair("type", "fuzzy"));
+            jsonDetails.push_back(Pair("strength", tvp->getMean()));
+            jsonDetails.push_back(Pair("count", tvp->getCount()));
+            jsonDetails.push_back(Pair("confidence", tvp->getConfidence()));
```
misgeatgit: Fuzzy truth value  is sopposed to store only strength and count. https://github.com/opencog/opencog/blob/master/opencog/atomspace/FuzzyTruthValue.h#L41

williampma: It is suppose to store only strength and conf.

But there's also a count KKK that starts at 800, but the count is actually calculated from the conf.

Anyway, all TV type can output all 3 values, even STV.

williampma: P.S.  I don't know why STV and FTV's constructor in C++ takes count, when the construction from Scheme takes confidence value.  Even the output is in confidence value.

linas:
I am not sure how many float-point numbers Fuzzy and Probabilistic are "supposed" to have -- either 2 or 3. It currently appear that having 3 might be better (strength, confidence, count) but that would require that some agent or process or function would look at the count every now and then, and update the confidence to match.

By contrast, the current SimpleTV does this "automatically" if you increment the count by 1, then it adds 800, uses that weird formula and comes up with a confidence. I always thought the 800 was arbitrary and hacky, but Ben has an argument that says that 800 is "reasonably OK for most cases."

So: I'm not sure which strategy is preferred: both fuzzy and probabilistic could use two values +800, or they could use 3 values, and some other approach.  You will have to talk to Ben and/or decide what the best solution is.  (I myself don't know)

PR #1325, opened 2015-02-28
Completely reworked Python initialization 

inflector:
Moved all Python initialization to a single place.
Removed some unnecessary GIL lock calls.
Separated out PythonEvalUTest into two parts:
   one that tests PythonEval separate from the CogServer
   and the other that tests “py-eval” through the server.
Fixed problems with Python C API borrowed references that were causing crashes in more than one Python finalize / initialize cycle.
Fixed some problems with multiple server instances in tests that was causing a crash on CogServer destruction.
Fixed a crash in the NetworkServer threading code on teardown.

These changes fix bug #535.

### DIFF
--- a/opencog/benchmark/AtomSpaceBenchmark.cc
+++ b/opencog/benchmark/AtomSpaceBenchmark.cc
@@ -327,9 +327,9 @@ void AtomSpaceBenchmark::startBenchmark(int numThreads)
             if (pymo == NULL) pymo = new PythonModule(*cogs);
             pymo->init();
             asp = &cogs->getAtomSpace();
-            pyev = &PythonEval::instance(asp);
+            pyev = &PythonEval::instance();
 
-            // And now ... create an instance of the atomspace.
+            // And now ... create a Python instance of the atomspace.
             // Pass in the raw C++ atomspace address into cython.
             // Kind-of tacky, but I don't see any better way.
             // (We must do this because otherwise, the benchmark would
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -1,8 +1,12 @@
 /*
  * @file opencog/cython/PythonEval.cc
- * @author Zhenhua Cai <czhedu@gmail.com> Ramin Barati <rekino@gmail.com>
+ * @author Zhenhua Cai <czhedu@gmail.com>
+ *         Ramin Barati <rekino@gmail.com>
  *         Keyvan Mir Mohammad Sadeghi <keyvan@opencog.org>
+ *         Curtis Faith <curtis.m.faith@gmail.com>
  * @date 2011-09-20
+ *
+ * @todo When can we remove the singleton instance?
  *
  * Reference:
  *   http://www.linuxjournal.com/article/3641?page=0,2
@@ -26,6 +30,7 @@
 
 #include <boost/filesystem/operations.hpp>
 
+#include <opencog/util/Config.h>
 #include <opencog/util/exceptions.h>
 #include <opencog/util/Logger.h>
 #include <opencog/util/misc.h>
@@ -34,7 +39,10 @@
 #include <opencog/server/CogServer.h>
 
 #include "PythonEval.h"
+
 #include "opencog/atomspace_api.h"
+#include "opencog/agent_finder_types.h"
+#include "opencog/agent_finder_api.h"
 
 using std::string;
 using std::vector;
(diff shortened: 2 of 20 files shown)


opencog/benchmark/AtomSpaceBenchmark.cc
```diff
@@ -327,9 +327,9 @@ void AtomSpaceBenchmark::startBenchmark(int numThreads)
             if (pymo == NULL) pymo = new PythonModule(*cogs);
             pymo->init();
             asp = &cogs->getAtomSpace();
-            pyev = &PythonEval::instance(asp);
```
linas: python eval should work with the atompsace that it is given. It cannot assume a default atomspace, there may be several.  Yes I understand the current design might not be able to do this, but I thought I should mention it.

opencog/cython/PythonEval.cc
```diff
@@ -44,60 +52,239 @@ using namespace opencog;
...
+    #ifndef WIN32
+    "/usr/local/share/opencog/python",
+    "/usr/share/opencog/python",
+    #endif // !WIN32
+    NULL
+};
+
+
 PythonEval* PythonEval::singletonInstance = NULL;
+AtomSpace* PythonEval::singletonAtomSpace = NULL;
```
linas: again, in the long-run, the atomspace can't be a singleton, although, short-term, that's OK-ish.  Other atomspaces are expected to hold temporaries of various sorts, or to hold special-purpose subsets of the data to avoid clutter.  Its still rather experimental, though.

opencog/cython/PythonEval.cc
```diff
@@ -44,60 +52,239 @@ using namespace opencog;
...
+            logger().debug("    %2d > %s", pathIndex, sysPathCString);
+            // NOTE: PyList_GetItem returns borrowed reference so don't do this:
+            // Py_DECREF(pySysPathLine);
+        } 
+    }
+
+    // Initialize the auto-generated Cython api. Do this AFTER the python
+    // sys.path is updated so the imports can find the cython modules.
+    import_opencog__atomspace();
+    import_opencog__agent_finder();
```
linas: You may need agent-finder for now, but in the long run, I'd like to see a version that doesn't drag agents into the mix, because the long-term strategy is to not use agents -- they're "deprecated"

opencog/cython/PythonEval.cc
bgoertzel: Well, right now Shujing's pattern miner relies on there being 2 Atomspaces
working togeter

So, if the python interface assumes only one Atomspace, then the python
interface can't be used to work with the pattern miner

This seems undesirable, thus I suggest the python interface should not
assume a singleton Atomspace

-- Ben

On Sat, Feb 28, 2015 at 12:40 PM, Linas Vepštas <notifications@github.com>
wrote:

> In opencog/cython/PythonEval.cc
> <https://github.com/opencog/opencog/pull/1325#discussion_r25554330>:
>
> >  PythonEval* PythonEval::singletonInstance = NULL;
> > +AtomSpace* PythonEval::singletonAtomSpace = NULL;
>
> again, in the long-run, the atomspace can't be a singleton, although,
> short-term, that's OK-ish. Other atomspaces are expected to hold
> temporaries of various sorts, or to hold special-purpose subsets of the
> data to avoid clutter. Its still rather experimental, though.
>
> —
> <https://github.com/opencog/opencog/pull/1325/files#r25554330>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

opencog/server/BaseServer.h
```diff
@@ -65,8 +65,41 @@ class BaseServer
...
+// problems with the way this is being used, in some cases, as it is confusing
+// that there are examples of both stack based and factory function singleton
+// server creation in the tests and examples. For example, I saw some code
+// that did this:
+//
+// 	  void testAddPunishmentPredicate() {
+//        server(SpaceTimeCogServer::createInstance);  << creates one instance
+//        SpaceTimeCogServer srv;                      << creates another
+//        AtomSpace& atomSpace = srv.getAtomSpace();
+//
```
linas: Yes, all this needs fixing.  Its not at all clear what the correct fix should be.  The space server may need to be replaced by or integrated with a ROS-based space server, so that spatial and motion-planning information can be shared with robot platforms. This is still wildly unclear to me.

The Embodiment server inherits from the CogServer, but I don't know of any reason that it should. I don't see how it derives any benefit from doing so. It seems like some old arbitrary design decision.  Again, its not clear to me what the correct fix should be.

bgoertzel: The whole design of the CogServer, SpaceTimeCogServer, etc., needs to be
radically changed, as we've discussed before

However, this is a big job in itself, and the fixing of the python
interface shouldn't wait for this, obviously...

On Sat, Feb 28, 2015 at 12:51 PM, Linas Vepštas <notifications@github.com>
wrote:

> In opencog/server/BaseServer.h
> <https://github.com/opencog/opencog/pull/1325#discussion_r25554388>:
>
> > @@ -65,8 +65,41 @@ class BaseServer
> >      virtual ~BaseServer(void);
> >  }; // class
> >
> > -// singleton instance (following meyer's design pattern)
> > +// Singleton instance (following meyer's design pattern) NOTE: There are
> > +// problems with the way this is being used, in some cases, as it is confusing
> > +// that there are examples of both stack based and factory function singleton
> > +// server creation in the tests and examples. For example, I saw some code
> > +// that did this:
> > +//
> > +// 	  void testAddPunishmentPredicate() {
> > +//        server(SpaceTimeCogServer::createInstance);  << creates one instance
> > +//        SpaceTimeCogServer srv;                      << creates another
> > +//        AtomSpace& atomSpace = srv.getAtomSpace();
> > +//
>
> Yes, all this needs fixing. Its not at all clear what the correct fix
> should be. The space server may need to be replaced by or integrated with a
> ROS-based space server, so that spatial and motion-planning information can
> be shared with robot platforms. This is still wildly unclear to me.
>
> The Embodiment server inherits from the CogServer, but I don't know of any
> reason that it should. I don't see how it derives any benefit from doing
> so. It seems like some old arbitrary design decision. Again, its not clear
> to me what the correct fix should be.
>
> —
> <https://github.com/opencog/opencog/pull/1325/files#r25554388>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

tests/cython/PyEvalUTest.cxxtest
```diff
@@ -0,0 +1,122 @@
+#include <string>
```
linas: Missing legal boilerplate

linas:
Thanks Curtis!  Major piece of work, much appreciated!

inflector:
My pleasure. A few more bugs and I'll start to work on expanding the
bindings and making them easier to work with.

PR #1324, opened 2015-02-27
Fuzzy hypergraph matching

leungmanhin:
Two scheme primitives are added:

**cog-fuzzy**: Find similar hypergraphs in the atomspace
Basic example:
(cog-fuzzy (ListLink (ConceptNode "A") (ConceptNode "B")))

Possible result:
(ListLink (ConceptNode "A") (ConceptNode "C"))

**get-similar-sentences**: Find similar sentences in the atomspace
Basic example:
(get-similar-sentences "What did Pete eat?")

Possible result:
(Pete ate apples .)

PR #1328, opened 2015-02-28
Simplify fuzzy match scheme binding

linas:
Follow up to pull request #1324.  Its easier for me to do this than to explain it.


leungmanhin:
Thanks Linas.

PR #1329, opened 2015-02-28
Added dependency on agent_finder to PythonEval

inflector:


inflector:
This should fix the failure of `make all` on the buildbot. It appears to be stuck in a failure loop which it can't recover from because this dependency was missing and it can't find the auto-generated agent_finder_api.h file because Cython isn't being run on agent_finder.pyx before the compile for PythonEval.cc without this fix.

See: 
http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/257/steps/compile/logs/stdio

inflector:
Doing a `make all` test first gives a cyclical dependency problem without an additional change.

Issue #1279, opened 2015-02-18
Merge function in SimpleTruthValue.cc does not implement the right math


bgoertzel:
Currently the merge function for SimpleTruthValue appears to merge old and new TVs via choosing the one with the highest confidence.   This is not a good formula.   A host of better, simple formulas for this are given in the PLN book.   See e.g. section 5.10.2 where the formula for merging (s1, n1) with (s2,n2) is given as:

s = s1 w1 + s2 w2, where w1 = n1 /(n1 +n2), w2 = n2 / (n1+n2)

n = n1 + n2 - c min(n1,n2)

c is a parameter between 0 and 1 ....  Ideally there would be a version of merge taking c as a parameter, and then a version using a default value for c (which could be c=0 or c=.2 or whatever...)

The STV uses (s,c) where c = n /(n+k), where n is the count...

Getting this merge function right is important because without it we can't use SimpleTruthValue to accumulate frequency information about data.

linas:
truth-value merge is a sticky problem in the atomspace.

1) there is no way to selected a "prefered" truth-value merge formula (as different things may want different formulas)

2) there is no clear way for merging truth values from two different cogservers that are sharing data.  Lets say they are sharing via the sql backend. So cogserver A writes out a TV,  cogserver B pulls that value, merges with its own private copy, writes out the merged result.  cogserver A notices the TV changed, pulls that TV, merges with its own TV, writes out the result.  Now repeat ... 

3) confusion about merging default TV's. was the default TV there "by accident", and thus should not be merged?  Or did we really mean to average it into the mix?

4) if there are multiple different kinds of truth values, how should these be merged?

5) When is a SimpleTV a "Fuzzy set membership TV" and when is it something else?

These kinds of questions made me conclude that "automatic" merging of TV's in the atomspace is almost surely a bad idea, and that the algo that is generating the TV's should take explicit control, and decide what to do in each case, so that its done right for that particular ago.

Regarding frequency: keeping track of frequency is on reason I introduced the "CountTV",  with a third field devoted exclusively to counting.  Since that time, I also have a burning desire for a fourth field: a place to hold the (logarithm of) the frequency (i.e. a normalized count).  I have no thoughts about how these last two could/should interact with the traditional SimplTV strength/confidence values.

bgoertzel:
I think that CountTV and SimpleTV are very similar...  a CountTV can always
be interpreted as a SimpleTV  straighforwardly; and a SimpleTV can always
[...]

PR #1336, opened 2015-03-02
modify STV merge implementation

misgeatgit:
https://github.com/opencog/opencog/issues/1279

### DIFF
--- a/opencog/atomspace/SimpleTruthValue.cc
+++ b/opencog/atomspace/SimpleTruthValue.cc
@@ -38,6 +38,7 @@
 using namespace opencog;
 
 #define KKK 800.0f
+#define CVAL  0.2f
 
 SimpleTruthValue::SimpleTruthValue(strength_t m, count_t c)
 {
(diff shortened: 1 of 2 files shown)


opencog/atomspace/SimpleTruthValue.cc
```diff
@@ -74,16 +75,22 @@ confidence_t SimpleTruthValue::getConfidence() const
...
-    }
-    return clone();
+	switch(ms){
+	case DEFAULT:
+	{
+		if (other->getType() != SIMPLE_TRUTH_VALUE)
+		        throw RuntimeException(TRACE_INFO,
+		           "Don't know how to merge %s into a SimpleTruthValue using the default style",
+		           typeid(*other).name());
+		auto count2 = other->getCount();
```
linas: It would be nice to quote chapter and verse of the text which defines this formula.  I assume it is in some book or pdf somewhere or maybe it is discussed in a wiki page.

PR #1341, opened 2015-03-02
Preload Python modules in PYTHON_PRELOAD_FUNCTIONS directory

inflector:
Added a new configuration variable PYTHON_PRELOAD_FUNCTIONS which holds a directory in which all the python files are loaded as modules on startup. This is to support using Python functions in GroundedSchemaNode and EvaluationLink.
Tested that these are accessible from the CogServer Python shell.


opencog/cython/PythonEval.cc
```diff
@@ -579,50 +585,87 @@ void PythonEval::addSysPath(std::string path)
...
 
+
+const int ABSOLUTE_IMPORTS_ONLY = 0;
+
 /**
 * Add all the .py files in the given directory as modules to __main__ and
 * keep the references in a dictionary (this->modules)
 */
-void PythonEval::add_module_directory(const boost::filesystem::path &p)
+void PythonEval::add_module_directory(const boost::filesystem::path &directory)
```
linas: As mentioned before, it does not make sense (to me) to have functions like this in PythonEval .. surely, some other class or function would be better? I'm thinking modularity -- don't make Eval do things that are not actually needed for evaluation.

inflector: I responded in the email list a bit earlier.

inflector: @linas I'm going to dig into the bug in PythonEval that relates to its interactions with the GenericShell etc. The bug where you can't enter multi-line functions from the shell.

Once I do that, I should know enough to be able to effect a reasonable separation into a Python API calling class and one that defers to it for handling only the evaluation. I'll also check into the way the scheme code does this at the moment as we probably want to keep the Python and Scheme implementations as similar as possible given the current design of shell and evaluator interactions.

PR #1343, opened 2015-03-03
added reference

misgeatgit:
Added  a comment that refers to a section of the PLN book which talks about STV revision rule. https://github.com/opencog/opencog/issues/1279


linas:
thanks.

PR #1344, opened 2015-03-03
Fixed Python error message generation

inflector:
Python error was causing SIGSEGV crash
Fixed Python error message generation
Fixed test_function sample for use in GroundedSchemaNode

### DIFF
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -600,8 +600,7 @@ std::string PythonEval::apply_script(const std::string& script)
     // script had a problem.
     if (errorRunningScript) {
         logger().error() << errorString;
-        throw (RuntimeException(TRACE_INFO, "Python error: %s",
-                errorString.c_str()));
+        throw (RuntimeException(TRACE_INFO, errorString.c_str()));
     }
 
     // printf("Python says that: %s\n", result.c_str());
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -383,7 +383,7 @@ Handle PythonEval::apply(const std::string& func, Handle varargs)
 {
     PyObject *pyError, *pyModule, *pyUserFunc, *pyExecuteUserFunc;
     PyObject *pyArgs, *pyUUID, *pyReturnValue = NULL;
-    PyObject *pyErrorMessage, *pyDict;
+    PyObject *pyDict;
     string moduleName;
     string funcName;
     bool errorCallingFunction;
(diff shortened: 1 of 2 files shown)


inflector:
Tested after pull and merge and there were no problems once I got a good build.

inflector:
Is anything holding this request up? I noticed a few others were merged. I've got another bug fixed and am waiting on this one so I don't submit too large a request. :-)

linas:
I got 80 emails yesterday, and 80 more today.

opencog/cython/PythonEval.cc
```diff
@@ -415,8 +414,7 @@ Handle PythonEval::apply(const std::string& func, Handle varargs)
     // Py_DECREF(pyDict);
 
     // If we can't find that function then throw an exception.
-    if (!pyUserFunc)
-    {
+    if (!pyUserFunc) {
```
linas: coding style regression? after all that talk about style and readability and cognitive load, why would you do this?

inflector: I've been doing this all along you probably didn't notice because I was making other changes. I find the real cognitive load to be having things mix around so your eye has to look for one and the other at the same time.

I actually like the non-attached brace style that was there better, but I've been trying to conform to the standards which specify attaching braces when editing routines that are using different styles in one routine so at least that routine is consistent. So if you want that to be different, let's change the style on the wiki to be the better one. I had assumed you liked the other style for brevity, or something. It doesn't make sense to have coding standards that you don't like.

I'm happy to revert the above changes but I'd like to make the file be consistent in its use of one or the other. The only reason I've been nagging a bit about coding styles is because I'd like to write my code consistently and I just don't know what that style should be. I'm happy for that to be 100% Linas style. :-) As we seem to agree on more than I even thought.

inflector: Though my absolute preference is to do:

```
if (some_test)
    {
    function_goes_here();
    and_more_here();
    }
else
    {
    different_function_goes_here();
    and_more_here();
    }
```
because it's easier to visually match the braces with the code and to see when there is an `else` clause.

So my personal order of preference is the above, then this:

```
if (some_test)
{
    function_goes_here();
    and_more_here();
}
else
{
    different_function_goes_here();
    and_more_here();
 }
```
and then this:
```
if (some_test) {
    function_goes_here();
    and_more_here();

} else {
    different_function_goes_here();
    and_more_here();
 }
```

inflector: What got me started on this particular function was this:

```
    }
    else{
```

PR #1347, opened 2015-03-03
MSTLinkNodes created

rohit12:
This pull request is just a rehash of the previous one that I made. That time, there were many merge conflicts and some minor changes that had to be made. I have made those changes which were required.


linas:
Pleae double-check. Seems to be identical to the previous one, with all of the same objections unresolved: white-space changes to files that don't need change, changes to config files that should not be changed. changes to .gitignore that don't make sense. 

You want a pull request with only two files in it: make-disjuncts.scm and atom-types.script  None of the other changes to the other files should be part of the pull request.

williampma:
This should be closed, as https://github.com/opencog/opencog/pull/1352 is merged.

PR #1350, opened 2015-03-04
remove fig.yml

AmeBel:


linas:
based on some recent confusion, perhap[s its time to remove dockerfiles from the main opencog directory?

AmeBel:
for a new developer just experimenting with opencog i think it is good to
have it in the repo. But i a have no clue what the right way is. @githart
what do you think?

linas:
Hi Amen, sorry, you're right.  I hadn't read the other thread.

PR #1355, opened 2015-03-04
Checked Python C API reference counts, consolidated module import

inflector:


inflector:
I figured out the git voodoo that I'm doing wrong to get the double-commits, so that should not happen in the future. It comes from pushing commits up to my fork, and then finding out I have a conflict that can't be auto merged and then doing another `git rebase`, the second time I just need to do a merge not a rebase since I've already rebased and git's not smart enough to know that.

linas:
Not to put a load on you, but ... does this work with python3?

opencog/cython/PythonEval.cc
```diff
@@ -609,13 +622,83 @@ std::string PythonEval::apply_script(const std::string& script)
...
+    // When this list is empty, this corresponds to an import of the
+    // entire module as is done in the simple import statement:
+    //
+    // import <module>
+    //
+    string fileName, moduleName;
+
+    // Get the module name from the Python file name by removing the ".py"
+    fileName = file.filename().c_str();
+    moduleName = fileName.substr(0, fileName.length()-3);
```
keyvan-m-sadeghi: hmmm, just wondering, what happens if the module is a directory? In that case the file name would be `__init__`

opencog/cython/PythonEval.cc
```diff
@@ -12,6 +12,19 @@
...
+ * DO NOT change the reference count calls below using Py_INCREF
+ * and the corresponding Py_DECREF until you completely understand
+ * the Python API concepts of "borrowed" and "stolen" references.
+ * Make absolutely NO assumptions about the type of reference the API
+ * will return. Instead, verify if it returns a "new" or "borrowed"
+ * reference. When you pass PyObject* objects to the API, don't assume
+ * you still need to decrement the reference count until you verify
+ * that the exact API call you are making does not "steal" the
+ * reference: SEE: https://docs.python.org/2/c-api/intro.html?highlight=steals#reference-count-details
+ * Remember to look to verify the behavior of each and every Py_ API call.
```
keyvan-m-sadeghi: nice one, didn't know about this whole mess back then ;)

opencog/cython/PythonEval.cc
```diff
@@ -648,85 +730,42 @@ void PythonEval::add_module_directory(const boost::filesystem::path &directory)
     //
     // import <module>
     //
-    PyObject* pyModule = NULL;
     PyObject* pyFromList = PyList_New(0);
 
     // Import each of the ".py" files as a Python module.
```
keyvan-m-sadeghi: same caution, `__init__` might prove being devil

opencog/cython/PythonEval.cc
```diff
@@ -749,6 +792,8 @@ void PythonEval::addModuleFromPath(std::string pathString)
         logger().error() << modulePath << " doesn't exists";
     }
 
+    // Release the GIL. No Python API allowed beyond this point.
+    PyGILState_Release(gstate);
```
keyvan-m-sadeghi: there, this used to cause segfault!

opencog/cython/PythonEval.cc
inflector: At the moment, these are private functions, the only place that is calling this is the code that preloads functions for use in GroundedSchemaNodes and EvaluationLinks which is called by the PythonEval initialization code. This is not designed to load normal python modules and there is no reason to do so as Python loads them already by itself. Even the Cython modules are loaded automatically by Python.

If someone puts an `__init__` file in there, this code will give them an `"error trying to load '__ini'"` message. If you think people will be confused by the README in the `preload_functions` and by that message, I can certainly test for it and skip it.

opencog/cython/PythonEval.cc
inflector: Who would have imagined they would have done something so silly?

opencog/cython/PythonEval.cc
inflector: They need to be very carefully balanced.

inflector:
>Not to put a load on you, but ... does this work with python3? - @linas 

It does to the extent that it used to, but I have not tested it. I made no specific alterations which should have busted it, but I have been focusing on getting our Ubuntu Dockerfile setup working reliably first. Ubuntu, Python 2.7.x, etc. If Python 3 is important, I can put some time into testing Python 3.

I was going to start focusing on changing the bindings to be more pythonic first, and then on expanding them to include more of OpenCog to cover as much as the Scheme bindings do. I'd naively put Python 3  testing / support after that in priority.

opencog/cython/PythonEval.cc
keyvan-m-sadeghi: right!

opencog/cython/PythonEval.cc
keyvan-m-sadeghi: Ah I remember now. You're right ;)

keyvan-m-sadeghi:
@inflector I wanted to thank you for the work you've done here, its of great value to OpenCog and is very much appreciated, cheers.

inflector:
@keyvan-m-sadeghi Happy to be of service.

PR #1363, opened 2015-03-06
Added support for EvaluationLink python functions

inflector:


opencog/cython/PythonEval.cc
```diff
@@ -528,10 +544,182 @@ Handle PythonEval::apply(const std::string& func, Handle varargs)
...
+
+    // Release the GIL. No Python API allowed beyond this point.
+    PyGILState_Release(gstate);
+
+    // Throw on error.
+    if (errorCallingFunction)
+        throw RuntimeException(TRACE_INFO, errorString.c_str());
+
+    // Return the truth value pointer from the function call.
+    return tvp;
```
linas: Gahh!!! A giant cut-n-pate of the code immediately above, with assorted tweaks to handle TV instead of handle.  Surely it would have been better to abstract the common parts away, and just have a little kernel that holds only the differences? I think you must be getting tired of this game. ... but I think you are just about out of that long tunnel, anyway...

inflector: We agree on this one. The only difference is that I decided to do this in two steps.

I wanted to do that when I get rid of the bogus Python function executors. You can't do the abstraction properly the way it was setup. That's why I added the todo in the comment. I was planning on doing this next. But I like to make changes in smaller pieces to working systems.

I want to completely eliminate the injected Python functions: `execute_apply_function` and `execute_apply_tv_function`. Those functions specify the return values in hard-coded fashion. That is going to be a bit of work, so I wanted to do that as step two once I got the old method working. Unfortunately since the hard-coded Python functions are responsible for both the parameter unwinding from the ListLink before calling and the return values after calling, there wasn't nearly as much that could be abstracted cleanly as compared to what will be possible when they are redone in C++. Then it will be one larger function with two very short callers to handle the different PyObject* result values.

linas: OK. Maybe I was just projecting... its late, I'm tired.

inflector: No worries, I meant to add something in the pull request comment about the two part process and forgot.

Issue #1293, opened 2015-02-21
Removing InverseHebbianLinks


bgoertzel:
I think it may be possible to remove InverseHebbianLink, thus simplifying the HebbianLink module, by changing the way the math of HebbianLinks is done...

Where AF = the AttentionalFocus, perhaps we may consider

HebbianLink A B <s>

s= s_hebbAB = P( B in AF at time $T | A in AF at time $T)

We can then also look at

InverseHebbianLink A B <s1>

s1= s_invHebbAB = P( B *not* in AF at time $T | A in AF at time $T)

But in this case s_invHebbAB = 1 - s_hebbAB

so I am unsure why we need InverseHebbianLink at all...

We can also look at the expected probability of being in the AF,

s_AF = P(Atom $x is in AF at time $T)

i.e. the percentage of Atoms in the AF at a random time

Only if

s_HebbAB >> s_AF

is s_HebbAB interesting; only if

s_invHebbAB << s_AF

is s_invHebbAB interesting...

Perhaps, rather than having HebbianLinks and InverseHebbianLinks, we
just need to make sure we boost the LTI of HebbianLinks whose
strengths are either surprisingly large or surprisingly small?  (or am
I overlooking something?)

moikle:
Hmmm -- this might be a bit trickier than we thought. Combining s_invHebbAB = 1 - s_HebbAB with s_invHebbAB << s_AF gives s_HebbAB >> 1-s_AF. I don't think we want to boost STI values whenever s_HebbAB >> min(s_AF, 1-s_AF). 

Instead, I propose that we institute an interval of values for being outside the AF. That is, boost STI values whenever s_HebbAB << s_AF_inv or s_HebbAB >> s_AF. Now this clearly only works, though, if s_AF_inv < s_AF. Otherwise we would be boosting STI values for all links! 

I think this mimics, in essence at least, the old design where we had a de facto value of s_AF_inv = 0.

bgoertzel:
Matt -- yes, you are right, this is a good correction....

There should be a comment in the code warning people about the s_AF_inv <
s_AF constraint...

-- Ben


On Fri, Feb 27, 2015 at 12:42 PM, moikle <notifications@github.com> wrote:

> Hmmm -- this might be a bit trickier than we thought. Combining
> s_invHebbAB = 1 - s_HebbAB with s_invHebbAB << s_AF gives s_HebbAB >>
> 1-s_AF. I don't think we want to boost STI values whenever s_HebbAB >>
[...]

PR #1368, opened 2015-03-06
HebbianLinks updating wihout creating InverseHebbianLink  

misgeatgit:
This work is in an effort to address https://github.com/opencog/opencog/issues/1293

### DIFF
--- a/opencog/dynamics/attention/CMakeLists.txt
+++ b/opencog/dynamics/attention/CMakeLists.txt
@@ -35,6 +35,7 @@ ADD_LIBRARY(attention SHARED
 	AttentionModule
 	ForgettingAgent
 	HebbianUpdatingAgent
+	SimpleHebbianUpdatingAgent
 	ImportanceSpreadingAgent
 	ImportanceDiffusionAgent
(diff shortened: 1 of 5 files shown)


linas:
Should @cosmoharrigan  be reviewing and merging instead of me?  Or maybe even @moikle ?

williampma:
Is this breaking the HebbianUpdatingAgentUTest?

misgeatgit:
Thanks @williampma the culprit  is https://github.com/misgeatgit/opencog/commit/55280717e7f59d36fce1fbadbe1300ad2284e72d. I believe what needs to be done is changing the test case as per our previous discussion that truth value updating should mean merging(applying a revision rule) instead of assigning a new truth value. @bgoertzel and @moikle  what is your opinion?

moikle:
That would make sense to me. Also, the HebbianUpdatingAgentUTest code is still referencing InverseHebbian links which no longer exist?

—matt

misgeatgit:
@moikle  the referencing won't cause a problem since the new implementation is in a separate class.

PR #1369, opened 2015-03-07
Refactor user function calls and added test to PythonEvalUTest

inflector:
1. Refactored the way user functions are called so now both `Handle PythonEval::apply` and `TruthValuePtr PythonEval::apply_tv` use the same code.
1. Refactored the Python error message building code so it's now in once place.
1. Added test of cog-execute and cog-evaluate nodes to test `PythonEval::apply` and `PythonEval::apply_tv` to PythonEvalUTest.

(EDIT: This is done ->) There is one more significant refactor left that is not part of this pull request, the change to direct execution of user functions from the current Python function `execute_user_function` calling the user function.

### DIFF
--- a/opencog/cython/opencog/atomspace.pxd
+++ b/opencog/cython/opencog/atomspace.pxd
@@ -82,7 +82,7 @@ cdef extern from "opencog/atomspace/atom_types.h" namespace "opencog":
     cdef Type NOTYPE
 
 # Handle
-ctypedef long UUID
+ctypedef public long UUID
 
 cdef extern from "opencog/atomspace/Handle.h" namespace "opencog":
     cdef cppclass cHandle "opencog::Handle":
--- a/opencog/cython/PythonEval.cc
+++ b/opencog/cython/PythonEval.cc
@@ -81,6 +81,11 @@ AtomSpace* PythonEval::singletonAtomSpace = NULL;
 
 const int NO_SIGNAL_HANDLERS = 0;
 const char* NO_FUNCTION_NAME = NULL;
+const int SIMPLE_STRING_SUCCESS = 0;
+const int SIMPLE_STRING_FAILURE = -1;
+
+// The Python functions can't take const flags.
+#define NO_COMPILER_FLAGS NULL
 
 static bool already_initialized = false;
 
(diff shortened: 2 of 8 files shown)


inflector:
Okay, this refactoring is now complete. This latest set of changes removed the Python code which was wrapping the function calls for user-defined functions called via `PythonEval::apply` and `PythonEval::apply_tv`.

The change was done for clarity, but there are also significant performance advantages to the new approach. The old way for calling Python functions in GroundedSchemaNodes and GroundedPredicateNodes was very high-overhead:

```
Pack Python tuple to call Python execute function
Call the Python execute function which would:
    Check the argument count and throw an exception if it didn't match
    Get out vector for ListLink
    Convert it to a Python list by iterating over the vector creating Python Atoms
    Convert that list to a tuple for the function call
    Call the user function.
```
With these changes, we are just checking the argument count and creating the tuple directly in C++ and then calling the user function directly from C++. This eliminates one tuple creation, one creation of a Python list object with object additions, one iteration over that list, and one Python function call overhead per execution of the Python user function defined in the "py:some_function" clause. I'd be surprised if this didn't speed up calls for smaller functions by a factor of two or more.

There is still a lot of room for optimization in the future, if need be. We could cache the argument count checking and a lot of the Python objects which are being created dynamically for each execution, for instance.

opencog/cython/opencog/atomspace.pxd
```diff
@@ -82,7 +82,7 @@ cdef extern from "opencog/atomspace/atom_types.h" namespace "opencog":
     cdef Type NOTYPE
 
 # Handle
-ctypedef long UUID
+ctypedef public long UUID
```
linas: You can get a minor performance bump by figuring out how to store the entire Handle, instead of the integer UUID only.  By working with UUID's , you incurr the cost of looking up the actual C++ atom that corresponds to the UUID.

inflector: Thanks, I'll look into that.

inflector: Okay, I submitted pull request #1372 which has to perform some very ugly incantations to get the reference to the AtomPtr back to the C++ Handle constructor.

PR #1372, opened 2015-03-08
Pass the AtomPtr through to Python Atom initializers

inflector:


opencog/atomspace/Handle.cc
```diff
@@ -34,6 +34,8 @@ using namespace opencog;
 const Handle Handle::UNDEFINED(ULONG_MAX);
 
 Handle::Handle(AtomPtr atom) : _uuid(atom->_uuid), _ptr(atom) {}
+Handle::Handle(void* atom_ptr) :  _uuid( (*((AtomPtr*)atom_ptr))->_uuid ),
+                                    _ptr(*((AtomPtr*)atom_ptr)) {}
```
linas: No, this is wrong. You can't do that. you'll get broken code.

opencog/cython/opencog/atomspace.pxd
```diff
@@ -88,6 +88,7 @@ cdef extern from "opencog/atomspace/Handle.h" namespace "opencog":
     cdef cppclass cHandle "opencog::Handle":
         cHandle()
         cHandle(UUID)
+        cHandle(void*)
```
linas: No, I was trying to say that the primary interface to cHandle should be cHandle(AtomPtr) instead of cHandle (UUID). You definitely must not have a void* interface, that will result in null pointer derefs and all sorts of other badness.

opencog/cython/opencog/atomspace_details.pyx
```diff
@@ -411,8 +435,10 @@ cdef api object py_atomspace(cAtomSpace *c_atomspace) with gil:
     cdef AtomSpace atomspace = AtomSpace_factory(c_atomspace)
     return atomspace
 
-cdef api object py_atom(UUID uuid, object atomspace):
```
linas: I'm not sure, but I think the above should be ``cdef api object py_atom(Handle h, object atomspace):`` instead of UUID

opencog/atomspace/Handle.cc
inflector: Yes, you can. You may not want to, but you can. It is ugly and hacky as hell but the only way to do what you wanted. The atom_ptr is the address of the shared_ptr, not the address of the Atom. It is dereferenced to be the shared_ptr that is an AtomPtr. 

The names are confusing, I admit. It was originally called atom_ptr_ptr but lots of places call `AtomPtr` variables `atom` in the code, like the line above, 36.

opencog/atomspace/Handle.h
```diff
@@ -85,6 +86,8 @@ friend class AtomspaceHTabler;    // persistance
         return _uuid;
     }
 
+    inline const AtomPtr* atom_ptr_ptr() const { return &_ptr; }
```
inflector: Note that this returns the address of the shared_ptr not the shared_ptr itself.

opencog/cython/PythonEval.cc
```diff
@@ -575,7 +575,7 @@ PyObject* PythonEval::call_user_function(   const std::string& moduleFunction,
         it != argumentHandles.end(); ++it) {
 
         // Place a Python atom object for this handle into the tuple.
-        PyObject* pyAtom = py_atom(it->value(), pyAtomSpace);
+        PyObject* pyAtom = py_atom((void*) it->atom_ptr_ptr(), pyAtomSpace);
```
inflector: And this is getting the address of the shared_ptr at the current iteration in the HandleSeq.

linas:
Hrm. The cure is worse than the disease.  Changing Handle.h and Handle.cc is out of the question.  The goal was to have python be aware of handles, and work with handles, rather working with UUID's

opencog/cython/opencog/atomspace.pxd
inflector: I definitely understand what you were saying. I too thought there must be some way to do that but there isn't because you can't pass AtomPtr (shared_ptr) objects as Python objects into Cython's Python object initializers, which is where these initializers are called from. 

The problem is that the Python Atom object required for the tuple for the call of the Python function has to be initialized using Python PyObject* object arguments. The Python PyObject* objects have very limited types. Cython lets you do a lot but it doesn't change the semantics of Python objects and it doesn't let you pass anything by reference.

The only device for handling more complicated structures that need to be passed through the PyObject* argument interface is a pointer. So you can pass pointers to the shared_ptrs reliably by converting them to PyLong objects first

In fact, the only reliable way to pass pointers across the interface from C++ to the Python object initializers is by passing them as void*s converted to Python variables using the two routines provided for this purpose:

PyLong_AsVoidPtr
PyLong_FromVoidPtr

opencog/cython/opencog/atomspace_details.pyx
inflector: See my comment for line 440 on the right.

Note however, that Handle at this point means Cython/Python Handle and the C++ type has been aliased to cHandle, so you probably mean:
`cdef api object py_atom(cHandle h, object atomspace):`

opencog/cython/opencog/atomspace_details.pyx
```diff
@@ -411,8 +435,10 @@ cdef api object py_atomspace(cAtomSpace *c_atomspace) with gil:
...
-    cdef Handle temphandle = Handle(uuid)
-    cdef Atom atom = Atom(temphandle, atomspace)
+cdef api object py_atom(void* atom_ptr, object atomspace):
```
inflector: The api for this can take any type you want, as far as I can tell. So yes, you could pass a cHandle to this function, by value. But that won't help below.

opencog/cython/opencog/atomspace_details.pyx
```diff
@@ -411,8 +435,10 @@ cdef api object py_atomspace(cAtomSpace *c_atomspace) with gil:
...
+    cdef atom_ptr_long = PyLong_FromVoidPtr(atom_ptr)
+    cdef Handle temp_handle = Handle(None, atom_ptr_long)
```
inflector: This line is the problematic line. You want a Python Handle object to be created, and you must initialize it with Python object arguments only. In this case, I'm initializing it with the None singleton Python object, and a python number object that contains the address of the shared pointer which will, in turn, be passed into the C++ cHandle object's constructor.

Perhaps I should go back to my original variable name, `atom_ptr_ptr`? I thought that was equally confusing because then one wonders what it means when the code uses `atom` to refer to objects of type AtomPtr, then atom_ptr_ptr seems like there is one more layer of indirection than there really is.

opencog/cython/opencog/atomspace_details.pyx
inflector: This was originally:
```
cdef api object py_atom(void* atom_ptr_ptr, object atomspace):
```
which made it clearer in this context that we are not dealing with an `AtomPtr` but an `AtomPtr*` that is being passed in as a void* because Cython doesn't seem to support type casting of pointers and it needs to be a void* just below anyway.

opencog/cython/opencog/atomspace.pxd
inflector: I should say that semantically, if Cython supported reference parameters, this would be:

```
cHandle(AtomPtr&)
```
and not
```
cHandle(void*)
```
because that is what is happening behind the scenes.

opencog/atomspace/Handle.cc
inflector: The fragment `*((AtomPtr*)atom_ptr)` is the `shared_ptr<Atom>` that will get passed by reference into this constructor:

```
shared_ptr( const shared_ptr& r )
```

for the shared_ptr template class. So the reference count will be incremented when this copy gets made as in normal shared_ptr semantics. This is the first and only copy that gets made, here, in the constructor for the C++ Handle (cHandle in Cython). No other copies are made in the process of object initialization and no pointers to the underlying C++ Atom object are passed around anywhere.

linas:
All of your comments are making it more and more plain that doing this is a terribly bad idea. Let's give up now, while we're ahead.

inflector:
Hah, I almost put a comment in the initial pull request stating that I'd not be at all surprised if you don't want to merge this request because it is so ugly. It works, but it sure as hell isn't clear clean code.

PR #1373, opened 2015-03-09
Disjunct extracted from previous MSTNodes

rohit12:


opencog/nlp/types/atom_types.script
```diff
@@ -118,6 +118,7 @@ LG_CONN_DIR_NODE <- NODE    // e.g. "+"
 // Consists of the connector node, the direction, and possibly the
 // multi-connector node.
 LG_CONNECTOR <- LINK   // e.g. "MX+"
+MST_Connector <- "MSTConnector"
```
williampma: This can't be right...  What type of atom is MST_Connector?

rohit12: Ah! I am really sorry! I'll correct it right away! It should be of type link.

PR #1374, opened 2015-03-09
Disjunct extracted from previous MSTNodes

rohit12:
I closed my previous pull request because of a minor error in the atom_types.script file.

### DIFF
--- a/opencog/nlp/learn/make-disjuncts.scm
+++ b/opencog/nlp/learn/make-disjuncts.scm
@@ -93,8 +93,6 @@
 		#f
 		#t))
 
-<<<<<<< HEAD
-=======
 ;Applies the map function to create MST nodes for all the word pairs. It 
 ;creates nodes for only those word pairs which have a mutual information not 
 ;equal to -1000
--- a/opencog/nlp/learn/make-disjuncts.scm
+++ b/opencog/nlp/learn/make-disjuncts.scm
@@ -1,7 +1,8 @@
 ;This is the MST module for creating MSTLinkNodes.
 (use-modules (srfi srfi-1))
 
-;Defines letters of the alphabet to be used in constructing a name for the node.
+;Defines letters of the alphabet to be used in constructing a name for the 
+;node.
 (define letters "ABCDEFGHIJKLMNOPQRSTUVWXYZ")
 
 ;The function which actually constructs a name for the node.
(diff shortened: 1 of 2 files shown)


williampma:
For future record, you don't need to open a new pull request.  If you push to the same git branch, existing pull request on that branch will be update with your newest changes.

opencog/nlp/learn/make-disjuncts.scm
```diff
@@ -84,17 +85,133 @@
...
+;This function takes as input a word and a list of MST nodes created by the 
+;create-MST-nodes function. For the given word, it gives the disjuncts of that 
+;word as found in the list of nodes give to it as input.
+(define (extract-disjunct word nodes)
+	(if (not (null? nodes))
+		(let* (
+				[node (car nodes)]
+				[outset (cog-outgoing-set node)]
+				[linkname (get-MST-name outset)]
+				[firstword (get-first-evaluation-word outset)]
```
linas: This seems to be the only place where `get-first-evaluation-word` is used (and is ever likely to be used)  Thus, could you move its definition to just after line 144? That way, `get-first-evaluation-word` will be scoped only inside of the `extract-disjunct` function, instead of being part of the global namespace.  Similar remarks apply for many of the other functions, especially the short ones.

The other way to hide locally-defined functions and keep them out of the global namespace is to use modules, (i.e. to define this file as a module) ... but, right now, in opencog, we do not yet have uniform module-coding standards, and module support is a bit shaky and possibly buggy.   I hope to clear that up by summertime.

opencog/nlp/types/atom_types.script
```diff
@@ -118,6 +118,7 @@ LG_CONN_DIR_NODE <- NODE    // e.g. "+"
 // Consists of the connector node, the direction, and possibly the
 // multi-connector node.
 LG_CONNECTOR <- LINK   // e.g. "MX+"
+MST_Connector <- LINK "MSTConnector"
```
williampma: Didn't saw it, but lower case naming `MST_Connector` is breaking build.  Need to change to `MST_CONNECTOR`

rohit12: I am really sorry about that. I don't test my build every time I change something since I only need to test Scheme files. I apologize again. I'll correct the mistake immediately.

williampma: No worries.  It's a easy to make mistake. :)

inflector: Rohit,

Good practice for a project like this is to always merge the changes from opencog/master and then run:

```
make test
```

to make sure that you didn't break any of the OpenCog unit tests before submitting a pull request. If you go to:

you can see the status of the builds off of new commits to opencog/master run by buildbot. If one or more of the tests are failing you'll see the red indicator for the build and you can look at the `stdio` output to see what test failed. It will also fail in `make all` if someone breaks the build itself.

So if `make test` fails on your machine, you can check to see if the test that is failing is the one that already fails on the buildbot. If so, then your changes probably didn't cause the failure, especially if the test is for an area of OpenCog where you are not developing. Once you have established that you didn't break the build and you didn't break any unit tests, then you submit the pull request.

Additionally, I always like to check the buildbot after my pull request has been merged to make sure that it also passes the tests and the build works, just in case. In the event that I do break something, I prefer to have the fix ready before anyone else notices. ;-)

PR #1376, opened 2015-03-09
Defined one-use functions locally

rohit12:
I have defined locally all the short functions which could be defined locally. I think I have defined some 10-11 functions locally.
If you have any more changes to suggest please do!


linas:
thank you!

PR #1384, opened 2015-03-10
Experimental number-valued atoms

linas:
This is an demo prototype showing one way in which number-valued NumberNodes could be accomodated in the AtomSpace.  The demo "works" in the sense that you can now say
```
  (cog-evaluate! (NotLink (GreaterThanLink (NumberNode 3)
     (PlusLink (NumberNode 6) (TimesLink (NumberNode 7) (NumberNode 11))))))
```
and get back TRUE_TV.  However, the code is very inefficient, were you to compare it to the equivalent combo code.   There are many reasons for this, and the README lists some of them.  Thus, this really should be considered to be "experimental" code, rather than the 'one true way'.

### DIFF
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -357,8 +357,7 @@ UnorderedHandleSet AtomTable::getHandlesByNames(const char** names,
                                      bool* subclasses,
                                      Arity arity,
                                      Type type,
-                                     bool subclass)
-    const throw (RuntimeException)
+                                     bool subclass) const
 {
     std::vector<UnorderedHandleSet> sets(arity);
 
@@ -471,7 +470,7 @@ bool AtomTable::inEnviron(AtomPtr atom)
     return false;
 }
 
-Handle AtomTable::add(AtomPtr atom, bool async) throw (RuntimeException)
+Handle AtomTable::add(AtomPtr atom, bool async)
 {
     // Is the atom already in this table, or one of its environments?
     if (inEnviron(atom))
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -38,6 +38,7 @@
 #include <opencog/atomspace/Link.h>
 #include <opencog/atomspace/Node.h>
 #include <opencog/atomspace/TLB.h>
+#include <opencog/execution/NumberNode.h>
 #include <opencog/util/exceptions.h>
 #include <opencog/util/functional.h>
 #include <opencog/util/Logger.h>
@@ -510,7 +511,11 @@ Handle AtomTable::add(AtomPtr atom, bool async) throw (RuntimeException)
     if (at != NULL) {
         NodePtr nnn(NodeCast(atom));
         if (nnn) {
-            atom = createNode(*nnn);
+            // Experimental support for NumberNodes
+            if (NUMBER_NODE == nnn->getType())
+                atom = createNumberNode(*nnn);
+            else
+                atom = createNode(*nnn);
         } else {
             LinkPtr lll(LinkCast(atom));
             atom = createLink(*lll);
--- a/opencog/atomspace/AtomTable.cc
+++ b/opencog/atomspace/AtomTable.cc
@@ -644,10 +644,20 @@ Handle AtomTable::add(AtomPtr atom, bool async) throw (RuntimeException)
         }
     }
 
+    // Experimental NumberNode support code
+    if (atom->getType() == NUMBER_NODE) {
+        NumberNodePtr nnp(NumberNodeCast(atom));
+        if (NULL == nnp) {
+           nnp = createNumberNode(*NodeCast(atom));
+           nnp->_uuid = atom->_uuid;
+           atom = nnp;
+        }
+    }
+
     // Its possible that the atom already has a UUID assigned,
     // e.g. if it was fetched from persistent storage; this
     // was done to preserve handle consistency. SavingLoading does
-    // this too.  XXX Review SavingLoading for corrrectness...
+    // this too.  XXX Review SavingLoading for correctness...
     if (atom->_uuid == Handle::UNDEFINED.value()) {
        // Atom doesn't yet have a valid uuid assigned to it. Ask the TLB
        // to issue a valid uuid.  And then memorize it.
--- a/opencog/atomspace/AtomTable.h
+++ b/opencog/atomspace/AtomTable.h
@@ -366,8 +366,7 @@ public:
      * criteria in their outgoing set.
      */
     UnorderedHandleSet getHandlesByNames(const char**, Type*, bool*, Arity,
-                              Type type = ATOM, bool subclass = true) const
-    throw (RuntimeException);
+                              Type type = ATOM, bool subclass = true) const;
 
     /**
      * Returns the set of atoms with the given target names and/or types
@@ -440,7 +439,7 @@ public:
      * @param The new atom to be added.
      * @return The handle of the newly added atom.
      */
-    Handle add(AtomPtr, bool async) throw (RuntimeException);
+    Handle add(AtomPtr, bool async);
 
     /**
      * Read-write synchronization barrier fence.  When called, this
--- a/opencog/atomspace/ClassServer.cc
+++ b/opencog/atomspace/ClassServer.cc
@@ -57,7 +57,7 @@ ClassServer* ClassServer::createInstance(void)
     return new ClassServer();
 }
 
-Type ClassServer::addType(Type parent, const std::string& name)
+Type ClassServer::addType(const Type parent, const std::string& name)
 {
     // Check if a type with this name already exists. If it does, then
     // the second and subsequent calls are to be interpreted as defining
--- a/opencog/atomspace/ClassServer.h
+++ b/opencog/atomspace/ClassServer.h
@@ -75,7 +75,7 @@ public:
     static ClassServer* createInstance(void);
 
     /** Adds a new atom type with the given name and parent type */
-    Type addType(Type parent, const std::string& name);
+    Type addType(const Type parent, const std::string& name);
 
     /** Provides ability to get type-added signals.
      * @warning methods connected to this signal must not call ClassServer::addType or
--- a/opencog/atomspace/Node.h
+++ b/opencog/atomspace/Node.h
@@ -41,7 +41,7 @@ namespace opencog
  */
 class Node : public Atom
 {
-private:
+protected:
     // properties
     std::string name;
     void init(const std::string&) throw (InvalidParamException, AssertionException);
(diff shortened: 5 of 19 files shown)


ngeiswei:
Very interesting. In the README you say that the main cost is with inserting the atom in the atomspace, but depending on the context this cost could be little compared to say evaluating a program over thousands of inputs, wouldn't it?

I remember you also said that an atom can be defined without being inserted in the atomspace.

Otherwise, if the evaluation itself cost much more than with the combo interpreter, I guess the combo code can be kept around, maybe we don't have a choice, we need the 2 sides of the coin.

ngeiswei:
Linas, you haven't merged, I'll let you do it in case it was intentional.

linas:
Yes, you can use atoms without placing them in the atomspace, as long as you keep things simple. Both python and scheme often/usually need atoms to be in an atomspace to work with them; this is partly a historical accident, and partly a snow-ball effect coming from other requirements. Dis-entangling this would take a fairly large effort; there are various spots where some tough decisions would be needed.

bgoertzel:
On Wed, Mar 11, 2015 at 4:39 AM, ngeiswei <notifications@github.com> wrote:

> Very interesting. In the README you say that the main cost is with
> inserting the atom in the atomspace, but depending on the context this cost
> could be little compared to say evaluating a program over thousands of
> inputs, wouldn't it?
>
> I remember you also said that an atom can be defined without being
> inserted in the atomspace.
>
> Otherwise, if the evaluation itself cost much more than with the combo
> interpreter, I guess the combo code can be kept around, maybe we don't have
> a choice, we need the 2 sides of the coin.
>

If the evaluation itself is a lot more expensive in Atomspace than in
Combo, then we could still make the MOSES population a set of Atoms if we
want to --- but then associate with certain sub-hypergraphs a "compiled
version" (Combo) that is optimized for faster execution.    If the
Atomspace version of the sub-hypergraph changes then the compiled version
has to be re-generated....

For instance in the case of

  (cog-evaluate! (NotLink (GreaterThanLink (NumberNode 3)
     (PlusLink (NumberNode 6) (TimesLink (NumberNode 7) (NumberNode 11))))))

we could have the NotLink associated with a Combo tree...

Or would we need a SetLink that links to all the Atoms in the
expression, and then associate the Combo tree with that SetLink?

ben

ngeiswei:
Yes, it's actually an idea that can be used more broadly, have a "compiler", that turns a program as hypergraph into a compiled GSN (in combo, scheme, Python, or maybe even binary code using LLVM).

This makes me realize something, we probably want to keep combo as it is, its interpreter, etc, just remove the current to-be-obsolete  action/perception vocabulary part, and add support for Atoms (including GSN, etc), which will be a completely substitute for action/perception.

Well, this has been in the original design for a while, now is the time.

linas:
Technically, its a lot easier to stick atoms into combo vertexes, than it would be to turn atoms into combo vertexes.

If you follow this route, then I think you want to stick Handles (not Atoms and not AtomPtr's) into combo vertexes.

bgoertzel:
On Thu, Mar 12, 2015 at 6:21 AM, Linas Vepštas <notifications@github.com>
wrote:

> Technically, its a lot easier to stick atoms into combo vertexes, than it
> would be to turn atoms into combo vertexes.
>
> If you follow this route, then I think you want to stick Handles (not
> Atoms and not AtomPtr's) into combo vertexes.
>

So that e.g. the Combo Vertex for "+" would do addition in its usual way,
but would also contain the Handle of the "+" GroundedSchemaNode?

ben

bgoertzel:
On Thu, Mar 12, 2015 at 4:53 AM, ngeiswei <notifications@github.com> wrote:

> Yes, it's actually an idea that can be used more broadly, have a
> "compiler", that turns a program as hypergraph into a compiled GSN (in
> combo, scheme, Python, or maybe even binary code using LLVM).
>
> This makes me realize something, we probably want to keep combo as it is,
> its interpreter, etc, just remove the current to-be-obsolete
> action/perception vocabulary part, and add support for Atoms (including
> GSN, etc), which will be a completely substitute for action/perception.
>

yeah, so that for a Combo tree Vertex to refer to the "blink" action, it
would contain the Handle of the "blink" GSN ... ?


> Well, this has been in the original design for a while, now is the time.
>

Actually next month is probably the time -- now is the time for you to get
some examples of PLN doing something cool with exported MOSES program trees
;)

But yeah, sure, "now" is a variable-width entity !!

linas:
> So that e.g. the Combo Vertex for "+" would do addition in its usual way,
> but would also contain the Handle of the "+" GroundedSchemaNode?
>

If the goal is to have one-to-one, side-by-side layouts, then it would have
to point at PlusLink

If not that, then some sort of implicit understanding that when a
GroundedSchemaNode  occurs in combo, there is an implicty
ExecutionOutputLink just above it.

I have not really looked at or thought about how to define "user-defined
functions" in combo.

I'm also a little hazy about the desired use-case.  What exactly, do we
hope to do?  Nil's "compiler" can "easiy" turn combo trees into opencog
atoms. Writing something that does the reverse translation is not hard,
either .. depending on what you want to do.

If you want to support GroundedSchemaNode "scm: anything" or
GroundedSchemaNode "py: anything" in combo/moses ... that's possible, but
... ?? why? (Just because its cool doesn't mean its a good idea..)

There's a diferent moses issue I don't really understand: at the bottom of
those combo trees are always column names of columns from some table.  What
is the atomspace equivalent of these?

bgoertzel:
>
> I'm also a little hazy about the desired use-case. What exactly, do we
> hope to do? Nil's "compiler" can "easiy" turn combo trees into opencog
> atoms. Writing something that does the reverse translation is not hard,
> either .. depending on what you want to do.



My near-term use case is

-- importing some Combo programs into the Atomspace

-- using PLN to learn some new Boolean program trees, based on these Combo
programs plus other information

-- exporting the learned Boolean program trees back into MOSES to seed a
new deme

All this of course can be done without hooking MOSES and the Atomspace
together in any fancy way, just by importing/exporting...

OTOH a closer linkage would make it nicer...

- Ben

ngeiswei:
On 03/11/2015 11:21 PM, Linas Vepštas wrote:
> Technically, its a lot easier to stick atoms into combo vertexes, than
> it would be to turn atoms into combo vertexes.
>
> If you follow this route, then I think you want to stick Handles (not
> Atoms and not AtomPtr's) into combo vertexes.

Sounds good. Although I have a question:

Do Handle really uniquely identify an Atom? I mean that if MOSES is 
storing its atoms on a local AtomSpace, and then it wants to copy the 
combos (with those atoms) to another AtomSpace, are the Handle gonna 
stay the same (same UUID), or is there some translation needed?

ngeiswei:
On 03/12/2015 01:50 AM, bgoertzel wrote:
> On Thu, Mar 12, 2015 at 4:53 AM, ngeiswei <notifications@github.com> wrote:
>
>  > Yes, it's actually an idea that can be used more broadly, have a
>  > "compiler", that turns a program as hypergraph into a compiled GSN (in
>  > combo, scheme, Python, or maybe even binary code using LLVM).
>  >
>  > This makes me realize something, we probably want to keep combo as it is,
>  > its interpreter, etc, just remove the current to-be-obsolete
>  > action/perception vocabulary part, and add support for Atoms (including
>  > GSN, etc), which will be a completely substitute for action/perception.
>  >
>
> yeah, so that for a Combo tree Vertex to refer to the "blink" action, it
> would contain the Handle of the "blink" GSN ... ?

Exactly. This could be a nice substitute for the WorldWrapper code. 
There are issues however

1. It maybe introduces run-time inefficiencies.

2. While learning and running evaluation in its imaginary world, the 
GSNs are gonna have to trigger different things. Maybe they can just 
point to different code, for instance GSN "py: blink" could be 
reimplemented to run in this imaginary world, and that file would be 
loaded instead of the real world version

ngeiswei:
On 03/12/2015 05:17 AM, Linas Vepštas wrote:
>  > So that e.g. the Combo Vertex for "+" would do addition in its usual way,
>  > but would also contain the Handle of the "+" GroundedSchemaNode?
>  >
>
> If the goal is to have one-to-one, side-by-side layouts, then it would have
> to point at PlusLink
>
> If not that, then some sort of implicit understanding that when a
> GroundedSchemaNode occurs in combo, there is an implicty
> ExecutionOutputLink just above it.
>
> I have not really looked at or thought about how to define "user-defined
> functions" in combo.

This is already fully supported (though MOSES doesn't use it), see

opencog/comboreduct/combo/procedure_call.{h,cc}

>
> I'm also a little hazy about the desired use-case. What exactly, do we
> hope to do? Nil's "compiler" can "easiy" turn combo trees into opencog
> atoms. Writing something that does the reverse translation is not hard,
> either .. depending on what you want to do.

Well, frankly it's rather anticipatory, we've just started experimenting 
with exporting combos in the AtomSpace, it just feels we'd need that, 
but we don't need it today (the first step is to get PLN reason about them).

>
> If you want to support GroundedSchemaNode "scm: anything" or
> GroundedSchemaNode "py: anything" in combo/moses ... that's possible, but
> ... ?? why? (Just because its cool doesn't mean its a good idea..)

As I said it could be a nice substitute for the WorldWrapper. That way 
MOSES can evaluate combos involving GSN (action/perception or whatnot).

>
> There's a diferent moses issue I don't really understand: at the bottom of
> those combo trees are always column names of columns from some table. What
> is the atomspace equivalent of these?

Good question. It's probably possible to export the fitness functions in 
the atomSpace as well, so you'd have a GSN or GPN that takes a combo, a 
list of inputs/outputs (ListLink of ListLink, whatever) and compute the 
fitness. That GSN could actually be an hypergraph using map or fold 
operators, etc. But again it's rather anticipatory, we don't need that 
today.

To be honest I'm kinda revisiting my earlier opinion that we can just 
move MOSES entirely in the AtomSpace, or maybe we can but, but just for 
efficiency reasons we may still want to keep the combo language and 
interpreter and other structures.

linas:
On Thu, Mar 12, 2015 at 3:08 AM, ngeiswei <notifications@github.com> wrote:

>
>
> On 03/11/2015 11:21 PM, Linas Vepštas wrote:
> > Technically, its a lot easier to stick atoms into combo vertexes, than
> > it would be to turn atoms into combo vertexes.
> >
> > If you follow this route, then I think you want to stick Handles (not
> > Atoms and not AtomPtr's) into combo vertexes.
>
> Sounds good. Although I have a question:
>
> Do Handle really uniquely identify an Atom? I mean that if MOSES is
> storing its atoms on a local AtomSpace, and then it wants to copy the
> combos (with those atoms) to another AtomSpace, are the Handle gonna
> stay the same (same UUID), or is there some translation needed?
>

Whenever you place an atom into an atomspace, you are given a handle to it.
That is the handle that you should use (for that atom in that atomspace)

This solves multiple problems:

-- if you create two identical atoms, and put them into the same atomspace,
the returned hadnle for the second add will point at the first atom. (and
the first uuid)

-- If you loose or forget the handle, that's OK, as long as you remember
the name. If you remember the name, just ask the atomspace for it again.

-- if you create one atom and try to stick it into multiple independent
atomspaces, each atomspace will return the handle for its particular copy.
(each gets a different uuid)

-- the above does not hold when the atomspaces are not independent, but are
arranged as main-space & extension if there already is an atom in the main
space, and you try to put it into the extension, it won't go in there; you
just  get back the main handle.  the only things you can put in the
extension are atoms that are not yet in main.

(bad stuff happens if you put it into extension first, and main later. this
could be solved, but has not been yet)

Linas

PR #1387, opened 2015-03-11
DFCCBUTest changed scm loading code

williampma:
@misgeatgit the test was failing on buildbot.  This fixes that.


misgeatgit:
Thanks @williampma

PR #1390, opened 2015-03-11
Minor reasoning

williampma:
@misgeatgit rule-engine.conf looks like no longer needed, so I removed it.  If that's not the case, please close this pull request.  (The README update is just personal preference :) )


misgeatgit:
Grepping shows rule-engine.conf has been referred in the README once.However, changes to README is inevitably after the current iteration is done. So lets merge your changes.

Issue #1380, opened 2015-03-10
cog-fc causes segfault


AmeBel:
Steps for reproducing the segfault, after removing all rules except the deduction rule in `default_cpolicy.json`

```
guile> (clear) (load-scm-from-file "../opencog/python/pln/examples/deduction/atomspace_contents.scm")
0
guile> (define a (list-ref (cog-get-atoms 'InheritanceLink) 0))
guile> (cog-fc a)
(ListLink
)
guile> (cog-fc a)

```

Below is the Stack Trace:
```
[2015-03-10 01:39:52:630] [ERROR] Caught signal 11 (Segmentation fault) on thread 139717857892096
        Stack Trace:
        2: ./opencog/server/cogserver (opencog::Logger::Base::~Base() +0x5e) [0x40813e]
        3: ./opencog/server/cogserver (sighand(int) +0xe9) [0x407779]
        4: /lib/x86_64-linux-gnu/libc.so.6 ( +0x36d40) [0x7f12ae480d40]
        5: /home/opencog/build/opencog/atomspace/libatomspace.so (opencog::NodeIndex::getHandle(unsigned short, char const*) const +0x13) [0x7f12ae208993]
        6: /home/opencog/build/opencog/atomspace/libatomspace.so (opencog::AtomTable::getHandle(unsigned short, std::string const&) const +0x50) [0x7f12ae1d9780]
        7: /home/opencog/build/opencog/atomspace/libatomspace.so (opencog::AtomSpaceImpl::addNode(unsigned short, std::string const&, bool) +0x2e) [0x7f12ae1cfc9e]
        8: /home/opencog/build/opencog/reasoning/RuleEngine/rule-engine-src/libruleengine.so (DefaultForwardChainerCB::target_to_pmimplicant(opencog::Handle, std::map<opencog::Handle, std::string, std::less<op
encog::Handle>, std::allocator<std::pair<opencog::Handle const, std::string> > >) +0x54f) [0x7f129b649b8f]
        9: /home/opencog/build/opencog/reasoning/RuleEngine/rule-engine-src/libruleengine.so (DefaultForwardChainerCB::target_to_pmimplicant(opencog::Handle, std::map<opencog::Handle, std::string, std::less<op
encog::Handle>, std::allocator<std::pair<opencog::Handle const, std::string> > >) +0x2a4) [0x7f129b6498e4]
[...]

PR #1399, opened 2015-03-13
change InheritanceLink to EvaluationLink

misgeatgit:
@cosmoharrigan or @AmeBel  can you review this?


williampma:
Humm... I thought the file is for testing PLN "deduction"...  while the change is the correct representation, does it not affect the example?

AmeBel:
Except the deduction that Peter moves all should be fine.

williampma:
Might be just better to change (PredicateNode "Moves") to (ConceptNode "Movable") to keep both correct representation and all the deduction... but I guess that's not a major problem.

Does this change affect the testing of #1380?

misgeatgit:
@williampma It won't affect that. I just happen to see it while I was inspecting the outputs.

PR #1402, opened 2015-03-13
ss_delete takes atom or list; returns empty list

gavrieltal:
Hi,

My name is Gavriel Loria (a GSoC 2015 candidate), and I am making my first patch to OpenCog. I believe I have added to ss_delete() extra functionality: it returns an empty list on success, and it can take a list of atoms as an argument. I check that each member of the list is an atom prior to deleting any of them.

### DIFF
--- a/opencog/guile/SchemeSmobNew.cc
+++ b/opencog/guile/SchemeSmobNew.cc
@@ -1,5 +1,5 @@
 /*
- * SchemeSmobNew.cc
+ * SchemeSmobNew.cc 
  *
  * Scheme small objects (SMOBS) --creating new atoms -- for opencog.
  *
--- a/opencog/guile/SchemeSmobNew.cc
+++ b/opencog/guile/SchemeSmobNew.cc
@@ -13,6 +13,7 @@
 #include <cstddef>
 #include <libguile.h>
 
+#include <iostream>
 #include <opencog/atomspace/AtomSpace.h>
 #include <opencog/atomspace/ClassServer.h>
 #include <opencog/guile/SchemeSmob.h>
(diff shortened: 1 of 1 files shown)


williampma:
Thanks for the contribution :)  However, I am not sure where you want to use this.

Instead of
```
(cog-delete (list of atoms ...))
```
isn't it better to do
```
(map cog-delete (list of atoms ...))
```
which is the preferred way to handle list in Scheme.

gavrieltal:
I tried that, but I couldn't figure out how to use scm_map syntactically.
Something like

scm_map(SCM (*...::ss_delete)(SCM,SCM), satom)

looked right but would always throw an error (expected ',' before
expression..)

williampma:
Huh?  You don't happen to be trying to interact with the atomspace using guile C code directly right?

Guile is only used for adding bindings to interact with the atomspace using Scheme code (mostly in the cogserver scheme shell).

inflector:
Gavriel,

The ss_ functions are part of the scheme bindings which bind some of the
cog- functions in Scheme to their C++ underlying implementation. So, you
really only want to mess with them if you are working on changing the
behavior of the way that Scheme interacts with the C++ underlying code.

That's why William assumed you were trying to change the way that
cog-delete worked and why he pointed out to you that this was unnecessary,
because if you are in scheme, then map is always available to you.

If you are working with C++ to interact with OpenCog nodes and links and
the atomspace, you never go near SCM representations. If you want to delete
atoms in C++ you just delete them from the atomspace directly using the
public AtomSpace method defined in AtomSpace.h:

https://github.com/opencog/opencog/blob/master/opencog/atomspace/AtomSpace.h#L310

/** * Removes an atom from the atomspace, and any attached storage. * The
atom remains valid as long as there are Handles or AtomPtr's * that
reference it; it is deleted only when the last reference * goes away. * *
@param h The Handle of the atom to be removed. * @param recursive
Recursive-removal flag. If the flag is set, * then this atom, and
*everything* that points to it will * be removed from the atomspace. This
can cause a large * cascade of removals! If the flag is not set, then the *
atom will be removed only if its incoming set is empty. * By default,
recursion is disabled. * @return True if the Atom for the given Handle was
successfully * removed. False, otherwise. */ bool removeAtom(Handle h, bool
recursive = false)

Note that Atom references are implemented in OpenCog as the Handle class
which contains both a UUID long and a std::shared_ptr to the underlying
Atom C++ object. So OpenCog doesn't actually delete the underlying until
and unless the shared_ptr reference count drops to zero. You don't actually
need to delete Atom objects explicitly.

gavrieltal:
I suppose I am still unclear then on what type of change was intended (or
in which source code to make the appropriate change) when the README in the
guile directory recommended building a cog-delete that also takes a list of
atoms.

Thank you for the detailed explanations.

linas:
Sometimes it seems like if could be nice/convenient to have many of these functions take lists as well as individual atoms. However, as William points out, using 'map' requires only four more characters of typing, and is the generally-accepted way of doing such things.  Perhaps c++ code could be justified if it was a performance bottleneck: but this c++ code is not efficient: it copies lists, it measures lengths, it calls wrappers instead of performing the actions directly.  By contrast, map is probably more efficient, especially with newer versions of guile, where the guile compiler would automatically run.

gavrieltal:
Were you still interested in having ss_delete return an empty list instead
of #t?

linas:
The todo-list in the README is 6 years old and consists of some random ideas that were jotted down.  Not all of these ideas were good ideas, either then, or now. I am reviewing it, and will correct it.

linas:
The most interesting unfinished, open project within the guile bindings is to finish the modularization work -- that is, use opencog without having to use the cogserver.  There are examples for how to do this here: https://github.com/opencog/opencog/tree/master/examples/guile

However, there are two problems:

-- Not everything is modular like this; various parts need cleanup, porting, etc.
-- Even a very short moment of use reveals that the `export LTDL_LIBRARY_PATH=` mechanism is painful and clunky,  Something more elegant would be excellent.  No clue what that would be.

PR #1403, opened 2015-03-14
Benchmark updates and change of AtomTable _atom_set to use std::unordered_set

inflector:
Added options to benchmark.py to support:
*  averaging over iterations to reduce drift for smaller tests
*  columnar output
*  help and usage
*  benchmark coverage areas for targeted benchmarking
*  default of faster spread of benchmark timing across areas

Changed `AtomTable`'s `_atom_set` to use `std::unordered_set` which is about twice as fast

### DIFF
--- a/opencog/atomspace/AtomTable.h
+++ b/opencog/atomspace/AtomTable.h
@@ -93,7 +93,7 @@ private:
     // handle uuid and the actual atom pointer (since they are stored
     // together).  To some degree, this info is duplicated in the Node
     // and LinkIndex below; we have this here for convenience.
-    std::set<Handle, handle_less> _atom_set;
+    std::unordered_set<Handle, handle_hash> _atom_set;
 
     //!@{
     //! Index for quick retreival of certain kinds of atoms.
--- a/opencog/benchmark/README
+++ b/opencog/benchmark/README
@@ -93,7 +93,31 @@ and then run:
 
  $ python ./opencog/benchmark/benchmark.py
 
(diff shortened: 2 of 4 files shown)


opencog/atomspace/AtomTable.h
```diff
@@ -93,7 +93,7 @@ class AtomTable
     // handle uuid and the actual atom pointer (since they are stored
     // together).  To some degree, this info is duplicated in the Node
     // and LinkIndex below; we have this here for convenience.
-    std::set<Handle, handle_less> _atom_set;
+    std::unordered_set<Handle, handle_hash> _atom_set;
```
linas: I would prefer not to make this change, unless you have some evidence that this makes things faster ...

inflector: I wrote about this on the list first, you must have missed it in the flurry of GSoC posts:

https://groups.google.com/forum/#!topic/opencog/6_1G6GV_pKA

linas: As a general rule, it would be better to put unrelated changes into different pull requests.  I don't mind the benchmark changes, but will have to explore this one a bit.

inflector: Makes sense. I tried to lay out the issues on the list before submitting this. I wrote the benchmark changes while researching the effects of the change as I wanted to be able to say how much this change affected the Python binding performance. Still, I could have submitted the two parts separately certainly and will do so in the future.

PR #1405, opened 2015-03-14
avoid using scheme where it is not needed.

linas:
That, and a bunch of whitespace changes.

The only non-trivial changes here are those of lines 54, 62 and 96 of 
opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc

I think those changes are correct -- it seems to be what you are trying to do here, with scheme, except that scheme is not need to do any of this.  However, I cannot see what's really going on, because the test cases won't run, because the test_cpolicy.json cannot be found, because there is no checking of the search paths for this file, because the search paths are not being taken from the test-config file.  

You probably should be using the code in util/files.cc to find and load text files.

Or you can do something like this:

```
int load_json_file_relative (AtomSpace& as, const std::string& filename,
                            std::vector<std::string> search_paths)
{
    if (search_paths.empty())
        search_paths = DEFAULT_MODULE_PATHS;

    int rc = 2;
    for (const std::string& search_path : search_paths) {
        boost::filesystem::path modulePath(search_path);
        modulePath /= filename;
        logger().debug("Searching path %s", modulePath.string().c_str());
        if (boost::filesystem::exists(modulePath)) {
            rc = load_json_file(as, modulePath.string());
            if (0 == rc) {
                logger().info("Loaded %s", modulePath.string().c_str());
                break;
            }
        }
    }

    if (rc)
    {
       logger().warn("Failed to load file %s: %d %s",
                     filename.c_str(), rc, strerror(rc));
    }
    return rc;
}
```

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/ControlPolicyParamLoader.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/ControlPolicyParamLoader.cc
@@ -59,6 +59,9 @@ void ControlPolicyParamLoader::load_chaining_rules()
 		istringstream is(rule_names[0]);
 		string var_name;
 		while (getline(is, var_name, ',')) {
+			// I think this is what you want, right????  Don't use
+			// scheme... for something this simple?
+			// Rule *r = new Rule(as_->addNode(VARIABLE_NODE, var_name);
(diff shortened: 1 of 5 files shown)


misgeatgit:
Thanks @linas.

PR #1408, opened 2015-03-15
Changed shared_mutex to mutex 6.5% speed increase

inflector:
Details of the speed improvement and rationale in diary.txt which is part of this pull request.

### DIFF
--- a/opencog/atomspace/ClassServer.cc
+++ b/opencog/atomspace/ClassServer.cc
@@ -66,7 +66,7 @@ Type ClassServer::addType(const Type parent, const std::string& name)
     Type type = getType(name);
     if (type != NOTYPE) {
         /* unique_lock here upgrades the shared mutex to a writer lock */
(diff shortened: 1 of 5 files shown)


inflector:
I've been digging into the code for AtomTable execution to determine the relevant factors for a high-performance Python interface and noticed that getHandleSet (really getHandlesByType) was the slowest part. As I'm trying to learn the internals better so I can work more effectively, I spent some time looking at exactly what was happening in these calls and noticed some ares for improvement base on things I've successfully optimized in the past. No huge wins but some good incremental improvements which result from pretty localized changes internal to AtomSpace.

This is the first of three changes.

linas:
Hi Curtis, the fix is correct, but I got tangled up after looking at your merge requests in backwards order.

I'd like to ask you to push just the diary changes .. except that, with the other changes i've made, all sorts of performance figures are now completely different .. faster, it seems although some should not have been affected, so I'm surprised by that.  Final commit coming "real soon now"

PR #1409, opened 2015-03-15
Added function to get number of atoms of type

inflector:


opencog/atomspace/TypeIndex.cc
```diff
@@ -48,6 +48,31 @@ void TypeIndex::removeAtom(AtomPtr a)
...
+		if ((type == current_type) || 
+		    (count_subclasses && (classserver().isA(type, it.type))))
+		{
+			// Add the size of the atom set for this type.
+			atom_count += idx.at(current_type).size();
+		}
+		current_type++;
+		++it.s;
+	}
+
```
inflector: This loop or something similar, appears in many places, One thing I'd like to do, but wanted to ask first, is build subclass indices in the ClassServer so each type has a vector of subclasses so all this iteration isn't necessary just to find the subclasses.

There aren't that many types so the index won't take up much space, and types are generally going to be fairly static when compared to everything else. Typical use case is probably add some new types on startup and never delete any, I'm guessing.

linas: Yes, that seems like a good idea.  Actually, I think that the classwerver lready has a vector of subclasses in it, and that isA uses that vector. Not sure.  So it would need to be exposed in the API.

Right now, the number of types is small and static.  There were recent discussions on the mailing list (about 2 months ago) which made a rather reasonable argument that, during runtime, there could be many millions of atom types autogenerated.  Basically, every EvaluationLink+PredicateNode would be converted to a new link type, probably inheriting from EvaluationLink.  The classerver would need some work if/when this were to ever happen.

linas:
can you please backport this set of changes to the new atomspace?  The bit difference now is that the index do not use UUID's any more.

inflector:
I'll definitely do that. And I'll implement the Class Server optimization as well and submit a separate pull request. Won't be done tonight or tomorrow but Tuesday or Wednesday I will have time to do it.

opencog/atomspace/TypeIndex.cc
inflector: The existing ClassServer vectors implement a two-dimensional matrix of size number_of_atoms which only contain the IsA relationship as a `std::vector< std::vector<bool> >` which is there for the array-style lookups in IsA. What I'm envisioning is a `vector< vector<Type> >` so the types can be directly iterated over without any boolean checks and skipping of all non-subclasses. Would be faster for iterating and slower for lookups of IsA so we really need both for the two different ways these types are accessed: both sequential and random.

inflector:
This has been ported over with pull request #1495

PR #1410, opened 2015-03-15
Reduce handle copies in AtomTable typeIndex interation

inflector:


opencog/atomspace/AtomTable.h
```diff
@@ -219,10 +219,20 @@ class AtomTable
...
-                            result,
-                            isDefined);
+
+        // We're going to use this one handle so we don't construct/destruct
+        // googles of temporaries in the process of doing the copies.
+        Handle temp_handle;
+        std::for_each(typeIndex.begin(type, subclass),
+                      typeIndex.end(),
+            [&](UUID uuid)->void {
+                if (UNDEFINED_UUID == uuid) return;
```
linas: all atoms in the atom table are always defined, so this check does not make sense.

Yes, I realize it was there in the original code; I am removing it now.

opencog/atomspace/AtomTable.h
```diff
@@ -232,12 +242,26 @@ class AtomTable
...
+                
+                // Since the function could and probably will resolve the
+                // AtomPtr, we need to reset the handle since we are going to
+                // bypass the normal shared_ptr semantics here to avoid
+                // a copy per handle which is much more expensive than
+                // resetting and setting a new UUID. So we call the safe
+                // version which takes a bit longer.
+                temp_handle.set_uuid(uuid);
```
linas: I don't understand the comment. Are you implying that copying a Handle is more expensive than resolving it? Surely, that cannot be right!

opencog/atomspace/Handle.h
```diff
@@ -71,13 +73,33 @@ friend class AtomspaceHTabler;    // persistance
...
+    // if you have any doubts at all about the possibility that the handle
+    // may have been resolved. This sets the new UUID and also calls 
+    // resets the shared_ptr AtomPtr so it doesn't point to anything,
+    // the reference is decremented, and the object deleted if this is the
+    // last one. This is used by AtomTable to avoid having to construct
+    // temporaries while iterating when the iteration itself can resolve the
+    // AtomPtr in order to clear the AtomPtr reference so it won't be pointing
+    // to an atom that has a UUID that is different than the UUID passed
+    // to set_uuid.    
+    void set_uuid(const UUID uuid) { _ptr.reset(); _uuid = uuid; }
```
linas: I don't like this, this is going backwards to the bad old days, where everything was done with UUID's. The UUID is really not meant to be used this way; it is intended only to guaranteed that atoms in different atomspaces (i.e. on different network servers) have a unique identifier.  They aren't supposed to be used for referencing that are within the same address space (i.e within the same process, or shared by different threads.

opencog/atomspace/AtomTable.h
```diff
@@ -271,18 +312,31 @@ class AtomTable
...
+                // Since we resolve the pointer below, we call the safe
+                // version of set_uuid which clears the AtomPtr reference.
+                temp_handle.set_uuid(uuid);
+
+                // If the predicate function returns true, then copy this
+                // handle into the result. The -> resolves the AtomPtr.
+                if (temp_handle->isType(resultAtomType, resultAtomSubclass))
+                    *result = temp_handle;
              });
+        return result;
```
linas: all of these different getHandlesByWhatever functions are semi-deprecated. The correct fix is not to improve them, but figure out who is using them, and why, and try to understand and eliminate those usages.

opencog/benchmark/diary.txt
```diff
@@ -816,3 +816,117 @@ Observations:
...
+==================================================================
+15 March 2015:
+
+by Curtis Faith
+
+Baseline test before change from boost::shared_mutext to boost::mutex 
+in ClassServer::isA(Type sub, Type super) which I tested after doing
+research, including this insightful post:
+
+http://permalink.gmane.org/gmane.comp.lib.boost.devel/211180
```
linas: The anthony williams comments are excellent, I will change these to std::mutex

opencog/atomspace/AtomTable.h
inflector: I was wondering about that. Even added a ??? to one of the files hoping you'd answer it.

opencog/atomspace/AtomTable.h
inflector: No, copying is most certainly much less than resolve.

Except, typeIndex and targetTypeIndex don't hold handles, they hold UUIDs. They return Handles that they create for the express purpose of returning them to this iterator function. It is that creation that I have eliminated, and the subsequent copying on the stack.

In this particular case of iterating over types to get atoms, we're going to need to do a resolve no matter what we do, unless we change typeIndex and targetTypeIndex to use Handles. That is a much bigger change and one I didn't want to even propose without talking to you. I don't know if that would make things faster or slower as I don't know how often these routines are actually used for long-running processes like pattern matching and such. As I learn more, I'll be able to make better proposals.

opencog/atomspace/Handle.h
inflector: It is not going backwards, it is surfacing the backwards that was hidden because someone changed the interface to use Handles at the higher levels while still using UUIDs in the lower internals.

opencog/atomspace/AtomTable.h
inflector: Agreed, I'm still learning the who uses them and why.

opencog/atomspace/Handle.h
inflector: Nevertheless, I'm happy to do it the right way. I didn't think you knew about this stuff since someone else appears to have written the typeIndex code based on the names in the header files. So I pushed the minimal changes necessary to make things faster. That's also why I have tried to carefully split all these changes into minimal pull requests so you can see what makes a difference and more easily refuse parts and accept others.

inflector: Note that `TypeIndex` and `TargetTypeIndex` subclass from `FixedIntegerIndex` and the internal index for `FixedIntegerIndex` is:

```
std::vector<UnorderedUUIDSet> idx; 
```
from: https://github.com/opencog/opencog/blob/master/opencog/atomspace/FixedIntegerIndex.h#L44

So we've got a set of `UUID`s at the core of the `TypeIndex` and `TargetTypeIndex`. This is turned into a handle here in `TypeIndex::iterator::operator*` which would be quite easy to miss during a checkin:

https://github.com/opencog/opencog/blob/master/opencog/atomspace/TypeIndex.cc#L106

and since TargetTypeIndex subclasses TypeIndex:

https://github.com/opencog/opencog/blob/master/opencog/atomspace/TargetTypeIndex.h#L42

you end up with the same operator *. So it looks like these use Handles, but they don't.

opencog/atomspace/AtomTable.h
inflector: See `TypeIndex::iterator::operator*`:

https://github.com/opencog/opencog/blob/master/opencog/atomspace/TypeIndex.cc#L106

Note the passing of a UUID into the Handle(UUID) constructor. This is the object that gets copied onto the stack, and if the optimizer is really good, might only result in one extra copy and constructor / destructor sequence.

So since we don't have a resolved Handle returning from this index, only one that has been newly created using a UUID here in `TypeIndex::iterator::operator* `, my code doesn't add any new resolves.

inflector: I'm going to revise my prior statement: "No, copying is most certainly much less than resolve." In thinking about it, the "most certainly" is just plain dumb of me.

The fact, is, it is possible that it is cheaper to do an unordered_set hash lookup with unsigned in keys than to do several shared_ptr operations and construction / destruction calls along with a memory copy of the members.

I'll test this and see. I've been surprised when my assumptions didn't hold enough to prefer testing in cases that actually matter.

linas: The gcc and the clang optimizers easily optimize away calls to constructors when the constructors are relatively simple. Thus, a typical call to a consructor on stack becomes just an increment of the stack pointer by the object size, and possibly some initialization of parts of it.

linas:
I should have guessed that you are working today .. and let you do all of this ... but instead, I reverted to old habits, and started making (large) changes myself. The Index design is old, and based on some mis-guided/incorrect requirements, so I removed and simplified large parts of that code.  I'll push this out in th next hour or so, once I get a chance to run unit tests.

linas:
I think ll of the changes I'm making should result inperformance speedups, too ...

linas:
To be clear: we are effectively on AtomSpace version 4.0, and there is assorted cruft left over from earlier versions.  The copies were expedients to avoid other large changes...

opencog/benchmark/diary.txt
```diff
@@ -816,3 +816,117 @@ Observations:
...
+
+==================================================================
+15 March 2015:
+
+by Curtis Faith
+
+Noticed that the result set which is being back_inserted did not do a reserve
+for the size ahead of the insertions and most of the time is being spent
+copying and initializing. So I added a getNumAtomsOfType function to TypeIndex
+so that the getHandleSet test could call reserve on the result handle before
```
linas: In the code that I'm re-doing, I added a function size(Type)  that returns the size but I cannot figure out wehre you wanted to call this.

inflector: I just added a pull request that shows the example of this usage. I don't expect you to merge it, but it gives you some idea what I tested that was 13% faster:

https://github.com/opencog/opencog/pull/1414/

linas:
And one more remark: yes, there is almost surely a large over-use of the constructor for Handle ... and I was wrong, the compiler won't usually inline it because its not defined in the header file, it is defined in the source file, and cannot be moved to the header file, due to circular dependency.

I'm also thinking that, for many places where we currently use handles, we could use Atomptrs instead; this would lower mem usage, maybe lower calls to constructors.   The wide-spread use of handles is another legacy design choice of somewhat questionable validity. ...

PR #1412, opened 2015-03-15
Added print of randseed and -R flag to set randseed

inflector:
Was finding about a 5% to 6% drift in the times for successive getHandleSet() benchmarks. The algorithm is very sensitive to the actual random sequence because Atom and other node types high in the hierarchy will result retrievals of all the atoms and leaf nodes will have relatively few, even for the default 100,000 iterations with 50 types, since only 4% or so of the atoms dominate the time, uneven distribution clustering in our out of those values results in higher drift. With this change drift has settled to around 1% when running the same randseed value in succession.

Still even 1% seems high for tests that should no longer be dependent on any values that aren't repeating. For further investigation...


opencog/benchmark/AtomSpaceBenchmark.cc
```diff
@@ -68,14 +70,16 @@ AtomSpaceBenchmark::AtomSpaceBenchmark()
     doStats = false;
     testKind = BENCH_AS;
 
+    randomseed = (unsigned long) time(NULL);
```
linas: please don't do this. This makes results unreproducible.

opencog/benchmark/atomspace_bm.cc
```diff
@@ -25,6 +27,8 @@ int main(int argc, char** argv)
      "          \t(default: 100000)\n"
      "-r <int>  \tLooping count; how many times a python/scheme operation is looped\n"
      "          \t(default: 1)\n"
+     "-R <int>  \tUse specific randomseed; useful for benchmark comparisons\n"
+     "          \t(default: time(NULL))\n"
```
linas: I don't mind the -R flag, but the default is wrong.  Is there any reason to beleive that the results depend on the random number?

opencog/benchmark/AtomSpaceBenchmark.cc
```diff
@@ -68,14 +70,16 @@ AtomSpaceBenchmark::AtomSpaceBenchmark()
...
     asp = NULL;
     atab = NULL;
 #if HAVE_CYTHON
     cogs = NULL;
     pymo = NULL;
 #endif
-
-    rng = new opencog::MT19937RandGen((unsigned long) time(NULL));
+    rng = NULL;
 }
```
inflector: This was already here. I didn't introduce it. I made it possible to do something else.

opencog/benchmark/AtomSpaceBenchmark.cc
inflector: That was the only way it was done before. See lines 77 and 78. I added the ability to use a fixed randseed. It didn't exist before. I copied this setting based on time() from the 

I do find it useful to see different random values at times so you can determine the algorithmic sensitivity to different sequences. Most tests aren't sensitive but the getHandlesOfType was.

opencog/benchmark/atomspace_bm.cc
inflector: Yes, there is. I was seeing a spread of about 4% to 5% at times for the timing of the getHandlesOfType call that dropped to about 1% once I started using any particular fixed randomseed.

It's also useful for stress testing to have different values at times as bugs sometimes show up for rare occurrences. That's also why I'm now printing the value so you can go back and repeat a sequence if you run into a problem.

inflector: The shape of the distribution of numbers to time is really skewed because a very few number to types have many child atoms while others have none. The number of 0s and 1s in the series matters far more than the number of most larger digits.

At least that's my working theory to explain the sensitivity even for 100,000 calls.

inflector: I didn't introduce the default to `time()` as randseed, I preserved what was there. But I can certainly see how most of the time it would be more useful the other way. We could default it to some fixed large number and then have a flag to set it to time() for sensitivity testing in the future. But I'd probably add that with support for multiple test runs and some distribution statistics in columnar display to get some sense for the dependence on random sequences.

At some point, it probably makes more sense to build some type histograms from actual data sets which we can load so we can run some tests that reflect NLP usage and other tests that reflect other OpenCog usages more directly.

linas:
Yes, you are completely right.  That explains a lot --- it certainly explains why this benchmark was always so fruity.  Like everything else, I inherited this block of code, and it is wildly mis-designed and flaky, but I've been too lazy to try to fix it.   Oh well.

Anyway, I cannnot merge because of merge conflicts, can you revise?

inflector:
Yes, I'll do this.

inflector:
Fixed the conflicts.

linas:
thanks.

PR #1414, opened 2015-03-15
Faster get handles by type

inflector:
Here's a change I made which is faster still that uses the size information. I don't expect you to merge this, since it wasn't built off your latest changes, but I wanted you to see the changes I made which made  the 13% better speed improvement due to reduce memory management. It also uses emplace_back.

### DIFF
--- a/opencog/atomspace/AtomTable.h
+++ b/opencog/atomspace/AtomTable.h
@@ -93,7 +93,7 @@ private:
     // handle uuid and the actual atom pointer (since they are stored
     // together).  To some degree, this info is duplicated in the Node
     // and LinkIndex below; we have this here for convenience.
-    std::set<Handle, handle_less> _atom_set;
+    std::unordered_set<Handle, handle_hash> _atom_set;
 
     //!@{
     //! Index for quick retreival of certain kinds of atoms.
--- a/opencog/atomspace/AtomTable.h
+++ b/opencog/atomspace/AtomTable.h
@@ -236,6 +236,60 @@ public:
         return result;
     }
 
+    /**
+     * Appends a set of handles that matches with the given type
+     * (subclasses of type optionally). Same functionality as above but with
+     * a reference to a HandleSeq as the first parameter rather than
+     * an output iterator. This version is less flexible but faster.
+     *
+     * @param the HandleSeq to which the results will be appended.
+     * @param atomsOfType The desired type.
+     * @param includeSubclasses Whether subclasses of atomsOfType should be appended.
+     *
+     * @note Matched entries are appended to HandleSeq passed as first parameter.
+     *          Example of call to this method, which would return all entries in AtomSpace:
+     * @code
+     *         HandleSeq all_atoms;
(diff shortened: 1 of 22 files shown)


opencog/atomspace/AtomTable.h
```diff
@@ -219,10 +220,74 @@ class AtomTable
...
+        // bit pessimistic as there may be undefined handles which will
+        // get filtered. Question: How do undefined handles get into
+        // the typeIndex???
+        size_t size_of_append = typeIndex.getNumAtomsOfType(atomsOfType,
+                                                            includeSubclasses);
+
+        // Now reserve size for the addition. This is much faster for large
+        // append iterations since appends to the list won't require new
+        // allocations and copies whenever the allocated size is exceeded.
+        appendToHandles.reserve(initial_size + size_of_append);
```
inflector: This is the call that make the big difference in speed. It requires line 269 above.

opencog/atomspace/AtomTable.h
```diff
@@ -219,10 +220,74 @@ class AtomTable
...
+            [&](UUID uuid)->void {
+                if (UNDEFINED_UUID == uuid) return;
+
+                // Emplace the Handle(uuid). This constructs the handle in
+                // place where it won't have to be moved eliminating the
+                // temporary and copy required by .push_back(Handle). The
+                // emplace_back call takes the same variants as Handle
+                // constructors since it forwards to the same underlying
+                // C++ mechanisms.
+                appendToHandles.emplace_back(uuid);
```
inflector: This call saves the copy on the stack and tested about 4% faster for me over the call which used the back_inserter.

opencog/benchmark/AtomSpaceBenchmark.cc
```diff
@@ -987,7 +994,7 @@ timepair_t AtomSpaceBenchmark::bm_getHandleSet()
     case BENCH_AS: {
         HandleSeq results;
         clock_t t_begin = clock();
-        asp->getHandlesByType(back_inserter(results), t, true);
+        asp->getHandlesByType(results, t, true);
```
inflector: Here's the new caller, just pass the results in as the first parameter since we're not using the iterators in a loop.

This case is similar to the one that will speed up Python as that's what it is currently doing for many calls, getting a big list of Handles and then copying them. So being able to not have to copy the whole vector from C++ into Python world will be a big win for the yield iterators in Python as we'll be able to iterate over the underlying Handle vector directly in Cython code. The Cython code will be able to pass it's vector into the AtomSpace if we use this approach eliminating a copy. There may be other ways to improve it too, but this way is an easy change compared to the other ways.

inflector:
These changes were ported to the new AtomSpace with pull request #1495

inflector:
The #1495 changes are now 37% faster. Probably because the new AtomPtrs copies are more expensive than Handle copies that before they are resolved to AtomPtrs (the old indices used UUIDs so the Handle copies weren't as expensive).

PR #1416, opened 2015-03-16
Scheme primitive parallel test

williampma:
Just wanted to create something simple to see if the unit test hangup on the buildbot awhile back can be triggered by this.

The way it hanged with no output to the test log seems to indicate deadlock... but maybe not since even the test initialization log are not there.  Maybe it's just the log not being flushed.

### DIFF
--- a/tests/scm/SCMPrimitiveUTest.cxxtest
+++ b/tests/scm/SCMPrimitiveUTest.cxxtest
@@ -67,6 +67,7 @@ class SCMPrimitiveUTest :  public CxxTest::TestSuite
 
 	void test_H_H(void);
 	void test_throw(void);
+	void test_parallel(void);
 };
 
 // Some example class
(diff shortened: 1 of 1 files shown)


tests/scm/SCMPrimitiveUTest.cxxtest
```diff
@@ -200,4 +225,36 @@ void SCMPrimitiveUTest::test_throw(void)
...
+
+	as->clear();
+
+	// Create "mango" for creating, and "tango" for deleting
+	MyExampleClass *mtc = new MyExampleClass(as, 42);
+	define_scheme_primitive("mango", &MyExampleClass::my_more_func_1, mtc);
+	define_scheme_primitive("tango", &MyExampleClass::my_more_func_2, mtc);
+
+	eval->eval("(use-modules (ice-9 threads))");
+	eval->eval_h(R"(
```
linas: What is "R" ?

williampma: C++ string literal.  Just make it easier to type in the scheme code without worrying about the the quote.

PR #1418, opened 2015-03-16
Update the control policy base class

williampma:


AmeBel:
should the rules be in the conf file or a json file ?

williampma:
The ForwardChainer load .json by default.

I am just updating the base class for consistency.

misgeatgit:
I prefer json. conf file is designed for key values.You have to do a lot of string manipulation to support other ways of description in a conf file.

misgeatgit:
The ForwardChainer load .json by default.
I am just updating the base class for consistency.

>Yes, the default is json. we might end up throwing the other one.

AmeBel:
@misgeatgit when you say '.. end up throwing .." does it imply there are cases were it is used?

misgeatgit:
@Amebel  No it is not being used except for the fact that it is a base class for the json loader class.It can be easily refactored out when we really want to do so.But its not a priority now.

AmeBel:
thanks, got it

linas:
After pulling and building this commit, I still cannot run DefaultForwardChainerCBUTest It always fails because it cannot find the json file.  I tried to run it three different ways:
```
cd $PROJECT_SRC_DIR;  ./build/tests/reasoning/RuleEngine/DefaultForwardChainerCBUTest
cd build; ./tests/reasoning/RuleEngine/DefaultForwardChainerCBUTest
cd tests/reasoning/RuleEngine;   ./DefaultForwardChainerCBUTest
```
all three fail with the error:  `DefaultForwardChainerCBUTest.cxxtest:65: Error: Expected (1 == rule.size()), found (1 != 0)`  which means that it cannot read the json file.

misgeatgit:
@linas as I mentioned in another thread its because of the choose_rule function not behaving properly.I am fixing that.

Issue #1401, opened 2015-03-13
PLN deduction formula


moikle:
At some point, the independence-based deduction rule should be corrected to include the consistency conditions. 

IF: 0 <= sAB-max((sA+sB-1)/sA, 0) <= min(1, sB/sA) -max((sA+sB-1)/sA, 0)
AND 0 <=  sBC-max((sB+sC-1)/sB, 0) <= min(1, sC/sB)-max((sB+sC-1)/sB, 0)
   sAC = [sAB.sBC + ((1-sAB)(sC - sBsBC))/(1-sB)] 
ELSE:
   sAC = 0
END

The trickiest part about the corrected formula is that the truth value sAC now depends upon five independent variables sA, sB, sC, sAB, and sBC, rather than only four. This change will need to be propagated through all rules that depend upon deduction so as to include the appropriate conditional requirements.

AmeBel:
What do you mean by " This change will need to be propagated through all
rules"?

moikle:
I will provide the corresponding conditionals for the other rules :-) But generally what I mean is that deduction is used to calculate induction, abduction, transitive similarity, evaluationImplication, etc. 

Now there two forms for the conditions and these can be put into "helper" functions -- then the only thing that changes are the inputs into those functions.

moikle:
So, for example, we can form the functions

consistency1(A, B, AB) = A*B-max((A+B-1)/A, 0) and
consistency2(A, B) = min(1, B/A).

For deduction we then have 

IF consistency1(A, B, AB) >=0 and consistency1(A, B, AB) <= consistency2(A, B)
    and consistency1(B, C, BC) >=0 and consistency1(B, C, BC) <= consistency2(B, C)
THEN AC = [AB.BC + ((1-AB)(C - B*BC))/(1-B)] 
ELSE AC=0

I just worked out the case for induction. The conditions need to be
IF consistency1(B, A, BA) >=0 and consistency1(B, A, BA) <= consistency2(B, A)
    and consistency1(B, C, BC) >=0 and consistency1(B, C, BC) <= consistency2(B, C)
If this is the case then we can simply apply
THEN AB = inversionFormula(BA, B, A)
    AC = deduction(A, B, C, AB, BC) 
ELSE AC=0

[...]

PR #1422, opened 2015-03-16
Added PLN rules required for Socrates Example

prateeksaxena2809:
Added 5 PLN rules:
1. Abduction Rule
2. Member To Evaluation Rule
3. Evaluation To Member Rule
4. Member To Inheritance Rule
5. Inheritance To Member Rule

With these five rules plus the deduction rule, the socrates assertions written in `socrates-r2l.scm` in the python PLN examples directory runs as expected.

### DIFF
--- a/opencog/reasoning/RuleEngine/rules/pln/abduction.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/abduction.scm
@@ -1,0 +1,99 @@
+;==============================================================================
+; Abduction Rule
+;
+; AND(Inheritance A C, Inheritance B C) entails Inheritance A B
+;------------------------------------------------------------------------------
+
+(define pln-rule-abduction
+	(BindLink
+		(ListLink
+			(VariableNode "$A")
+			(VariableNode "$B")
+			(VariableNode "$C"))
+		(ImplicationLink
+			(AndLink
+				(InheritanceLink
+					(VariableNode "$A")
+					(VariableNode "$C"))
+				(InheritanceLink
+					(VariableNode "$B")
+					(VariableNode "$C"))
+				(VariableNode "$A")
+				(VariableNode "$B")
+				(VariableNode "$C"))
+			(ExecutionOutputLink
+				(GroundedSchemaNode "scm: pln-formula-abduction")
+				(ListLink
+					(InheritanceLink
+						(VariableNode "$A")
+						(VariableNode "$B"))
+					(InheritanceLink
+						(VariableNode "$A")
+						(VariableNode "$C"))
+					(InheritanceLink
+						(VariableNode "$B")
+						(VariableNode "$C"))
+					(VariableNode "$A")
+					(VariableNode "$B")
+					(VariableNode "$C"))))))
+; -----------------------------------------------------------------------------
+; Abduction Formula
+; -----------------------------------------------------------------------------
+
+; -----------------------------------------------------------------------------
+; Side-effect: TruthValue of AB may be updated
+; -----------------------------------------------------------------------------
+
+(define (pln-formula-abduction AB AC BC A B C)
+	(cog-set-tv!
+		AB
+		(pln-formula-abduction-side-effect-free
+			AB
+			AC
+			BC
+			A
+			B
+			C)))
+; -----------------------------------------------------------------------------
+; This version has no side effects and simply returns a TruthValue
+; -----------------------------------------------------------------------------
+
+(define (pln-formula-abduction-side-effect-free AB AC BC A B C)
+	(let
+		(
+			(sCB 
+				(/ 
+					(* 
+						(cog-stv-strength BC) 
+						(cog-stv-strength C)
+					) 
+					(cog-stv-strength B)
+				)
+			)
+			(sAC (cog-stv-strength AC))
+			(sC (cog-stv-strength C))
+		)
+			(stv
+				(if (eq? 0 (- 1 sAC))
+					0 ; Set Strength to 0 if sNotB equals 0 to avoid
+					  ; division by zero
+					(+                         ; Strength
+						(* sAC sCB)
+						(/
+							(*
+								(-
+									1
+                                    sAC)
+                                (-
+                                    sCB
+                                    (*
+                                        sAC
+                                        sCB)))
+                            (-
+                                1
+                                sAC))))
+                (min
+                    (cog-stv-confidence AC)
+                    (cog-stv-confidence BC))))) ; Confidence
+
+; =============================================================================
--- a/opencog/reasoning/RuleEngine/rules/pln/abduction.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/abduction.scm
@@ -3,7 +3,7 @@
 ;
 ; AND(Inheritance A C, Inheritance B C) entails Inheritance A B
 ;------------------------------------------------------------------------------
-
+(include "formulas.scm")
 (define pln-rule-abduction
 	(BindLink
 		(ListLink
@@ -17,10 +17,7 @@
 					(VariableNode "$C"))
 				(InheritanceLink
 					(VariableNode "$B")
-					(VariableNode "$C"))
-				(VariableNode "$A")
-				(VariableNode "$B")
-				(VariableNode "$C"))
+					(VariableNode "$C")))
 			(ExecutionOutputLink
 				(GroundedSchemaNode "scm: pln-formula-abduction")
 				(ListLink
@@ -61,39 +58,18 @@
 (define (pln-formula-abduction-side-effect-free AB AC BC A B C)
 	(let
 		(
-			(sCB 
-				(/ 
-					(* 
-						(cog-stv-strength BC) 
-						(cog-stv-strength C)
-					) 
-					(cog-stv-strength B)
-				)
-			)
+			(sCB (inversion-formula 
+				(cog-stv-strength BC)
+				(cog-stv-strength B) 
+				(cog-stv-strength C)))
 			(sAC (cog-stv-strength AC))
 			(sC (cog-stv-strength C))
-		)
-			(stv
-				(if (eq? 0 (- 1 sAC))
-					0 ; Set Strength to 0 if sNotB equals 0 to avoid
-					  ; division by zero
-					(+                         ; Strength
-						(* sAC sCB)
-						(/
-							(*
-								(-
-									1
-                                    sAC)
-                                (-
-                                    sCB
-                                    (*
-                                        sAC
-                                        sCB)))
-                            (-
-                                1
-                                sAC))))
-                (min
-                    (cog-stv-confidence AC)
-                    (cog-stv-confidence BC))))) ; Confidence
+			(sB (cog-stv-strength B))
+			(sA (cog-stv-strength A)))
+		(stv 
+			(simple-deduction-formula sA sC sB sAC sCB) ;Strength
+                	(min
+                    		(cog-stv-confidence AB)
+                    		(cog-stv-confidence BC))))) ; Confidence
 
 ; =============================================================================
--- a/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
@@ -1,5 +1,7 @@
 ; ============================================================================= 
-; GeneralEvaluationToMemberRule
+; GeneralEvaluationToMemberRule 
+;	Takes EvaluationLinks with >=2 Arguments and creates a Member Link
+;
 ;	EvaluationLink (pred D (ListLink B C))
 ;		becomes
 ;	MemberLink( B (SatisfyingSetLink( Variable $X 
@@ -9,34 +11,22 @@
 (define pln-rule-evaluation-to-member
 	(BindLink
 		(ListLink
-			(VariableNode "$B")
-			(VariableNode "$C")
+			(VariableNode "$A")
 			(TypedVariableLink
 				(VariableNode "$D")
 				(VariableTypeNode "PredicateNode")))
-		(ImplicationLink
-			(EvaluationLink
-				(VariableNode "$D")
-				(ListLink
-					(VariableNode "$B")
-					(VariableNode "$C")))
-			(ExecutionOutputLink
-				(GroundedSchemaNode "scm:pln-formula-evaluation-to-member")
-				(ListLink
-					(MemberLink
-						(VariableNode "$B")
-						(SatisfyingSetLink
-							(VariableNode "$X")
-							(EvaluationLink
-								(VariableNode "$D")
-								(ListLink
-									(VariableNode "$X")
-									(VariableNode "$C")))))
-					(EvaluationLink
-						(VariableNode "$D")
-						(ListLink
-							(VariableNode "$B")
-							(VariableNode "$C"))))))))
+        (ImplicationLink
+            (EvaluationLink
+                (VariableNode "$D")
+                (VariableNode "$A"))
+            (ExecutionOutputLink
+                (GroundedSchemaNode "scm:pln-formula-evaluation-to-member")
+                (ListLink
+                    (EvaluationLink
+                        (VariableNode "$D")
+                        (VariableNode "$A")))))))
+
+
 ; -----------------------------------------------------------------------------
 ; Evaluation To Member Formula
 ; -----------------------------------------------------------------------------
@@ -45,19 +35,16 @@
 ; Side-effect: TruthValue of the new link stays the same
 ; -----------------------------------------------------------------------------
 
-(define (pln-formula-evaluation-to-member BXDXC DBC)
-	(cog-set-tv!
-		BXDXC
-		(pln-formula-evaluation-to-member-side-effect-free
-			BXDXC
-			DBC)))
-; -----------------------------------------------------------------------------
-; This version has no side effects and simply returns a TruthValue
-; -----------------------------------------------------------------------------
-
-(define (pln-formula-evaluation-to-member-side-effect-free BXDXC DBC)
-	(stv
-		(cog-stv-strength DBC)
-		(cog-stv-confidence DBC)))
+(define (pln-formula-evaluation-to-member DBC)
+	(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+		(car (cdr (cog-get-all-nodes DBC)))
+		(SatisfyingSetLink
+			(VariableNode "$X")
+			(EvaluationLink
+				(car (cog-get-all-nodes DBC))
+				(ListLink
+					(cons
+						(VariableNode "$X")
+						(cdr (cdr (cog-get-all-nodes DBC)))))))))
 
 ; =============================================================================
--- a/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
@@ -6,6 +6,9 @@
 ;		becomes
 ;	MemberLink( B (SatisfyingSetLink( Variable $X 
 ;		(EvaluationLink (pred D (ListLink X C))))))
+;						&
+;	MemberLink( B (LambdaLink( Variable $X 
+;		(EvaluationLink (pred D (ListLink X C))))))
 ; -----------------------------------------------------------------------------
 
 (define pln-rule-evaluation-to-member
@@ -32,27 +35,59 @@
 ; -----------------------------------------------------------------------------
 
 ; -----------------------------------------------------------------------------
-; Side-effect: TruthValue of the new link stays the same
+; Side-effect: TruthValue of the new links stays the same
 ; -----------------------------------------------------------------------------
 
 (define (pln-formula-evaluation-to-member DBC)
 	(if (= (cog-arity (gdr DBC)) 0)
-		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
-			(gdr DBC)
-			(SatisfyingSetLink
-				(VariableNode "$X")
-				(EvaluationLink
-					(gar DBC)
-					(VariableNode "$X"))))
-		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
-			(gadr DBC)
-			(SatisfyingSetLink
-				(VariableNode "$X")
-				(EvaluationLink
-					(gar DBC)
-					(ListLink
-						(cons
-							(VariableNode "$X")
-							(cdr (cog-get-all-nodes (gdr DBC))))))))))
+		(begin
+			(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+				(gdr DBC)
+				(SatisfyingSetLink
+					(VariableNode "$X")
+					(EvaluationLink
+						(gar DBC)
+						(VariableNode "$X"))))
+			(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+    			(gdr DBC)
+    			(LambdaLink
+        			(VariableNode "$X")
+        			(EvaluationLink
+            			(gar DBC)
+						(VariableNode "$X")))))
+		(pln-create-member DBC '() (cog-outgoing-set (gdr DBC)))))
+
+; -----------------------------------------------------------------------------
+; Creating Multiple Links for Multiple Values
+; -----------------------------------------------------------------------------
+
+(define (pln-create-member DBC preceding-nodes trailing-nodes)
+	(if (not (null? trailing-nodes))
+		(begin
+			(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+				(car trailing-nodes)
+				(SatisfyingSetLink
+					(VariableNode "$X")
+					(EvaluationLink
+						(gar DBC)
+						(ListLink
+							(append
+								preceding-nodes
+								(cons (VariableNode "$X")
+									(cdr trailing-nodes)))))))
+			(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+				(car trailing-nodes)
+				(LambdaLink
+					(VariableNode "$X")
+					(EvaluationLink
+						(gar DBC)
+						(ListLink
+							(append
+								preceding-nodes
+								(cons (VariableNode "$X")
+									(cdr trailing-nodes)))))))
+			(pln-create-member DBC 
+				(reverse (cons (car trailing-nodes) (reverse preceding-nodes)))
+				(cdr trailing-nodes)))))
 
 ; =============================================================================
--- a/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
@@ -36,15 +36,23 @@
 ; -----------------------------------------------------------------------------
 
 (define (pln-formula-evaluation-to-member DBC)
-	(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
-		(car (cdr (cog-get-all-nodes DBC)))
-		(SatisfyingSetLink
-			(VariableNode "$X")
-			(EvaluationLink
-				(car (cog-get-all-nodes DBC))
-				(ListLink
-					(cons
-						(VariableNode "$X")
-						(cdr (cdr (cog-get-all-nodes DBC)))))))))
+	(if (= (cog-arity (gdr DBC)) 0)
+		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+			(gdr DBC)
+			(SatisfyingSetLink
+				(VariableNode "$X")
+				(EvaluationLink
+					(gar DBC)
+					(VariableNode "$X"))))
+		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+			(gadr DBC)
+			(SatisfyingSetLink
+				(VariableNode "$X")
+				(EvaluationLink
+					(gar DBC)
+					(ListLink
+						(cons
+							(VariableNode "$X")
+							(cdr (cog-get-all-nodes (gdr DBC))))))))))
 
 ; =============================================================================
--- a/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
+++ b/opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
@@ -1,0 +1,64 @@
+; ============================================================================= 
+; GeneralEvaluationToMemberRule
+;	EvaluationLink (pred D (ListLink B C))
+;		becomes
+;	MemberLink( B (SatisfyingSetLink( Variable $X 
+;		(EvaluationLink (pred D (ListLink X C))))))
+; -----------------------------------------------------------------------------
+
+(define pln-rule-evaluation-to-member
+	(BindLink
+		(ListLink
+			(VariableNode "$B")
+			(VariableNode "$C")
+			(VariableNode "$D")
+			(TypedVariableLink
+				(VariableNode "$D")
+				(VariableTypeNode "PredicateNode")))
+		(ImplicationLink
+			(EvaluationLink
+				(VariableNode "$D")
+				(ListLink
+					(VariableNode "$B")
+					(VariableNode "$C")))
+			(ExecutionOutputLink
+				(GroundedSchemaNode "scm:pln-formula-evaluation-to-member")
+				(ListLink
+					(MemberLink
+						(VariableNode "$B")
+						(SatisfyingSetLink
+							(VariableNode "$X")
+							(EvaluationLink
+								(VariableNode "$D")
+								(ListLink
+									(VariableNode "$X")
+									(VariableNode "$C")))))
+					(EvaluationLink
+						(VariableNode "$D")
+						(ListLink
+							(VariableNode "$B")
+							(VariableNode "$C"))))))))
+; -----------------------------------------------------------------------------
+; Evaluation To Member Formula
+; -----------------------------------------------------------------------------
+
+; -----------------------------------------------------------------------------
+; Side-effect: TruthValue of the new link stays the same
+; -----------------------------------------------------------------------------
+
+(define (pln-formula-evaluation-to-member BXDXC DBC)
+	(cog-set-tv!
+		BXDXC
+		(pln-formula-evaluation-to-member-side-effect-free
+			BXDXC
+			DBC)))
+; -----------------------------------------------------------------------------
+; This version has no side effects and simply returns a TruthValue
+; -----------------------------------------------------------------------------
+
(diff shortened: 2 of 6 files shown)


prateeksaxena2809:
@AmeBel : Could you kindly have a look at this pull request and provide feedback?

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
```diff
@@ -0,0 +1,99 @@
...
+			(AndLink
+				(InheritanceLink
+					(VariableNode "$A")
+					(VariableNode "$C"))
+				(InheritanceLink
+					(VariableNode "$B")
+					(VariableNode "$C"))
+				(VariableNode "$A")
+				(VariableNode "$B")
+				(VariableNode "$C"))
```
williampma: There's no need to AndLink these nodes right?  They are already in the InheritanceLink.

misgeatgit: Yes, I think the VariableNodes immediate to the AndLink should be removed.

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
```diff
@@ -0,0 +1,99 @@
...
+		(
+			(sCB 
+				(/ 
+					(* 
+						(cog-stv-strength BC) 
+						(cog-stv-strength C)
+					) 
+					(cog-stv-strength B)
+				)
+			)
```
williampma: So this is the inversion formula.

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
```diff
@@ -0,0 +1,99 @@
...
+                                    (*
+                                        sAC
+                                        sCB)))
+                            (-
+                                1
+                                sAC))))
+                (min
+                    (cog-stv-confidence AC)
+                    (cog-stv-confidence BC))))) ; Confidence
+
```
williampma: And I think this is the same as pln-formula-simple-deduction-side-effect-free, hard to tell...

I wonder if there is any way to reuse the formula from other files...

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
```diff
@@ -0,0 +1,64 @@
...
+
+(define pln-rule-evaluation-to-member
+	(BindLink
+		(ListLink
+			(VariableNode "$B")
+			(VariableNode "$C")
+			(VariableNode "$D")
+			(TypedVariableLink
+				(VariableNode "$D")
+				(VariableTypeNode "PredicateNode")))
```
williampma: You defined VariableNode $D twice.

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
prateeksaxena2809: They seemed necessary for the inversion formula and I wanted to reduce duplicating nodes.

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
prateeksaxena2809: Yes

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
prateeksaxena2809: yes it is. Inversion one implication(Here BC) makes it a simple deduction formula. So:
1. Calcucating Strength for (A->C,B->C)-> A->B becomes
2. Calculating Strength for (A->C,C->B) ->A->B

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
AmeBel: Yes, there is no need.

williampma: You can still refer the the VariableNode without AndLink them in the implicant.

prateeksaxena2809: Ok. I will change that.

opencog/reasoning/RuleEngine/rules/pln/inheritance-to-member.scm
```diff
@@ -0,0 +1,75 @@
+; =============================================================================
+; InheritanceToMemberRule
+;	InheritanceLink( B (SatisfyingSetLink (Variable $X
+; 		(EvaluationLink (pred D (ListLink X C))))))
+;			becomes
+;	MemberLink( B (SatisfyingSetLink ( Variable $X 
+;		(EvaluationLink (pred D (ListLink X C))))))
+; -----------------------------------------------------------------------------
```
williampma: Is this really what the Inheritance-to-Member rule does?  The python version seems to be more basic:

```
InheritanceLink(Jade robot) => MemberLink(Jade robot).
```

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
```diff
@@ -0,0 +1,99 @@
...
+			AC
+			BC
+			A
+			B
+			C)))
+; -----------------------------------------------------------------------------
+; This version has no side effects and simply returns a TruthValue
+; -----------------------------------------------------------------------------
+
+(define (pln-formula-abduction-side-effect-free AB AC BC A B C)
```
AmeBel: i think it is best to use the deduction formula as defined in https://github.com/opencog/opencog/blob/master/opencog/python/pln/rules/formulas.py#L88 . that way changes made to the deductionformula will be mirrored here

AmeBel: refer to https://github.com/opencog/opencog/issues/1401 on the updates to the deductionformula

opencog/reasoning/RuleEngine/rules/pln/abduction.scm
AmeBel: it would be great if all the helper functions are defined in another scheme file and then called as needed.  And it best to  have them in few lines otherwise it is really hard to read.

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
```diff
@@ -0,0 +1,64 @@
...
+		(ImplicationLink
+			(EvaluationLink
+				(VariableNode "$D")
```
AmeBel: The predicate is taking only two arguments, you should define a function similar to https://github.com/opencog/opencog/blob/master/opencog/python/pln/rules/direct_evaluation_rules.py#L305 , because the predicate could have 1, 2, or 3 arguments in most cases, but there is nothing preventing it from having more than that

opencog/reasoning/RuleEngine/rules/pln/member-to-evaluation.scm
```diff
@@ -0,0 +1,70 @@
...
+; -----------------------------------------------------------------------------
+
+(define pln-rule-member-to-evaluation
+	(BindLink
+		(ListLink
+			(VariableNode "$B")
+			(VariableNode "$C")
+   			(VariableNode "$D")
+   			(TypedVariableLink
+    			(VariableNode "$D")
```
AmeBel: redundancy here too

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
prateeksaxena2809: The current structure of this code does not account for undefined number of arguments. So, to workaround that problem, I am planning to use the following approach:
For evaluation-to-member:-
1. Extract all Evaluation Links in atomspace using (cog-get-atoms 'EvaluationLink).
2. For each Link in the list created above, Create a Member Link manually using 'car' and 'cdr'.
3. Set TVs of the New Member Link same as that of the Evaluation Link from which it was created.

I started implementing this but haven't yet completed it. I'll wrap it up tomorrow. Does this approach seem good or is there a cleaner/better way to do this?

prateeksaxena2809:
Thanks for your reviews. I'll address all the comments and submit an updated PR tomorrow.

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
AmeBel: i don't get why you would 1. but can't judge untill i see the implementation. Also please note the discussion @ https://github.com/opencog/opencog/issues/1424

AmeBel:
you can push your new commits to the same branch, no need to for opening a new PR. That way it is easier which issues have been handled and what are left. Thanks

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
williampma: It sounds like this solution won't be using the RuleEngine at all?  The new RuleEngine PLN design uses BindLink.

I imagined another hacky approach:

  1. Wrap the undefined number of arguments as one untyped VariableNode like

     ```
  EvaluationLink
    (VariableNode "$D")  ; PredicateNode type
    (VariableNode "$P")  ; no type restriction
     ```

  2. Passes (VariableNode "$P") to the helper function `scm:pln-formula-evaluation-to-member` and construct the new MemberLink in the helper instead.

In my mind it seems that would work, but I haven't tried it.  Anyone got a better idea?

AmeBel: @williampma for 1. What does it got to do with the typing? The issue i raised was about the number of arguments for the predicate, am i missing something?

williampma: Wouldn't the `(VariableNode "$P")` be able to takes care of varying number of arguments?  When there are more than 1 argument to a Predicate, they are always in a ListLink, so `(VariableNode "$P")` will end up as matching to the ListLink, right?

bgoertzel: I hope so...

On Wed, Mar 18, 2015 at 11:13 AM, William Ma <notifications@github.com>
wrote:

> In opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
> <https://github.com/opencog/opencog/pull/1422#discussion_r26635496>:
>
> > +;	MemberLink( B (SatisfyingSetLink( Variable $X
> > +;		(EvaluationLink (pred D (ListLink X C))))))
> > +; -----------------------------------------------------------------------------
> > +
> > +(define pln-rule-evaluation-to-member
> > +	(BindLink
> > +		(ListLink
> > +			(VariableNode "$B")
> > +			(VariableNode "$C")
> > +			(VariableNode "$D")
> > +			(TypedVariableLink
> > +				(VariableNode "$D")
> > +				(VariableTypeNode "PredicateNode")))
> > +		(ImplicationLink
> > +			(EvaluationLink
> > +				(VariableNode "$D")
>
> Wouldn't the (VariableNode "$P") be able to takes care of varying number
> of arguments? When there are more than 1 argument to a Predicate, they are
> always in a ListLink, so (VariableNode "$P") will end up as matching to
> the ListLink, right?
>
> —
> <https://github.com/opencog/opencog/pull/1422/files#r26635496>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

AmeBel: @williampma [just checked it](https://gist.github.com/AmeBel/8b9dc2e4257f2dfc1b8f) yes it works

williampma: That's good to know :)

prateeksaxena2809: I started implementing the rule as discussed in the thread but I am running into an issue. [The code](https://gist.github.com/prateeksaxena2809/bae56405a4aed3898088) on compiling gives the following error.
```
ERROR: In procedure cog-outgoing-set: Wrong type argument in position 1 (expecting opencog atom): ()
ABORT: wrong-type-arg
```
I think this is because on compiling, cog-outgoing-set is not able to detect that the (VariableNode "$A") is of type ListLink. If I change [this line](https://gist.github.com/prateeksaxena2809/bae56405a4aed3898088#file-evaluation-to-member-scm-L24)  to (Variable Node "$A") and [these lines](https://gist.github.com/prateeksaxena2809/bae56405a4aed3898088#file-evaluation-to-member-scm-L30-L33)  to (VariableNode "$X") (VariableNode "$A"), the code compiles successfully (but gives conceptually wrong results as expected).

Any ideas on how I could fix the issue?

AmeBel: you just want [these line](https://gist.github.com/prateeksaxena2809/bae56405a4aed3898088#file-evaluation-to-member-scm-L34-L36) to be passed to the function, that will actually do more than calculate the TV . The function should check the arguments and produce the possible outputs and set caluclate the TV

prateeksaxena2809:
I have updated the PR and applied the changes as discussed in the comments. Please provide feedback.

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
```diff
@@ -0,0 +1,50 @@
...
+; Evaluation To Member Formula
+; -----------------------------------------------------------------------------
+
+; -----------------------------------------------------------------------------
+; Side-effect: TruthValue of the new link stays the same
+; -----------------------------------------------------------------------------
+
+(define (pln-formula-evaluation-to-member DBC)
+	(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+		(car (cdr (cog-get-all-nodes DBC)))
```
williampma: Just do `(gadr DBC)`.  There's no need to get all the nodes first.

`gar`, `gadr`, etc are written in `scm/utilities.scm` for getting stuff out of basic links.

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
```diff
@@ -0,0 +1,50 @@
...
+		(SatisfyingSetLink
+			(VariableNode "$X")
+			(EvaluationLink
+				(car (cog-get-all-nodes DBC))
+				(ListLink
+					(cons
+						(VariableNode "$X")
+						(cdr (cdr (cog-get-all-nodes DBC)))))))))
```
williampma: Unfortunately there is no `gar`, `gadr` equivalent for this, so it would be `(cdr (cog-outgoing-set (gdr DBC)))`.  It would be a bit faster than getting all the nodes.

williampma: Actually, none of what I just wrote, and what you wrote, handle EvaluationLink like this:
```
EvaluationLink
   PredicateNode "laughed"
   ConceptNode "he"
```
which is a valid EvaluationLink...  I guess it requires some checking.

prateeksaxena2809: I checked the code with one member EvaluationLink and it seems to work for
```
(EvaluationLink
    (PredicateNode "laughed")
    (ListLink
        (ConceptNode "he")))
```
But not for
```
(EvaluationLink
    (PredicateNode "laughed")
    (ConceptNode "he"))
```
I'll add a check for such links. I'll also replace the `car, cdr` versions with the `gar, gdr` versions.

prateeksaxena2809:
I think the correct MemberLink for:

```
(EvaluationLink
    (PredicateNode "laughs")
    (ConceptNode "he"))
```

should be

```
(MemberLink
    (ConceptNode "he")
    (SatisfyingSetLink
        (VariableNode "$X")
        (EvaluationLink
            (PredicateNode "laughs")
            (VariableNode "$X"))))
```

The updated code creates this MemberLink for above EvaluationLink. Is this the right conversion?
I have also replaced the ```car, cadr ``` conversions with ```gar, gdr``` conversions as suggested.

ngeiswei:
I think we really ought to change how SatisfyingSetLink is defined in the wiki (and subsequently how it is implemented here), and use the version that is defined in the PLN book. With that definition we'd have merely

```
GeneralEvaluationToMemberRule:

EvaluationLink
    A
    B
|-
MemberLink
    B
    SatisfyingSetLink
        A
```

(same thing for GeneralMemberToEvaluationRule)

ngeiswei:
I think the MemberToInheritance rule
```
MemberLink( B C )
|-
InheritanceLink( B C )
```
is not right.

In the PLN book, it is

```
MemberLink( B C )
|-
InheritanceLink( {B} C )
```
where B is a singleton. See page 112. I don't know how to define a singleton though...

linas:
nill you seem to be flip-flopping on the satisfying set link. Or maybe I don't understand.... note also that there is a disucssion page in back of the wiki, perhaps drafts should be placed there.

bgoertzel:
The Singleton {B} would be written

SetLink
    B


I guess...

On Thu, Mar 19, 2015 at 10:16 PM, ngeiswei <notifications@github.com> wrote:

> I think the MemberToInheritance rule
>
> MemberLink( B C )
> |-
> InheritanceLink( B C )
>
> is not right.
>
> In the PLN book, it is
>
> MemberLink( B C )
> |-
> InheritanceLink( {B} C )
>
> where B is a singleton. See page 112. I don't know how to define a
> singleton though...
>
> —
> <https://github.com/opencog/opencog/pull/1422#issuecomment-83596231>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
```diff
@@ -0,0 +1,58 @@
...
+		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
+			(gadr DBC)
+			(SatisfyingSetLink
+				(VariableNode "$X")
+				(EvaluationLink
+					(gar DBC)
+					(ListLink
+						(cons
+							(VariableNode "$X")
+							(cdr (cog-get-all-nodes (gdr DBC))))))))))
```
williampma: I probably should have mentioned another reason to not use `cog-get-all-nodes` for this.  It is possible the arguments for the Predicate inside the `EvaluationLink` are not nodes, but instead is another link.  For example, current there's this being created by the R2L pipeline:

```
EvaluationLink
    PredicateNode "seems"
    ListLink
        InheirtanceLink
             ConceptNode "he"
             ConceptNode "happy"
```

Looking at the old python PLN rule, the output looks to be
```
MemberLink
    InheirtanceLink
         ConceptNode "he"
         ConceptNode "happy"
    SatisfyingSetLink
         VariableNode $X
         EvaluationLink
              PredicateNode "seems"
              ListLink
                   VariableNode $X
```

Although it looks like the proper output of the EvaluationToMember rule is currently being discussed...

P.S.  The old version of the rule actually generate a MemberLink for each argument, not just the first one.

prateeksaxena2809: I think replacing ```cog-get-all-nodes``` with ```cog-outgoing-set``` should be able to handle links as arguments of predicate. Also, I am not sure about the expected output of EvaluationLink generating MemberLink for each argument.
For Example:-
```
(EvaluationLink (stv 0.9 0.9)
    (PredicateNode "give")
    (ListLink
        (ConceptNode "Ram")
        (ConceptNode "Rahim")
        (ConceptNode "Book")))
```
should give us
```
(MemberLink
    (ConceptNode "Ram")
    (SatisfyingSetLink
        (VariableNode "$X")
        (EvaluationLink
            (PredicateNode "give")
            (ListLink
                (VariableNode "$X")
                (ConceptNode "Rahim")
                (ConceptNode "Book")))))
```
and
```
(MemberLink
    (ConceptNode "Rahim")
    (SatisfyingSetLink
        (VariableNode "$X")
        (EvaluationLink
            (PredicateNode "give")
            (ListLink
                (ConceptNode "Ram")
                (VariableNode "$X")
                (ConceptNode "Book")))))
```
and
```
(MemberLink
    (ConceptNode "Book")
    (SatisfyingSetLink
        (VariableNode "$X")
        (EvaluationLink
            (PredicateNode "give")
            (ListLink
                (ConceptNode "Ram")
                (ConceptNode "Rahim")
                (VariableNode "$X")))))
```
Does this look right? Also, if yes, should the TVs associated with all the links created be same or should they be divided in a particular precedence?

prateeksaxena2809: @williampma Also, can you suggest  a resource for understanding the difference between implicit binding and explicit binding. I am trying to follow the discussion on #1424, regarding the correct conversion of evaluation link to member links, in which these terms are being used frequently.

ngeiswei:
Linas, I'm not flip-flopping on the SatisfyingSetLink definition, it's been defined without binding in the PLN book and I don't see any reason to change that. The note at the end of the SatisfyingSetLink page

http://wiki.opencog.org/w/SatisfyingSetLink#Notes

suggests that you consider SatisfyingSetLink to be a sort of predicate constructor, but that is not its original purpose. It's purpose is merely to convert a predicate into a concept, thus rules like

```
EvaluationLink
    P
    X
|-
MemberLink
    X
    SatisfyingSetLink
        P
```

a predicate is supposed to have it's variables already bound (otherwise, there is no unambiguous way to use it). So rather than using SatisyfingSetLink as a predicate constructor let's use a dedicated predicate constructor, which I think should be called LambdaLink, cause there is no better name for such a constructor.

ngeiswei:
Ben, SetLink seems right. I barely knew it existed, I guess more PLN rules should be defined to thoroughly take care of it.

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
AmeBel: @prateeksaxena2809 the 3 MemberLinks are correct in the sense that was what the python PLN used to do. In the Python implementation of PLN the TV of the EvaluationLink was used for the MemberLink, I am not sure if that is correct though.

AmeBel:
@ngeiswei what is the difference between (SetLink B) and (ConceptNode B)?

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
williampma: @prateeksaxena2809 You meant those implicit BindLink and explicit BindLink stuff?  I just took it literally; ie. using a VariableNode without first declaring it inside a BindLink is "implicit", like
```
  (InheritanceLink
     (VariableNode "$X")
     (ConceptNode "dog"))
```

And since SatisfyingSetLink inherit from BindLink, the same applies.  ie.
```
    (SatisfyingSetLink
        (VariableNode "$X")
        (EvaluationLink
            (PredicateNode "give")
            (ListLink
                (ConceptNode "Ram")
                (ConceptNode "Rahim")
                (VariableNode "$X"))))
```
is "explicitly" binding `VariableNode $X`.

At least that what I think Nil et. al. meant!  If they meant something else they should speak up! :)

ngeiswei:
@AmeBel (SetLink (ConceptNode "B")) is the singleton {ConceptNode "B"}, or you may say

for all X, Y such that X is different than Y

```
MemberLink <1, 1>
    X
    SetLink
        X

and

MemberLink <0, 1>
    Y
    SetLink
        X
```

this is semi formal, this has to be formally defined, and the fuzzy side of it has to be taken care of (cause the notion of different/dissimilar is fuzzy in PLN).

opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
ngeiswei: Well, I'm not the one who added the binding into SatisfyingSetLink (I think it's Linas). I see no reason to have binding inside SatisfyingSetLink, we already need a LambdaLink for that for let's not duplicate.

I've pushed a merge for a LambdaLink #1440 

OpenCog is a very complicated project, we need to simplify and make things as clear as we possibly can.

ngeiswei: Regarding the implicit vs explicit. The debate initially was whether we want hyperpgraphs like

```
EvaluationLink
    P
    X
```
to represent a predicate or a body function. The tradition in computer science has been to have a function constructor like lambda in order to turn a body function into a function. The function constructor allows to know what arguments the function takes, in which order, etc. Doing that implicitly I believe if risky, and then confusing for humans as well, that is why I suggest to use a LambdaLink instead (formerly called BindLink, but let's be honest, LamdbaLink is a much better name! right?). And BindLink can be used as the base type for all that pertains to binding, forall, lambda, etc.

bgoertzel: I'm starting to agree with you....   Implicit binding is likely to confuse
people, and right now we should value minimizing confusion over a somewhat
dubiously significant savings in memory....

LambdaLink seems a good step...


On Fri, Mar 20, 2015 at 6:42 PM, ngeiswei <notifications@github.com> wrote:

> In opencog/reasoning/RuleEngine/rules/pln/evaluation-to-member.scm
> <https://github.com/opencog/opencog/pull/1422#discussion_r26831162>:
>
> > +			(gdr DBC)
> > +			(SatisfyingSetLink
> > +				(VariableNode "$X")
> > +				(EvaluationLink
> > +					(gar DBC)
> > +					(VariableNode "$X"))))
> > +		(MemberLink (stv (cog-stv-strength DBC) (cog-stv-confidence DBC))
> > +			(gadr DBC)
> > +			(SatisfyingSetLink
> > +				(VariableNode "$X")
> > +				(EvaluationLink
> > +					(gar DBC)
> > +					(ListLink
> > +						(cons
> > +							(VariableNode "$X")
> > +							(cdr (cog-get-all-nodes (gdr DBC))))))))))
>
> Regarding the implicit vs explicit. The debate initially was whether we
> want hyperpgraphs like
>
> EvaluationLink
>     P
>     X
>
> to represent a predicate or a body function. The tradition in computer
> science has been to have a function constructor like lambda in order to
> turn a body function into a function. The function constructor allows to
> know what arguments the function takes, in which order, etc. Doing that
> implicitly I believe if risky, and then confusing for humans as well, that
> is why I suggest to use a LambdaLink instead (formerly called BindLink, but
> let's be honest, LamdbaLink is a much better name! right?). And BindLink
> can be used as the base type for all that pertains to binding, forall,
> lambda, etc.
>
> —
> <https://github.com/opencog/opencog/pull/1422/files#r26831162>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

ngeiswei: Excellent! I have subsequently merged #1440

AmeBel:
We have been using SetLink in a different way in the natural language generation pipeline.

AmeBel:
I am assuming the definition is the same as https://en.wikipedia.org/wiki/Singleton_(mathematics)

ngeiswei:
"We have been using SetLink in a different way in the natural language generation pipeline."

Oh, there's over 25 files using SetLink... I'll have a look at them.

Yes, same definition of singleton as in wikipedia (albeit perhaps the fuzzy aspect, if we decide so).

ngeiswei:
@AmeBel the use of SetLink I've seen in the code so far doesn't contradict Ben's suggestion to use SetLink to build a singleton. The only missing thing are additional PLN rules to to convert SetLink into concept, that kind of things.

AmeBel:
r2l's outputs are wrapped in a SetLink, which we assumed represents a Set (not a singleton set). For example, 

```

(SetLink
  (InheritanceLink (stv 0.99000001 0.99000001)
     (ConceptNode "this@81fdd925-b1f2-4963-a206-083f85523bbd" (stv 0.001 0.99000001))
     (ConceptNode "this" (ptv 0.001 0.99000001 1))
  )
  (InheritanceLink (stv 0.99000001 0.99000001)
     (ConceptNode "set@f95e8ea9-eedb-4540-86b4-4fd2602acb9e" (stv 0.001 0.99000001))
     (ConceptNode "set" (ptv 0.001 0.99000001 1))
  )
  (InheritanceLink (stv 0.99000001 0.99000001)
     (ConceptNode "this@81fdd925-b1f2-4963-a206-083f85523bbd" (stv 0.001 0.99000001))
     (ConceptNode "set@f95e8ea9-eedb-4540-86b4-4fd2602acb9e" (stv 0.001 0.99000001))
  )
  (ImplicationLink (stv 0.99000001 0.99000001)
     (PredicateNode "is@91f3d628-c55d-4f02-bea3-4a354b3d9eff" (stv 0.001 0.99000001))
     (PredicateNode "be" (ptv 0.001 0.99000001 1))
  )
  (InheritanceLink (stv 0.99000001 0.99000001)
     (PredicateNode "is@91f3d628-c55d-4f02-bea3-4a354b3d9eff" (stv 0.001 0.99000001))
     (ConceptNode "present" (stv 0.001 0.99000001))
  )
  (InheritanceLink (stv 0.001 0.99000001)
     (InterpretationNode "sentence@4c283cb8-c130-427c-8602-8eff4cc4f4f8_parse_0_interpretation_$X")
     (ConceptNode "DeclarativeSpeechAct" (stv 0.001 0.99000001))
  )
)

```
is the final output of running `(nlp-parse "This is a set.")`

prateeksaxena2809:
I am working on making all the three changes discussed in this thread namely
1. Multiple member link creation
2. Creating Lambda Links in addition to the Satisfying Set Links
3. Incorporating Singleton set links in the creation of Inheritance Links.

I have been able to implement 2 and 3, but while trying to [implement 1](https://gist.github.com/prateeksaxena2809/740e40542dfe1092d9ce), I am running into an issue. When I run the code by calling the pln-formula directly with the links as arguments(as mentioned below), the code works as expected.
```
(pln-formula-evaluation-to-member atom)
```
 But when i run it using cog-bind(as mentioned below), the code crashes with [a segmentation fault](https://gist.github.com/prateeksaxena2809/040144189df195dfb16d).
```
(cog-bind pln-rule-evaluation-to-member)
```
Any ideas on what could be causing this error?

linas:
Please be aware that the SatisfyingSet/LambdaLink discussion has not concluded, and that these are in the process of being designed and are subject to change. Nil will be busy for a few weeks, and then the discussion will resume: the discussion may last only a day, or it may last longer. The design of LambdaLink currently impacts multiple other systems, including #1432  #1445 #1460 -- is there something else that can be worked out?

prateeksaxena2809:
Ok, I will remove the lambda link for now and we can merge it in later once the implementation is finalized.
However, I am still getting segmentation fault, so the current error does not seem to be related to that.

williampma:
I have a feeling that ExecutionOutputLink requires the scm function to return an atom.  Since your scheme function for-each loop returns nothing, it might be what caused the crash.

Not 100% sure that's the cause though.  I haven't have time to try running it yet.

prateeksaxena2809:
The new code creates multiple member links for multiple argument predicates. I have removed Lambda Links and Singleton SetLink for now. Please give feedback.

prateeksaxena2809:
hey @AmeBel and @williampma , Could you have a look at the latest commits and let me know if it looks good or any changes need to be made?

Also, please note that I haven't added singleton because in order to infer the Socrates example correctly, the deduction rule should be changed after that because currently it is,
```
(A->B & B->C) -> A->C
```
which to incorporate for singleton should be
```
(A->B & {B} -> C) -> {A} ->C
```
But this is not the general deduction rule.

A work around would be to added another deduction rule for InheritanceLink created from MemberLinks. What according to you would be the proper way to incorporate this ?

AmeBel:
i will modify the [deductionrule](https://github.com/AmeBel/opencog/blob/deductionrule/opencog/reasoning/RuleEngine/rules/pln/deduction.scm) to take SatisfyingSetLinks as it is similar to ConceptNode http://wiki.opencog.org/w/SatisfyingSetLink. But the deduction rule should also work, check https://gist.github.com/AmeBel/8b9dc2e4257f2dfc1b8f for why

on a quick look it doesn't seem that the MemeberToInheritance rule's formula is not same as https://github.com/opencog/opencog/blob/master/opencog/python/pln/rules/formulas.py#L284

prateeksaxena2809:
The current deduction rule takes SatisfyingSetLinks similarly. For example, in the current link
```
(InheritanceLink (stv 0.99000001 0.99000001)
      (ConceptNode "men" (stv 0.001 0.99000001))
      (SatisfyingSetLink
         (VariableNode "$X")
         (EvaluationLink
            (PredicateNode "breathe" (stv 0.001 0.99000001))
            (ListLink
               (VariableNode "$X")
               (ConceptNode "air" (stv 0.001 0.99000001)))))))
```
and
```
(InheritanceLink (stv .99 .99)
  (ConceptNode "Socrates")
  (ConceptNode "men")
```
the deduction rule works correctly and gives
```
(InheritanceLink (stv 0.99000001 0.99000001)
      (ConceptNode "Socrates" (stv 0.001 0.99000001))
      (SatisfyingSetLink
         (VariableNode "$X")
         (EvaluationLink
            (PredicateNode "breathe" (stv 0.001 0.99000001))
            (ListLink
               (VariableNode "$X")
               (ConceptNode "air" (stv 0.001 0.99000001)))))))
```
I was talking about the deduction after using singleton SetLinks in the creation of inheritance links.

I have checked and corrected the TV formulae for the inheritance to member links and member to inheritance links.

AmeBel:
I think it is best we wait for https://github.com/opencog/opencog/issues/1445 to be resolved before moving forward, we don't want to solve a problem that might not exist :-) . In additon, there has been a significant additons https://github.com/opencog/opencog/pull/1477 https://github.com/opencog/opencog/pull/1474 that could change how we would want to define rules.

In the mean time you could choose some of the syllogisms listed @ http://wiki.opencog.org/w/List_of_Test_Syllogisms#Initial_test_list and work with [NL-comprehension pipeline](http://wiki.opencog.org/w/Running_Relex2Logic_with_OpenCog) to get the output that would be feed to PLN.

prateeksaxena2809:
@AmeBel, Sounds good. I'll start looking at this.

linas:
@AmeBel @williampma The changes that you have noticed should have very little effect on how you write rules. What they do is this:

* The syntax of a BindLink is checked when it is added to the atomspace. You cannot add bad BindlInks to the atomspace; an exception will be thrown.
* The contents of the BindLink are unpacked into C++ structs, and cached. Thus, the pattern matcher no longer needs to unpack each time it runs. That should make things ever so slightly faster.
* There is now a SatisfactionLink, which is used without the ImplicationLink. Use it if you only want to run the pattern matcher, but don't care about the second half of the ImplicationLink (e.g. if our callback records the variable groundings only, and doesn't care about the shape of the clauses.)  This makes thins slightly easier for a common case.

So, these changes have very little effect on PLN at this time.  They just lay groundwork for some other ideas I'm pursuing.

misgeatgit:
@prateeksaxena2809 can you make a new pull request of this in conformance with the recent refactorings?

prateeksaxena2809:
@misgeatgit Sure. will do.

prateeksaxena2809:
Submitted PR #1593 in conformance with recent directory structure changes. If that looks ok, should we close this PR?

AmeBel:
yes you can close this

PR #1426, opened 2015-03-17
Fixed choose rule and started revamping PLNCommonUTest

misgeatgit:
Now DefaultForwardChainerUTest Passes for me.


misgeatgit:
@williampma  and others, Review these commits before they are merged.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,175 @@
...
+    if (target == Handle::UNDEFINED or NodeCast(target))
+        throw InvalidParamException(TRACE_INFO,
+                                    "Needs a target atom of type LINK");
+    HandleSeq chosen_bindlinks;
+
+    if (LinkCast(target)) {
+        AtomSpace rule_atomspace;
+        SchemeEval* sceval = new SchemeEval(&rule_atomspace);
+        //Handle target_cpy=rule_atomspace.addAtom(target); xxx this doesn't work
+        Handle target_cpy = sceval->eval_h(SchemeSmob::to_string(target)); //xxx this works
```
williampma: Really... wonder what's the difference.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,175 @@
...
+        HandleSeq bindlinks;
+        for (Handle hm : matches) {
+            HandleSeq hs = get_rootlinks(hm, &rule_atomspace, BIND_LINK);
+            for (Handle hi : hs) {
+                if (find(bindlinks.begin(), bindlinks.end(), hi) == bindlinks.end()) {
+                    bindlinks.push_back(hi);
+                }
+            }
+        }
+        delete sceval; //delete to use SchemeEval with another atomspace.this might be a hack.
```
williampma: So this is the alternate fix for the crash?

misgeatgit: Yea, letting scheme be aware of atomspace change.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,175 @@
...
+        HandleSeq matches = imp.result_list;
+
+        if (matches.empty()) {
+            logger().debug(
+                    "No matching BindLink was found.Returning empty vector");
+            return vector<Rule*> { };
+        }
+
+        HandleSeq bindlinks;
+        for (Handle hm : matches) {
```
williampma: Is this loop trying to get all BindLink containing hm?

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
misgeatgit: BTW, there are some typos and couts which I want to clean in another commit? maybe.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,175 @@
...
 }
 
-HandleSeq DefaultForwardChainerCB::choose_input(FCMemory& fcmem) {
-	Handle htarget = fcmem.get_cur_target();
-	//get everything associated with the target handle
-	HandleSeq inputs = as_->getNeighbors(htarget,true,true,LINK,true);
-	return inputs;
+HandleSeq DefaultForwardChainerCB::get_rootlinks(Handle htarget, AtomSpace* as,
+                                                 Type link_type,
+                                                 bool subclasses)
```
williampma: Doxygen style commenting this function please.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
misgeatgit: Yes,Getting a bind link whose once of its elements structurally matches with the target. Which after we validate the match is actually a premise part.

tests/reasoning/RuleEngine/PLNCommonsUTest.cxxtest
```diff
@@ -16,146 +16,231 @@
...
+            "(ListLink"
+            "(InheritanceLink"
+            "(VariableNode \"$A\")"
+            "(VariableNode \"$B\"))"
+            "(InheritanceLink"
+            "(VariableNode \"$B\")"
+            "(VariableNode \"$C\"))"
+            "(InheritanceLink"
+            "(VariableNode \"$A\")"
+            "(VariableNode \"$C\"))))))";
```
williampma: Wow... please format this :)

misgeatgit: What is the standard for formatting it?

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,175 @@
...
 }
 
-//choose rule based on premises of rule matching the target
-//uses temporary atomspace to limit the search space and avoid
+/**
+ * choose rule based on premises of rule matching the target
+ * uses temporary atomspace to limit the search space and avoid
+ *
+ * @return a vector of chosen rules
+ **/
```
williampma: I don't think ending the comment block with **/ is Doxygen style.

tests/reasoning/RuleEngine/PLNCommonsUTest.cxxtest
williampma: Just format the string inside with proper spacing like in Scheme.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
williampma: Add this comment to the code.

williampma:
Looks OK to me.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
linas: Yeah, there is something seriously wrong here; the two different forms should result in exactly the same atomspace contents, with the direct-copy version being much faster than the to-string version.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
linas: again, using new and delete is going to be much slower than using get_evaluator. However, the current get_evaluator code perhaps fails to deal correctly with deleted atomspaces.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
linas: I just experimented: if I use `Handle target_cpy=rule_atomspace.addAtom(target);` instead of the string version, the test still passes, for me.

If I print out, using `target->toString()` and `target_cpy->toString()`, both variants print the same thing.

According to the atomspace design, both `Handle target_cpy = sceval->eval_h(SchemeSmob::to_string(target));` and `target_cpy=rule_atomspace.addAtom(target);` should result in exactly the same atomspace contents. If they are not the same, then there is a bug somewhere .. and thaat bug would need to be fixed.  The second form will, of course be much much faster than the first.

opencog/reasoning/RuleEngine/rule-engine-src/pln/DefaultForwardChainerCB.cc
```diff
@@ -25,111 +25,186 @@
...
+            }
+        }
+        delete sceval; //delete to use SchemeEval with another atomspace.this might be a hack.
+
+        //push back handles to main atomspace
+        for (Handle h : bindlinks) {
+            //auto bindlink = as_->addAtom(h);
+            SchemeEval *sc = new SchemeEval(as_);
+            auto bindlink = sc->eval_h(SchemeSmob::to_string(h));
+            chosen_bindlinks.push_back(bindlink);
```
linas: Three problems here:
1) you forgot a delete
2) this string thing is nutty, the direct add should work perfectly i.e. `Handle bindlink = as_->addAtom(h);` should work
3) But `Handle bindlink = as_->addAtom(h);` segfaults

I'm investigating this now.

linas: I think I found it .. the atom table indexes are failing to work correctly. I think they need an explicit compare function added. What a bug... amazing that the unit tests don't find it.  Very strange.

linas: Yep.  Found it it is fixed in fc31a4b24e6a3b35dacebaf7865c90c261283581  and a unit test tests in in 11ef7f927d8f14ebb34cb4f8d8913f0ff9de1342

PR #1430, opened 2015-03-18
Create the real fox for bug #1380

linas:
In fact, the bug was a subtle atomspace bug, involving multi-atomspace support, and previously unseen and untested.


williampma:
Thanks, Linas.

PR #1440, opened 2015-03-20
Add LambdaLink as well as a wiki page

ngeiswei:
http://wiki.opencog.org/w/LambdaLink

I've also modified

http://wiki.opencog.org/w/BindLink

### DIFF
--- a/opencog/atomspace/atom_types.script
+++ b/opencog/atomspace/atom_types.script
@@ -76,6 +76,9 @@ EXISTS_LINK <- BIND_LINK
 AVERAGE_LINK <- FORALL_LINK
 SATISFYING_SET_LINK <- BIND_LINK
 
+// Function constructor (predicate and schema)
+LAMBDA_LINK <- BIND_LINK
+
 SCHOLEM_LINK <- ORDERED_LINK
 
 // Basic first-order logic operations


opencog/atomspace/atom_types.script
```diff
@@ -76,6 +76,9 @@ EXISTS_LINK <- BIND_LINK
 AVERAGE_LINK <- FORALL_LINK
 SATISFYING_SET_LINK <- BIND_LINK
 
+// Function constructor (predicate and schema)
+LAMBDA_LINK <- BIND_LINK
```
linas: Umm,

I think you got this backwards.  I think you want this:
```
LAMBDA_LINK <- ORDERED_LINK
BIND_LINK <- LAMBDA_LINK
FORALL_LINK <- LAMBDA_LINK
EXISTS_LINK <- LAMBDA_LINK
SATISFYING_SET_LINK <- LAMBDA_LINK
```
right? that makes bind-link a special case of the lambda, that the pattern matcher "understands" but has no other specific "meaning".

Also, perhaps BindLink should be renamed to GroundingLink or something like that.

ngeiswei: I don't really understand the meaning of BindLink to be honest. The way I understood it is, a link of the type

BindLink
    Variables
    Expression

No more than that. But a lambda is a bit more than that, it abstract away an expression to turn it into a function (that can then be used for application). And a quantifier is not a lambda, but it also has the same binding pattern in common. So that is why I thought it makes sense to have both lambda and quantifiers inherits bindLink.

As for SatisfyingSetLink I don't think it should do any binding at all.

Rather than renaming BindLink by GroundingLink, perhaps you want to add a new link GroundingLink that inherits BindLink, and that would be used by the pattern matcher, that makes sense to me.

linas: OK, there is a fair amount of confusion going on here. You keep thinking and talking in  terms of functional programming and functional evaluation, whereas I am thinking of and talking about satisfiability and satisfaction. These two notions are very different from each other, and this is at the root of the confusion.

Satisfiability is about taking an expression that contains variables, and discovering values for those variables, that renders the expression "true".  So perhaps BindLink should be called "SatisfiabilityLink". The "meaning" of a BindLink is that it encloses an expression that may or may not be true, depending on whether there exist items in the universe that, when substituted for the variables, render the expression true.  Note that there might be multiple values that render a BindLink true: these form a set. Thus, perhaps BindLink should be called "SatisfiabilitySetLink" .  So, in this sense, a BindLink is quite similar to a SatisfyingSetLink.  The main difference between the two is that the BindLink requires that an ImplicationLink be inside of it.  Perhaps this is a design mistake.  Thus, today, all bindLinks must have the following form:

```
BindLink
    VariableDecls x
    ImplicationLink
          P(x)
          Q(x)
```

I'm guessing that what this means, in terms of "classical" PLN, is this:
```
 ForAllLink
     VariableDecls x
     ImplicationLink
         SatisfyingSetLink
               P(x)
          Q(x)
```
Note the indentation of P w.r.t. Q. Basically, if some values satisfy P(x), i.e. if some values belong to the sattisfying set of P(x), then it is true that those values imply Q(x).  The (cog-evaluate!) function does two things: it finds all values v satisfying P(x), (i.e. it finds all values v such that P(v) is true) and then it creates Q(v)  where Q(v) means Q(substitute v for x).

Note in particular, the below is wrong:
```
BindLink
    Variables
    Expression
```
The above is NOT a valid BindLink. the Expression MUST be an ImplicationLink for it to be valid.

linas:
I just opened issue #1445 to try to clarify the distinction between BindLinks and SatsifactionLinks.

A satisfactionLink should be a kind of LambdaLink that is "true" when the expression is satisfiable.  Thus, SatsifactionLink is maybe almost identical to a SatisfyingSetLink, except that ti requires explicit variable binding...!?

inflector:
Some change in this merge broke AllPythonModuleTests:

test_all (test_rules_new.PLNUnitTester) ... FAIL

linas:
AllPythonModuleTest has not passed in years. Its broken somehow.  I though you were working on fixing it?

inflector:
It was passing in build 372 and stopped passing in 373 after Nil's merge built:

AllPythonModuleTests has been passing for me and buildbot at least,  since my first week when I put in skips for the failing tests. The PLN tests were one of the relatively few remaining Python unit tests. 

If you still getting failures, what are they?

linas:
I don't know I just get 
```
The following tests FAILED:
	 95 - AllPythonModuleTests (Failed)
```
and there is no easy way of figuring out what went wrong with spending half an hour trying to arrange all the pythonpaths to be just so, and decoding the smelltest infrastructure.  There should be some easy way of just running the ptyhon tests from the command line.

williampma:
Actually, I think AllPythonModuleTests is failing randomly rather than because of a specific change.  Maybe it's because one of the test inside is assuming the atomspace will return the atoms in the same order everytime, which is no longer true after recent changes to the AtomTable.  The atoms traversal and incoming set order is no longer deterministic.

opencog/atomspace/atom_types.script
ngeiswei: OK, Linas, there was indeed a confusion on my part regarding the semantics of the BindLink, basically I didn't imaging that the implicationlink was mandatory. So now I understand the binding exaclt connecting P and Q.

Yes, it feels to me too that the use of the ImplicationLink is not right, I'll look carefully at issue #1445

linas:
Hey Nil,

you changed: http://wiki.opencog.org/w/SatisfyingSetLink -- I don't mind changes, but now I have no clue what a SastisfyingSetLink is, what it does, or how it works.  It may as well be written in Latin: I can sort-of-ish guess, but I don't really understand it.  Its confusing.

The old wiki page had a bunch of examples so that you could see how it actually worked, and see how to use it.  Now it says that it "is used to convert a predicate into a concept."  WTF is a predicate? I really don't know. WTF is a concept? I really don't know. These are crazy mystery words that need to be defined.  I mean, I know what a predicate is classical logic, and I know that, in opencog, predicates are sort-of-like conditional probabilities, but different. I still do not really understand predicates in OpenCog. and if I don't understand them... how about the poor shmuck who is reading that page for the first time?

bgoertzel:
I can remedy this but not right now, I have very little time online for the
next few days...

On Tue, Mar 24, 2015 at 4:35 AM, Linas Vepštas <notifications@github.com>
wrote:

> Hey Nil,
>
> you changed: http://wiki.opencog.org/w/SatisfyingSetLink -- I don't mind
> changes, but now I have no clue what a SastisfyingSetLink is, what it does,
> or how it works. It may as well be written in Latin: I can sort-of-ish
> guess, but I don't really understand it. Its confusing.
>
> The old wiki page had a bunch of examples so that you could see how it
> actually worked, and see how to use it. Now it says that it "is used to
> convert a predicate into a concept." WTF is a predicate? I really don't
> know. WTF is a concept? I really don't know. These are crazy mystery words
> that need to be defined. I mean, I know what a predicate is classical
> logic, and I know that, in opencog, predicates are sort-of-like conditional
> probabilities, but different. I still do not really understand predicates
> in OpenCog. and if I don't understand them... how about the poor shmuck who
> is reading that page for the first time?
>
> —
> <https://github.com/opencog/opencog/pull/1440#issuecomment-85183346>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

ngeiswei:
Linas, I wanted to use predicate and concept as defined in the type system but it is not really correct at the present time, I decided to let it undefined. I'll get back to all that after 10 days...

Issue #1424, opened 2015-03-17
EvaluationToMemberRule output choice


AmeBel:
@bgoertzel  @moikle
Let us assume we have 
```
(EvaluationLink
    (PredicateNode "eat")
    (ListLink
         (ConceptNode "dog")
         (ConcpetNode "dragon")
))
```
There are two possible outputs for EvaluationToMember Rule

```
(MemberLink
	(ConceptNode "dog")
	(SatisfyingSetLink
		(VariableNode "$X")
		(EvaluationLink
			(PredicateNode "eat")
			(ListLink
				(VariableNode "$X")
				(ConceptNode "dragon")))))

(MemberLink
	(ConceptNode "dragon")
	(SatisfyingSetLink
		(VariableNode "$X")
		(EvaluationLink
			(PredicateNode "eat")
			(ListLink
				(ConceptNode "dog")
				(VariableNode "$X")))))
```

How should the chaining be performed? and what is the formula associated with it?

moikle:
Amen,

Good question as both approaches make sense to me. It seems to me that the original EvaluationLink contains more information than either of your two suggestions by themselves. Instead the original EvaluationLink provides the information for constructing both of your outputs.

In terms of handling the chaining, I just don’t know. Do we wish to create both outputs since they provide different senses, or do we wish to merge the two TVs?

Ben?

—matt

bgoertzel:
The rule should produce both outputs IMO.   The chainer can then choose
which one to use via whatever heuristics it wants, e.g. by which one has
nodes with higher STI in a more prominent position...

On Tue, Mar 17, 2015 at 1:54 PM, moikle <notifications@github.com> wrote:

> Amen,
>
> Good question as both approaches make sense to me. It seems to me that the
> original EvaluationLink contains more information than either of your two
> suggestions by themselves. Instead the original EvaluationLink provides the
> information for constructing both of your outputs.
>
> In terms of handling the chaining, I just don’t know. Do we wish to create
> both outputs since they provide different senses, or do we wish to merge
> the two TVs?
>
> Ben?
>
> —matt
>
> > On Mar 17, 2015, at 1:07 PM, Amen Belayneh <notifications@github.com>
> wrote:
> >
> > @bgoertzel <https://github.com/bgoertzel> @moikle <
> https://github.com/moikle>
>
> > Let us assume we have
> >
> > (EvaluationLink
> > (PredicateNode "eat")
> > (ListLink
> > (ConceptNode "dog")
[...]

PR #1442, opened 2015-03-20
Change SatisfyingSetLink to matches the PLN book definition

ngeiswei:
Update as well the wiki page
http://wiki.opencog.org/w/SatisfyingSetLink

This is related to issue #1424 as it provides a modified (fixed according to the PLN book) definition of SatisfyingSetLink.

I should note that almost all instances of SatisfyingSetLink in the code was already correct (following PLN), only one in opencog/opencog/nlp/relex2logic/rule-helpers.scm had to be corrected.


AmeBel:
@anitzkin please review this

ngeiswei:
Oh, I've mechanically merge that pull request, of course I wanted you to have a look at it first. Anyway, it's pretty straight forward and I'm sure we all already it's the right thing to do.

linas:
Nil,

The wiki page for SatsifyingSetLink is a cut-n-paste from one of Ben's emails, that he wrote years after the PLN book .. so clearly his thanking had changed. Or he mis-rememebred things.  But .. also you are contradicting yourself: you gave three or four examples where implicit variables are bad/wrong, in issue #1424 I quote: "I can think of 3 reasons not to use implicit BindLink/LambdaLink:" so -- why did you change your mind? Why is it better that satisfyingset use implicit, instead of explicit variables?  I don't get it.

bgoertzel:
First, a general comment: In working out the PLN theory described in the
PLN and Real World Reasoning books, generally I thought a lot about the
probabilistic (and sometimes fuzzy-probabilistic) truth value formulas ---
but not that much about the types and quantifiers involved.   So PLN as
formulated in those books is pretty "loose", it's not a fully formalized
logic system.   I've been planning for years to spend a week or two fully
formalizing it as a logic system, and getting my friend Min Jiang (a
Chinese logician) to prove some theorems about it....  But this hasn't
happened...  So what I see now is that this effort to make a rigorous type
system for the Atomspace, is basically requiring PLN to be more fully
formalized as a logic system.   It seems that if PLN is formalized to work
on an Atomspace with a rigorous type system, we'll be 90% of the way there
to a fully formalized PLN.....  (Probably we should have worked out and
published a full formalization of PLN long ago, but instead we proceeded
pragmatically, basically trying to make a system that would make
humanly-simple inferences relatively short and simple, and would give
plausible truth value outputs in commonsensical cases...)

About implicit vs. explicit quantifiers, I'm coming to agree w/ Nil's
earlier-expressed opinion that implicit quantification adds confusion.  If
we're gonna require explicit types then I guess we should require explicit
quantification also.   Leaving out some of the quantification makes
examples more concise -- but We don't really need a lot of "syntactic
sugar" in the Atomspace, since it's not a programming language per se, but
rather it's "the guts of a programming language", intended to be used with
the user-level syntax of some other programming language like Scheme,
python or Haskell...

-- Ben



On Sat, Mar 21, 2015 at 7:09 AM, Linas Vepštas <notifications@github.com>
wrote:

> Nil,
>
> The wiki page for SatsifyingSetLink is a cut-n-paste from one of Ben's
> emails, that he wrote years after the PLN book .. so clearly his thanking
> had changed. Or he mis-rememebred things. But .. also you are contradicting
> yourself: you gave three or four examples where implicit variables are
> bad/wrong, in issue #1424 <https://github.com/opencog/opencog/issues/1424>
> I quote: "I can think of 3 reasons not to use implicit
> BindLink/LambdaLink:" so -- why did you change your mind? Why is it better
> that satisfyingset use implicit, instead of explicit variables? I don't get
> it.
>
> —
> <https://github.com/opencog/opencog/pull/1442#issuecomment-84181253>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

ngeiswei:
Linas, I didn't know you got that from a a cut-n-paste email from Ben (I'm surprised actually).

I am not contradicting myself but I misunderstood your definition of BindLink. I thought you considered that BindLink is essentially a lambda, while in your pattern matcher context it is more than that. All I suggest is to have operator for all lambdas, a LambdaLink, no need to have SatifyingSetLink, BindLink (or whatever it may be called in the future) playing the additional role of a lambda. You know, the Unix way: "do one thing and do it well".

linas:
Hi Nil,

On Mon, Mar 23, 2015 at 6:54 AM, ngeiswei <notifications@github.com> wrote:

> Linas, I didn't know you got that from a a cut-n-paste email from Ben (I'm
> surprised actually).
>

Well, a prettied-up and re-iterpreted cut-n-pste. I read the email, I say
to myself "oh I get it now" and I copy to the wiki. Then I expand and
elucidate and flesh out ... and I may introduce errors during the expansion.

I am not contradicting myself but I misunderstood your definition of
> BindLink.
>
I contradict myself all the time. I'm error-prone.

> I thought you considered that BindLink is essentially a lambda,
>
Yes, I probably said this or implied this.  When it was originally
conceived, I was thinking of lambda.  But then things changed...

> while in your pattern matcher context it is more than that. All I suggest
> is to have operator for all lambdas, a LambdaLink,
>
Yes.

> no need to have SatifyingSetLink, BindLink (or whatever it may be called
> in the future) playing the additional role of a lambda. You know, the Unix
> way: "do one thing and do it well".
>

I don't know what you mean by  "additional role".  Think of BindLink as a
"lambda that has been given a name and that name is BindLink"  If we have a
DefineLink (which we don't) then one might have perhaps this:

EquivalenceLink
    BindLink
         stuff
    DefineLink
         NameNode "pattern"
         LambdaLink
                stuff

For SatisfyingSet, we could have the following semantics:

    SatisfyingSetLink
          atom
          more_atoms

if atom is a VariableNode, then that variable is bound.  If atom is a list
of variable nodes, then those variables are bound.  If atom is neither a
variable, nor a list of variables, then all variables in the set are free.
   I was hoping that might work, but possibly there would be some ambiguity
or clashing... !?

-- Linas

PR #1443, opened 2015-03-20
Added support for CORS to Rest API

inflector:
This should fix the problem George is having.

### DIFF
--- a/opencog/python/requirements.txt
+++ b/opencog/python/requirements.txt
@@ -11,6 +11,7 @@ wsgiref
 Flask
 flask-restful
 flask-restful-swagger
+flask-cors
 mock
 six
 graphviz
--- a/opencog/python/web/api/apimain.py
+++ b/opencog/python/web/api/apimain.py
@@ -2,6 +2,7 @@ __author__ = 'Cosmo Harrigan'
 
 from flask import Flask
 from flask.ext.restful import Api
+from flask.ext.cors import CORS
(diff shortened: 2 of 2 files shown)


AmeBel:
I am not getting restapi.Start in opencog shell. PYTHONPATH is set as described in http://wiki.opencog.org/w/Python. What am i missing?

cosmoharrigan:
@AmeBel Does the REST API unit test pass on your machine? Do you have all the Python dependencies that it utilizes?

AmeBel:
The Unit test doesn't run, i get.

```
opencog@193046552ef0:~/tests/python$ python test_restapi.py
Traceback (most recent call last):
  File "test_restapi.py", line 13, in <module>
    raise unittest.SkipTest("ImportError exception: make sure the required "
unittest.case.SkipTest: ImportError exception: make sure the required dependencies are installed.
```
are there any other dependencies other than those listed in https://github.com/opencog/opencog/blob/master/opencog/python/requirements.txt ?

cosmoharrigan:
If you comment out 'try' and the entire 'except' clause, and just leave [this line](https://github.com/opencog/opencog/blob/6fad61aca3fb7af0ccfb01bf2ca98b69b5ec6699/tests/python/test_restapi.py#L10), then it should tell you what's missing.

AmeBel:
ok it is a new dependency pip install -U flask-cors

PR #1444, opened 2015-03-20
Disjuncts in order

rohit12:
I corrected the mistake that I had done previously. The disjuncts now occur in the same order as they occur in the sentence. So I solved that problem and I have also tested it on a couple of sentences and it works well.

### DIFF
--- a/opencog/nlp/learn/make-disjuncts.scm
+++ b/opencog/nlp/learn/make-disjuncts.scm
@@ -97,20 +97,7 @@
 			#t))
 	(map create-evaluation-link (filter criteria? (parse text)) ))
 
-;The extract-disjunct function returns a list of disjuncts for that particular 
-;word. However, that list contains many null lists also. These lists need to 
-;be eliminated before they can be processed further.
-(define (filter-disjunct disjuncts)
-	;This is the criteria for the filter function which cleans up the disjuncts 
-	;list for a list which does not contain any null lists.
-	(define (filter-disjunct-criteria? x)
-		(not (null? x)))
-	(filter filter-disjunct-criteria? disjuncts))
-
-;This function takes as input a word and a list of MST nodes created by the 
-;create-MST-nodes function. For the given word, it gives the disjuncts of that 
-;word as found in the list of nodes give to it as input.
-(define (extract-disjunct word nodes)
+(define (get-related-words-disjuncts word nodes)
 	;This function will return the first word present in the ListLink component of 
 	;an EvaluationLink.
 	(define (get-first-evaluation-word outset)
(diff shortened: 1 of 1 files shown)


linas:
sorry for delay I will try to look at this "real soon now".

opencog/nlp/learn/make-disjuncts.scm
```diff
@@ -134,14 +121,29 @@
...
 					'()
-					(cons (string-append linkname "-") (extract-disjunct word (cdr nodes))))
+					(cons (cons firstword (string-append linkname "-")) (get-related-words-disjuncts word (cdr nodes))))
 				(if (null? nodes)
 					'()
-					(cons '() (extract-disjunct word (cdr nodes)))))))
+					(get-related-words-disjuncts word (cdr nodes))))))
+		'()))
+
+(define (arrange-disjuncts-in-order listofdisjuncts sentence)
```
linas: Why do you need to arrange disjuncts in order ?? I'm confused.  

Also: You seem to have two concepts in your code, one of which you call a disjunct, but it isn't really.   A disjunct is this: (MB+ MC-)  but your are using "word-connector-lists"  ((the . MB+) (played. MC-))  It would be less confusing for me if you called this second thing "WCL" in the code, and explained that a wcl looks like ...  ((the . MB+) (played. MC-))

opencog/nlp/learn/make-disjuncts.scm
```diff
@@ -134,14 +121,29 @@
...
+	(if (not (null? sentence))
+		(let ([firstw (car sentence)])
+			(cons (get-disjunct-from-list listofdisjuncts firstw) (arrange-disjuncts-in-order listofdisjuncts (cdr sentence))))
+		'()))
+
+(define (get-disjunct-from-list listofdisjuncts firstw)
```
linas: Please add documentation.  I'm totally confused.  MC- is called a "connector"  and (played. MC-) is a word-connector pair. So I'm not sure, but this functions seems to return a connector for the word "firstw" ?  except that firstw doesn't have to be first, it just has to be any word in the sentence?  so this would be called "get-connector-for-word"

You've written the code in a generic fashion, so really, it should be called "get-item-for-word" where the input is a list of item-word pairs, and firstw is just a word.  item could be anything -- connectors, disjuncts, whatever ...

linas:
I've merged the code. It would be great if you added more documentation that explains what each function expects as input, and what it generates as output.  Also, in a separate email, can you send me examples of actual sentences, and the output that is generated for each?

rohit12:
I will write documentation soon. I will do it by Friday. I'll email you a set of around 5-10 sentences along with their output on Friday itself.

PR #1449, opened 2015-03-24
RuleEngine: Rule class update

williampma:
@misgeatgit 

Just added 2 new methods that I think will be needed for the BackwardChainer.

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/Rule.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/Rule.cc
@@ -59,8 +59,23 @@ string Rule::get_name()
 	return name_;
 }
 
-void Rule::set_handle(Handle h) throw (exception)
+void Rule::set_handle(Handle h) throw (InvalidParamException)
 {
+	// rules must be defined in a BindLink for pattern matching
+	if (h->getType() != BIND_LINK)
+		throw InvalidParamException(TRACE_INFO, "Expected BindLink for rule.");
+
+	HandleSeq outgoing = LinkCast(h)->getOutgoingSet();
+
+	// check for the BindLink's implication
+	if (outgoing.size() != 2 || outgoing[1]->getType() != IMPLICATION_LINK)
+		throw InvalidParamException(TRACE_INFO, "Rule's BindLink missing ImplicationLink.");
+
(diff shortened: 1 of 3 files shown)


opencog/reasoning/RuleEngine/rule-engine-src/Rule.cc
```diff
@@ -23,50 +23,116 @@
...
+
+/**
+ * Get the implicant (input) of the rule defined in a BindLink.
+ *
+ * @return the Handle of the implicant
+ */
+Handle Rule::get_implicant()
+{
+	// rules must be defined in a BindLink for pattern matching
+	if (rule_handle_->getType() != BIND_LINK)
```
misgeatgit: Wouldn't it be nice to throw Exception here?

opencog/reasoning/RuleEngine/rule-engine-src/Rule.cc
```diff
@@ -23,50 +23,116 @@
...
+ * Get the implicand (output) of the rule defined in a BindLink.
+ *
+ * XXX TODO Might want to do extra processing find the real output over an
+ *     ExecutionOutputLink.  ie, skip to the ListLink under the ExLink.
+ *
+ * @return the Handle of the implicand
+ */
+Handle Rule::get_implicand()
+{
+	// rules must be defined in a BindLink for pattern matching
```
misgeatgit: And also here.

opencog/reasoning/RuleEngine/rule-engine-src/Rule.cc
williampma: I don't know... I actually think all these checking should have been moved to set_handle to stop setting a rule with incorrectly structured BindLink.  If an exception has to be thrown, it makes more sense in this case to throw at the set function rather than the get function.

misgeatgit: Yes, you are right.

williampma:
@misgeatgit   Modified.

misgeatgit:
Thanks @williampma .

linas:
FYI, there is now a "cog-validate-bindlink" function which tells you if a bindlink has the proper format; it will tell you this without actually running it.

PR #1453, opened 2015-03-25
Reasoning: More Rule class update

williampma:
I am doing more mini push to avoid conflict.

- Switched to validate_bindlink (I noticed that function has an unneeded AtomSpace argument...)
- Added another functionality I am using, which is to make a Rule generate a standardize apart version of itself.
- Removed the disjunct_rules_ vector of pointers to make Rule object copy safe (it is better to have another class handle that).

I noticed two Handle to the same NodePtr cannot be compared if they are not added to an AtomSpace yet (because Handle == only compare uuid, which is not set yet), so ended up having to bypass validate_bindlink for the standardize apart version.

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/JsonicControlPolicyParamLoader.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/JsonicControlPolicyParamLoader.cc
@@ -249,17 +249,7 @@ template<typename> void JsonicControlPolicyParamLoader::read_primitive(
  */
 void JsonicControlPolicyParamLoader::set_disjunct_rules(void)
 {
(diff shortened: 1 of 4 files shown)


misgeatgit:
I think you should add your name in the Author section.

williampma:
What ordering are you referring to?

misgeatgit:
I was referring to the ordering of public and private declarations in b7433391ec507776b6b7

williampma:
Oh, just personal preference.  I am reading the header files like the "table of content" of a book, to allow me to quickly see what a class has available.  So I usually put public stuff at the top, since usually if I am looking at the header file, I do not care about the private stuff.

It is also why I am putting the doxygen stuff in the source rather than the header.

misgeatgit:
Good to know.

PR #1454, opened 2015-03-25
Misc pln improvements

ngeiswei:
Misgana, review that pull request and read carefully the commit messages. This is far from being exhaustive but it will give you some examples of the kind of simplifications you can do. Then merge it (unless you find something wrong of course).

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
@@ -51,9 +51,8 @@ Handle PLNCommons::create_quoted(Handle himplicant)
     return hquoted;
 }
 
-Handle PLNCommons::create_bindLink(
-        Handle himplicant, bool vnode_is_typedv)
-                throw (opencog::InvalidParamException)
+Handle PLNCommons::create_bindLink(Handle himplicant, bool vnode_is_typedv)
+    throw (opencog::InvalidParamException)
 {
     if (!LinkCast(himplicant)) {
         throw InvalidParamException(TRACE_INFO, "Input must be a link type ");
@@ -62,8 +61,7 @@ Handle PLNCommons::create_bindLink(
     //if(vnode_is_typedv)
     himplicant = replace_nodes_with_varnode(himplicant);
 
-    HandleSeq variable_nodes = get_nodes(himplicant, vector<Type> {
-            VARIABLE_NODE });
+    UnorderedHandleSet variable_nodes = get_nodes(himplicant, {VARIABLE_NODE});
     HandleSeq list_link_elem;
 
     //for searching implicationLinks with variables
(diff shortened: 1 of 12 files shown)


misgeatgit:
Thanks @ngeiswei. The commits are really good guidelines.

opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
```diff
@@ -51,9 +51,8 @@ Handle PLNCommons::create_quoted(Handle himplicant)
     return hquoted;
 }
 
-Handle PLNCommons::create_bindLink(
-        Handle himplicant, bool vnode_is_typedv)
-                throw (opencog::InvalidParamException)
+Handle PLNCommons::create_bindLink(Handle himplicant, bool vnode_is_typedv)
+    throw (opencog::InvalidParamException)
```
linas: FYI,  re throw, I discovered something weird I don't understand:  if I declare a throw, explicitly, like this, then neither gdb nor the stack trace will indicate the correct line number in the C file;  instead, it will use the line number of the very last line of the method.  If I don't declare the throw, then the line numbers work.

There is also a different and much more serious problem: if this method calls another method that throws the same exception, then the exception handling gets completely messed up.  The only way to make the exception work correctly is to remove the  throw declaration. 

I mention this because I've reworked scheme so that scheme exceptions are turned  into c++ exceptions, and c++ exceptinos are turned into scheme exceptions, so that you can throw anywhere for any reason, and can always catch, in either language, and everything "just works".

opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
```diff
@@ -84,39 +83,34 @@ Handle PLNCommons::create_bindLink(
     return as_->addLink(BIND_LINK, var_listLink, implicationLink);
 }
 
-HandleSeq PLNCommons::get_nodes(const Handle& hinput,
-                                vector<Type> required_nodes)
+UnorderedHandleSet PLNCommons::get_nodes(const Handle& hinput,
```
linas: FYI, there is a perhaps simpler, easier-to-use version of this method in the FindVars class in /opencog/query/PatternUtils.h  -- there are a bunch of this kind of utilities there.

opencog/reasoning/RuleEngine/rule-engine-src/pln/PLNCommons.cc
ngeiswei: Yes, specifying the throw at the function declaration is probably useless at best, see

http://www.drdobbs.com/sutters-mill-a-pragmatic-look-at-except/184401544

linas: ah ha!

PR #1466, opened 2015-03-26
Definition of OCTIME_MAX

aurelien-rainone:
Define and use OCTIME_MAX
    
i have defined OCTIME_MAX (max octime_t/timstamp value) to replace
hardcoded integer values in some tests or checks
    
 - fixed TemporalUTest failed test with the use of OCTIME_MAX
 - replaced some unsigned int/long timestamps with octime_t
 - fixed 32bit compiler issues
 - see #1448


williampma:
Just wondering, is this PR generating the warning now appearing on buildbot

http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/396/steps/compile/logs/warnings%20%288%29

aurelien-rainone:
ha, this...yes!
I'll issue a new PR for their suppression

aurelien-rainone:
%ul (like it was before) were issuing warnings on my x86 machine because octime_t corresponds an unsigned long long, so i changed for %ull to remove the warnings

But on the buildbot, and probably on most of the machines now, it's the opposite.

Unfortunately the only way to suppress these warnings on both platforms would be to use an horrible macro defined in <inttype.h>, and do this : 

<pre>
printf("a is %#" PRIx64, (uint64_t) 123465)
</pre>

or of course a better solution would be
<pre>
std::ostringstream
</pre>

what do you think is best?

aurelien-rainone:
i wrote a bit too fast.
The TemporalUTest fails on a 32 bits compiler if we don't use the %llu format placeholder

inflector:
You could keep the %llu and cast the octime_t variables to a (unsigned long
long) in the parameter list for the format. That should work in both 32-bit
and 64-bit machines and will definitely match the format specifier. Casting
to what might be a larger integer type is a guaranteed lossless cast.

aurelien-rainone:
good idea, yes that does the trick

aurelien-rainone:
i issued these 2 PR, both implements timestamp parsing in Temporal :

 * boost:format #1469
 * c++ standard streams #1470

both get rid of the compile [warnings](http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/396/steps/compile/logs/warnings%20%288%29) 
and with both, TemporalUTest passes
If these modifications are seen are too much performance-consuming, i'll issue a PR using an (unsigned long long) cast, as proposed by @inflector

PR #1470, opened 2015-03-27
use std streams for Temporal parsing

aurelien-rainone:
following the discussion in #1466 this PR is the solution using c++ streams for temporal parsing

### DIFF
--- a/opencog/spacetime/Temporal.cc
+++ b/opencog/spacetime/Temporal.cc
@@ -25,6 +25,8 @@
 #include "Temporal.h"
 
 #include <iostream>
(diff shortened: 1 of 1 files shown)


linas:
this cannot be merged it has a merge conflict

aurelien-rainone:
that's ok, one only of #1470 or #1469 had to be merged, not both, they are
exclusif OR.
they were 2 different ways to do the same thing
you merged the one using boost::format that's ok

2015-03-27 18:56 GMT+01:00 Linas Vepštas <notifications@github.com>:

PR #1471, opened 2015-03-27
Remove obsolete code, retire old code

linas:
A general cleanup of old deprecated API's in the atomspace.  Remove as much as possible without hurting current function.


opencog/atomspace/AtomSpace.h
```diff
@@ -97,19 +97,6 @@ class AtomSpace
...
-     * @param type  the type of atoms that should be printed.
-     * @param subclass  if true, matches all atoms whose type is
-     *              subclass of the given type. If false, matches
-     *              only atoms of the exact type.
-     */
-    void print(std::ostream& output = std::cout,
-               Type type = ATOM, bool subclass = true) const {
-        getAtomTable().print(output, type, subclass);
-    }
-
```
ngeiswei: Why remove print? It's a very convenient function for debugging utests.

linas: you're going to print a million atoms while debugging?  Does not seem likely... that was my reasoning.

linas: You can get the equivalent function by doing a 'getAtomsByType() if you really want to, and then you can print uuids, pointers, whatever.   I would rather have the atomspace  contain only what is needed to get high performance. If performance doesn't matter, then a 1 line long utility for (Handle h: as->getAtomsByType()) cout << h->toString();  is enough.

ngeiswei: Regarding printing, I usually use that in unit tests, where the AtomSpace is reasonably small and I can't necessairy log into a cogserver. But I can go with that for loop.

PR #1475, opened 2015-03-29
Added documentation

rohit12:
I added documentation for all those functions for which my changes were unclear. I have also renamed my variables a bit in line with your suggestion.


linas:
OK, thanks!

PR #1477, opened 2015-03-30
Make a cleaner separation of the pattern match code.

linas:
This was a "conceptually easy" movement of the pattern validation code into it's own directory. 

The new code will validate a bindlink once and only once: when it is first inserted into the atomspace. After that, no further validation is needed.  Bad patterns cannot be inserted into the atomspace.

I thought it would only require moving a few files, and take an hour .. instead, it took two days, and boiled the ocean. :-(


opencog/atomutils/AtomUtils.cc
```diff
@@ -107,26 +102,4 @@ HandleSeq getNeighbors(const Handle& h, bool fanin,
...
-        for (int i = 0; i < len; ++i) {
-            name += alphanum[rand() % (sizeof(alphanum) - 1)];
-        }
-        result = as.getHandle(t, name);
-    } while (as.isValidHandle(result));
-
-    return as.addNode(t, name);
-}
-
 }
```
ngeiswei: I don't understand what should pertain to AtomUtils.cc and what should pertain to AtomSpaceUtils.cc

ngeiswei: I get it! AtomSpaceUtils requires a reference pointing to an AtomSpace, AtomUtils doesn't! :-)

I guess I'll that in their respective readmes.

linas: > I get it! AtomSpaceUtils requires a reference pointing to an AtomSpace,
> AtomUtils doesn't! :-)
>
> I guess I'll that in their respective readmes.
>

Yeah, I hit a circular-dependency when linking, and had to split them up.

opencog/atoms/bind/BindLink.cc
```diff
@@ -0,0 +1,107 @@
...
+	{
+		const std::string& tname = classserver().getTypeName(t);
+		throw InvalidParamException(TRACE_INFO,
+			"Expecting a BindLink, got %s", tname.c_str());
+	}
+
+	// The LambdaLink contructor sets up _body and _varset
+	validate_body(_body);
+	unbundle_clauses(_hclauses);
+	validate_clauses(_varset, _clauses);
```
williampma: Hello Linas,

The early checking of BindLink does not allow me to create the complete BindLink outside the AtomSpace, and then adding the whole thing all at once to the AtomSpace, since the validation code are comparing Handles, which are UNDEFINED until everything is in the AtomSpace.

I could work around this in my code by adding things to AtomSpace immediately as I build.   Just thought you should know.

PR #1478, opened 2015-03-30
AtomTable: Added a test triggering a bug

williampma:
Bug report and pull request at the same time.

**The bug:**
  1. Create a node and add it to a link twice, then add the link to the atomtable.
  2. The resulting link contains two handles that are not equal (even though doing ->toShortString() print the same UUID).

**What else I noticed:**

Changing "ConceptNode" to "NumberNode" magically makes this bug go away.  Whatever the AtomTable is doing for NumberNode is more correct than for other nodes.


linas:
trying to fix this now.

PR #1480, opened 2015-03-30
PLN  module for interacting with  ECAN.

misgeatgit:
The following feature enables the rule engine to function like an event based rule engine and at the same time interact with ECAN. Besides the basic atom added  and attentional focus events, other kinds of events can be added later when the need arises.


misgeatgit:
Since this an initial step, I will merge it.

PR #1481, opened 2015-03-30
BackwardChainer: code rearrange

williampma:
This PR does nothing but just rearranged some stuff.  My correct work is based on this, so I am pushing this out of the way.

### DIFF
--- a/opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
+++ b/opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
@@ -28,8 +28,8 @@
 BackwardChainer::BackwardChainer(AtomSpace * as) :
 		 as_(as)
 {
-	commons_ = new PLNCommons(as_);
-	bcpm_ = new BCPatternMatch(as_);
+	_commons = new PLNCommons(_as);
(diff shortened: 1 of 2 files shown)


opencog/reasoning/RuleEngine/rule-engine-src/pln/BackwardChainer.cc
```diff
@@ -22,67 +22,338 @@
...
-BackwardChainer::BackwardChainer(AtomSpace * as) :
-		 as_(as)
+BackwardChainer::BackwardChainer(AtomSpace * as)
+    : _as(as)
 {
-	commons_ = new PLNCommons(as_);
-	bcpm_ = new BCPatternMatch(as_);
+	_commons = new PLNCommons(_as);
+	_bcpm = new BCPatternMatch(_as);
 }
```
ngeiswei: BTW, a good practice is to avoid using new/delete as much as possible. No new/delete means less code and less possibilities of memory leaks. Sometimes you'd have to, but it seems not the case in this instance.

williampma: yeah, it is already removed on my local copy.

PR #1486, opened 2015-03-31
Fix for bug #1478

linas:


williampma:
Thanks, Linas.

linas:
yah, my fix seems to break EinsteinUTest!!? which I did  not discover until after merging it. I'm unlike to nail it today, though.

linas:
Never mind. I just fixed that...

linas:
yeah, the atom-table add code s kind of hairy and hard to follow... :-(

PR #1492, opened 2015-04-01
Fix Atomspace Publisher client bug

sanuj:
@cosmoharrigan JSON was not getting parsed properly in the client. The key was wrong.


sanuj:
No need to touch gevent-socketio I suppose :)

cosmoharrigan:
Thanks for identifying this, @sanuj! Indeed, the further discussion around gevent-socketio was a false alarm for now.

PR #1495, opened 2015-04-01
Fixed atomspace_bm and added optimized getHandlesOfType (37% faster) 

inflector:
Consolidated the -A and moved it to AtomSpaceBenchmark.cc so that it runs valid tests
Fixed problems with benchmark generating random types that break with new validation
Ported over the old changes to add a getNumAtomsOfType as per pull Linas's request in #1409
Optimized getHandlesOfType to use vector.reserve to avoid copies and memory management.


linas:
sorry to take so long my back is out, I can't get to email

inflector:
no worries

sorry to hear about your back, back pain can be so debilitating, hope you
get better soon

PR #1499, opened 2015-04-03
Infrastructure work of pattern matcher and related tech.

linas:
BTW, I am out with a back injury and might not respond to emails for a while.


AmeBel:
Get well soon now

PR #1500, opened 2015-04-03
Added cog-get-pred equilvalent: get_predicates, and new get_predicates_for to C++ and Python

inflector:
This also includes the Cython Python bindings and a Python binding test for get_predicates and get_predicates_for as well as the yield generator versions: xget_predicates and xget_predicates_for.


inflector:
This also includes the Cython Python bindings and a Python binding test for get_predicates and get_predicates_for as well as the yield generator versions: xget_predicates and xget_predicates_for.

PR #1504, opened 2015-04-07
removed some x86 specific compile warnings

aurelien-rainone:
i just add some casts to unsigned long in order to suppress some 32 bits architecure specific warnings.
Those took place on debug printing calls, so i didn't want to add a dependency to boost::format


linas:
I don't want to merge this, because the warnings were coming from temp debug statements that were accidentally left on in a previous batch of commits.  Normally, these debug statements are never compiled.

PR #1510, opened 2015-04-09
Added execute_atom and evaluate_atom

inflector:
Lots of fun with circular CMake dependencies kept me from being able to add these functions as methods on the atomspace object.


linas:
looks nice!

PR #1515, opened 2015-04-10
Deductionrule

AmeBel:
Partial Implementation of the [Independence Based Deduction Formula](http://wiki.opencog.org/w/Independence_Based_Deduction_Formula) as per https://github.com/opencog/opencog/issues/1401

@ngeiswei @moikle please review.

One issue, how to handle the consistency condition when sB=0 ?


AmeBel:
@ngeiswei how should sNotB appraoching zero fit in within the consistency condition (as per your recent changes https://github.com/opencog/opencog/blob/master/opencog/reasoning/RuleEngine/rules/pln/deduction.scm#L67) ?

moikle:
What is the difficulty when sB = 0? If you mean when sB = 1 then use sAC = sC if sC = sBC or 0 otherwise. 

If you mean how to set min(1, sC/sB) when sB = 0, set it to 1.

—matt

AmeBel:
@moikle from your above comment the first consistence condition will be.
```
IF: max((sA+sB-1)/sA, 0) <= sAB<= min(1, sB/sA) AND max((sB+sC-1)/sB, 0) <= sBC <= [IF: sB=0, THEN 1, ELSE min(1, sC/sB)], 
```
similarly for the case of  sA = 0, the first consistency condition should be edited to
```
IF: max((sA+sB-1)/sA, 0) <= sAB<= [IF: sA=0, THEN 1, ELSE min(1, sB/sA)] AND max((sB+sC-1)/sB, 0) <= sBC <= [IF: sB=0, THEN 1, ELSE min(1, sC/sB)], 
```
right?

Second how about for max((sB+sC-1)/sB, 0) when sB=0? Should it be set to 0?

moikle:
Yes

moikle:
I just went through the merge and all looks OK to me.

—matt

AmeBel:
Thanks, Let me know when the meaning of 'min_count' in the nAC forumla is discovered.

moikle:
Yeah — Ben and I went through that count formula and it all seems OK, but we did not discuss what min_count represents — the heuristic formula is a hack but is sensible — I’m thinking that min_count should just be some parameter setting indicating the smallest count value for Atom B — say 1 or 5 or …. — not sure why Ari put it in there, but there probably was some reason he encountered such as division by zero :-)

—matt

AmeBel:
min_count = 1 makes sense for avoiding division of zero.

moikle:
Why don’t we use that for now?

—matt

AmeBel:
sure, i will do that

PR #1516, opened 2015-04-10
Added support for scheme-style python type construction

inflector:
Yes, you can now add types like:

```
result = execute_atom(atomspace, 
        ExecutionOutputLink( 
            GroundedSchemaNode("py: add_link"),
            ListLink(
                ConceptNode("one"),
                ConceptNode("two") 
            )
        )
    )
list_link = ListLink(
        ConceptNode("one"),
        ConceptNode("two")
    )
assert_equals(result, list_link)
```

just by adding:
```
from opencog.type_constructors import *
```
and:
```
initialize_opencog(atomspace)
```


cosmoharrigan:
Very interesting idea!

Does this syntax allow the TruthValue to be included at the time of atom definition?

opencog/cython/opencog/type_constructors.pyx
```diff
@@ -0,0 +1,132 @@
+
+# DO NOT EDIT THIS FILE! This file was automatically generated from atom
+# definitions in types.script by the macro OPENCOG_ADD_ATOM_TYPES
```
cosmoharrigan: If I understand correctly, this .pyx file is dynamically generated by the cmake file when the project is built? And if so, why does it need to be committed? Or maybe I misunderstood?

inflector: It doesn't really, but then it won't show up in the source on github and it will be far more mysterious where these types come from.

I contemplated adding it to .gitignore but decided against it.

Not a big deal but having a copy on github so people can search and find it via google, seemed like a good thing.

cosmoharrigan: Okay, that makes sense. Thanks. I can see how that could be helpful.

inflector: On adding the TruthValue, you can call any Atom functions on each of the parts as returned by each function. So:

```
ConceptNode("dog").tv(TruthValue(0.5,0.5))
```
should work fine, but I haven't yet tested it. If it doesn't, I'll fix it up so it works well.

I can add more convenience functions if there are common ways of initializing TruthValue objects. So perhaps something like:

```
ConceptNode("dog").truth_value(0.5,0.5)
```

cosmoharrigan: In the Scheme syntax, the atom is defined as a list that may contain a TruthValue object, for example,
```(ConceptNode "Anna" (stv 0.1667 1))``` as in
https://github.com/opencog/opencog/blob/master/opencog/python/pln/examples/tuffy/smokes/smokes.scm#L17
but perhaps something like the syntax you have illustrated makes more sense for Python. Or maybe ```ConceptNode("dog").tv = SimplePLNTruthValue(0.5, 0.5)```. But what if you want to include the truth value at the same time as you are creating the atom?

cosmoharrigan:
Thanks @inflector.

cosmoharrigan:
@inflector the build ended with an error ```Error compiling Cython file``` - please see http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/444/steps/compile/logs/stdio

AmeBel:
@inflector  thanks :-)

AmeBel:
It would be great if the examples are documented in the wiki, http://wiki.opencog.org/w/Python including the usage of python functions by the GroundedPredicateNode/GroundedSchemaNode. That way it would be easier to point the wiki for opencog-noobs. Just a suggestion.

inflector:
Yes, once I figure out the build problem on buildbot, I'm going to do some
documentation for the new work I've done and also do some tutorial-level
getting started with OpenCog in Python examples.

williampma:
Is there any reason PYTHON_FILE is being put into the source directory rather than the build directory?  This is creating untracked files that git is complaining about.

inflector:
William,

I'm working on a fix for this. There is a reason that the one file needs to
be there so that one can access the types via the sameL

from opencog.type_constructors import ConceptNode

in order to have it in the opencog package in Python the main Cython file
needs to be there unless I do something special. I've changed this so that
that the Cython file there type_constructors.pyx is no longer
auto-generated, but now includes the files in build that are
auto-generated. I didn't really understand the type system well enough when
I started and made a bad assumption that I've since implemented a fix for
and should have submitted a pull request for shortly. I still have a weird
bug I"ve been tracking down as I'm not an expert on CMAKE macros yet, but
those generated files will be placed into build when I'm done.

- Curtis

inflector:
@AmeBel I updated the Python page to reflect the new additions. It's not everything I want yet, but it includes the suggested new way and also shows the example of `execute_atom` in an example contrasting using `scheme_eval`.

inflector:
@williampma The auto-generated files are now placed into build and the main cython file just includes the build-directory files. That way we get the Cython shared libraries and source in the cython directory so Python can find it.

AmeBel:
Thanks @inflector

Issue #1506, opened 2015-04-08
Stimulating atoms outside of a mind agent


misgeatgit:
Currently atoms can only be stimulated with in an Agent class instance. Is there any particular reason for this design choice? What is the disadvantage of making stimulus attribute of an Atom class and stimulation a method defined by the AtomSpace class?

bgoertzel:
On Wed, Apr 8, 2015 at 1:38 PM, Misgana Bayetta <notifications@github.com>
wrote:

> Currently atoms can only be stimulated with in an Agent class instance. Is
> there any particular reason for this design choice?
>

There's no good reason for that, we just weren't thinking of a
non-Scheduler-based architecture at the time that was done ...


> What is the disadvantage of making stimulus attribute of an Atom class and
> stimulation a method defined by the AtomSpace class?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencog/opencog/issues/1506>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

"The reasonable man adapts himself to the world: the unreasonable one
persists in trying to adapt the world to himself. Therefore all progress
depends on the unreasonable man." -- George Bernard Shaw

PR #1521, opened 2015-04-11
PLN Attentional Focus fixes

linas:
I suspect that the default attentional-focus boundary is interfering with PLN development, and is making things confusing, by "hiding" some atoms from the forward chainer, etc.

The PLN pattern-matcher uses the attentional-focus filter. The default attentonal-focus boundary is **above** the default STI of atoms in the atomspace.  The net result seems to hide some/many/most atoms from PLN, butcause the attntional-focus agents are not currently running.  So, for example,  the ForwardChainerUTest was failing, when the AF filter was fixed to work correctly. 

During PLN development and testing, you almost surely want to call 
```
atomspace->setAttentionalFocusBoundary(AttentionValue::MINSTI);
```
as otherwise, you might not see all the atoms you are expecting to see ...

### DIFF
--- a/opencog/query/AttentionalFocusCB.cc
+++ b/opencog/query/AttentionalFocusCB.cc
@@ -36,7 +36,7 @@ AttentionalFocusCB::AttentionalFocusCB(AtomSpace* as) :
 	DefaultPatternMatchCB(as)
 {
 	// Temporarily disable the AF mechanism during PLN development
-	_as->setAttentionalFocusBoundary(AttentionValue::MINSTI);
+	// _as->setAttentionalFocusBoundary(AttentionValue::MINSTI);
 }
 
 bool AttentionalFocusCB::node_match(const Handle& node1, const Handle& node2)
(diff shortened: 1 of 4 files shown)


ngeiswei:
Linas, isn't "hiding" atoms from the forward chainer exactly the point of the attentional focus?

On a branch (not committed anywhere yet) I work around what you describe by boosting the STI of atoms I want to keep in the attentional focus (such as premises), it works pretty well and definitely speeds up reasoning.

bgoertzel:
Well, having PLN look *only* at Atoms in the AF is one useful strategy, but
not the only strategy we want to follow in the  medium term...

This basically simulates a "purely conscious thought process", whereas
running PLN more broadly across the whole Atomspace simulates an
"unconscious reasoning process" ...


On Mon, Apr 13, 2015 at 4:39 PM, ngeiswei <notifications@github.com> wrote:

> Linas, isn't "hiding" atoms from the forward chainer exactly the point of
> the attentional focus?
>
> On a branch (not committed anywhere yet) I work around what you describe
> by boosting the STI of atoms I want to keep in the attentional focus (such
> as premises), it works pretty well and definitely speeds up reasoning.
>
> —
> <https://github.com/opencog/opencog/pull/1521#issuecomment-92272486>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

misgeatgit:
@ngeiswei did you directly boost the STI or gave a stimulus? I was looking for a way to stimulate atoms from a module but in the current design stimulus can only be given by mind agents and then  later converted to STI by the ImportanceUpdatingAgent. There is an issue opened on this https://github.com/opencog/opencog/issues/1506.

ngeiswei:
@misgeatgit I'm boosting the STI directly via calling h->setSTI. I'm really hacking here, that is why I didn't commit anything, I do have a complete understanding of how attention allocation is supposed to work.

Issue #1518, opened 2015-04-10
Build error triggered by #1516


cosmoharrigan:
@inflector this issue is to track the build error triggered by #1516 as noted here: https://github.com/opencog/opencog/pull/1516#issuecomment-91679842

inflector:
Hey Cosmo, I won't be able to get to this for a while. Can you easily rollback the changes. I think I see what I missed. Doesn't show up for make only for make -j >1

linas:
doesn't show up for make -janything for me ...

cosmoharrigan:
It builds successfully for me on my machine as well. Hence, I'm not sure that it's necessary to revert the change immediately. Rather, when you are back perhaps you can check what might be the cause of it not building on Buildbot. Thanks.

linas:
It might be a race condition in the dependencies .. with -j6, it could
happen that the thing being used has not yet been built.  You'd need to
tweak some cmakefile somwhere, and add the dependency

inflector:
Just submitted a pull request, #1523 which should fix this problem.

The generated files are now going to the build directory and type_constructors.pyx just includes them.

cosmoharrigan:
Thanks. 'make' succeeds now on Buildbot, but 'make tests' fails. However, I can't tell whether that failure is related to this or not, can you? http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/451/steps/compile_2/logs/stdio

inflector:
It's possible but I didn't touch the scheme type generation code. At least not intentionally.

inflector:
That test builds on my machine just fine, as well.

inflector:
@AmeBel Do you have access to the buildbot machine?

inflector:
Actually, that isn't a test that's a distribution script. It doesn't build as a target for me. Perhaps because I am missing some of the cmake dependencies for embodiment?

Nevertheless, I have the file in the place it is expecting to find it, as that's the main core_types.scm file for atomspace.

So that's probably not an error I caused, but I'm still a bit suspicious since it is one of the generated files and I most definitely was in the code for auto-generation of types.

Hopefull, @AmeBel can get more information about the missing file.

AmeBel:
The core_types.scm is not being created. To reproduce, run in build directory
```
$ rm -rf *
$ cmake ..
$ make -j$(nproc)
$ make test
```

inflector:
[...]

PR #1523, opened 2015-04-13
Type constructor fixes and added truth_value function to Atom

inflector:
Fixed up the Python type constructor macros and mechanism so it uses separate generated files and includes them all from the `type_constructors.pyx` file. 

This should fix the build problem on buildbot: #1518

Also added the a truth_value function to Atom in Python. So you can now do something like:
```
InheritanceLink(
    ConceptNode("Spaceship"),
    ConceptNode("machine")
).truth_value(0.5,0.5)
```


cosmoharrigan:
Thanks. One thing to consider: What if one wanted to use a different truth value type than the simple truth value?

inflector:
@cosmoharrigan These are not supported in the Python interface at all at the moment. That's how things were all along. I don't know enough about the way the other types are used to fix it yet. 

It may be a bit tricky, but the way that the other truth values are used and differ is only a guess for me at the moment.

inflector:
See: https://github.com/opencog/opencog/blob/master/opencog/cython/opencog/truth_value.pyx#L9

cosmoharrigan:
Oh yeah, that's right.

PR #1529, opened 2015-04-15
Pattern Matcher: added abnormal untyped VariableNode handling

williampma:
Needed for weird stuff like

```
(SatisfactionLink
   (VariableNode "$A")
   (VariableNode "$A")
)
```

I am doing this in a temp small atomspace of course, for unification between two atoms.

Previously `(cog-satisfy ...)` of the above will always return `(stv 0 1)`.


williampma:
Merging since the Backward Chainer needs this to move forward (or should it be "backward"? ).

PR #1535, opened 2015-04-16
Finite State Machine

linas:
Demo of a simple finite state machine running entirely in the atomspace.  As per usual: each call to the pattern matcher cycles it thogh one state.

This is  surprisingly very easy to do -- you just write out all the state transitions, and that's it. The BindLink that runs through them all is essentially the same for all possible FSM's.  It's not complicated, either.


williampma:
Missing DeleteLink.cc!

linas:
Whoops!  pushed in commit 9a90d4b9961e3ae98de4814c7aef58145ef046b2

PR #1542, opened 2015-04-17
Backward Chainer: preliminary typed variables check

williampma:
Still stuck in the "Impossible situation"...


linas:
If you are hitting asserts in the pattern matcher, they are internal bugs and not user errors. Rather than working around them, they should be reported and treated as bugs ...

PR #1544, opened 2015-04-17
Backward Chainer: moving alone

williampma:
Backward chained a bit further.  Now stumble onto incorrect code for grounding a Rule with the selected premises.


williampma:
Actually, the comment is wrong.

It requires some sort of BindLink to forward chain the ImplicationLink....

PR #1547, opened 2015-04-17
Multiple demos; multi-atom-space bug fixes

linas:
A fairly serious bug was preventing the proper use of scheme-based GroundedPredicateNodes ... I guess no one is using these?  This bug is fixed here.

Also: add demos of:

How to use GroundedSchemaNode
How to use GroundedPredicateNode
Example finite state machine
How to use ContextLinks to store conditional probabilities
Markov chain demo, using above (incomplete/broken)


examples/pattern-matcher/README.md
```diff
@@ -0,0 +1,19 @@
...
+* simple.scm: a basic introduction to pattern matching.
+* gsn.scm: calling arbitrary functions upon a match.
+* gpn.scm: calling arbitrary functions to decide a match.
+
+Some simpler applications showing some things one can do, and how to do
+them.
+
+* fsm-basic.scm: A simple Finite State Machine demo.
+* fsm-full.scm: A generic finite-state-machine constructor
+* markov-chain.scm: A Markov chain (probabilistic FSM) based on above.
```
cosmoharrigan: It will be interesting to have this in the atomspace!

linas: I forgot to push DeleteLink.cc and the demos won't work without it. I just pushed; they should now work.

PR #1555, opened 2015-04-21
Add 'limit' parameter to REST API

eddiemonroe:
Add a new optional parameter, 'limit,' to the REST API, which can be used to specify the maximum number of atoms that should be returned to the client from a query. If query results are greater than the number specified, then the result set list is truncated to the first 'limit' number of atoms.


cosmoharrigan:
@eddiemonroe thanks; if you have a chance, please also consider adding a unit test for the new parameter to https://github.com/opencog/opencog/blob/master/tests/python/restapi/test_restapi.py

Issue #1091, opened 2014-11-06
Pattern Matcher: UnorderedLink matching failures.


linas:
Several unordered-link pattern matching issues remain.   They're hard to understand and fix...

In StackMoreUTest, test_uu is now stubbed out, because it hits an infinite loop. (It used to work fine)

In SudokuUTest, the test_2x2_any is stubed out, it finds one solution too many.  It used to work fine ...

The issue is that backtracking through UnorderedLinks, and making sure that each permutation is explored, and none are accidentally skipped, when there may be multiple such UnorderedLinks in multiple clauses, which may be nested ... is tricky.

In particular, there are state stack issues, when exploring upwards while clause checking (viz have to save/restore state each upward step), and then walking downwards to verify that the discovered subclauses actually match...

I ran out of time to fix this, ... need to fix later.

linas:
fixed by commit #1557

PR #1557, opened 2015-04-23
Fully working Unordered Link

linas:
This finally fixes issue #1091.

This merge concludes a major reorganization of the pattern matcher internals. Many state variables have been removed; this should make the flow of control much easier to understand. Many routines were renamed to give them better names. Lots of documentation was added; the code should be a lot more understandable than it used to be.  Numerous conceptual issues were resolved: the old code was spaghetti code, in part because I didn't clearly understand and delineate the recursive parts correctly.  Oh, BTW, the new code is considerably faster, at least, on some test cases.

All this also means that any future extensions (I still have a couple planned) should be a lot easier to integrate; but perhaps I'm done for now.


sigkill:
Wow thanks for the quick fix!

Issue #1539, opened 2015-04-16
Clean obsolete PLN code and documentation


ngeiswei:
The obsolete C++ and Python versions of PLN are still documented on the wiki

wiki.opencog.org/w/Category:PLN-Obsolete
http://wiki.opencog.org/w/New_PLN_Design,_2013
etc...

I don't think keeping everything and putting under the obsolete category is a good idea, we need to simplify as much as possible. I don't see any point keeping those pages unless they contain some really interesting material, and in such a case then should be moved to a better place.

The obsolete Python version of PLN is still in the repo, both code and documentation

opencog/reasoning/RuleEngine/README.md (that one is really confusing!!!!)
opencog/python/pln

under python/pln/examples there are still used and usable scheme code. Either I can let that in here and rename pln by pln_obsolete. Or move the scheme code under say

opencog/reasoning/scm

Let me know if you have any feedback before I start butchering all that.

AmeBel:
On the python PLN, i think it is best to keep it untill the new PLN is working cosistently and we have resolved the representation issues so as to port all the rules, examples and unit tests before butchering it.

AmeBel:
For me, it would be great if 
* the scheme-binding to URE/PLN works
* a python binding is added, for making mind agents
* the way to use the control policy is clearly documented. 
* the difference between URE and PLN is clearly documented. At present it is not clear how the PLN implementation is different/similar from URE,  at least for those of us who are not C++ gurus.

ngeiswei:
@AmeBel I agree on both points, so

1. I'll rename python/pln by python/pln_old for now

we may decide to remove it entirely once the port is complete

2. I'll strive toward making the difference between URE and PLN clear (it's not really clear to myself either BTW).

Thanks for the feedback.

williampma:
The "pln" folder under "rule-engine-src" really shouldn't be named so, as it's just rule engine code.  The "pln" folder under "reasoning" however, I have no idea what that is.  I think maybe it should have been "pln-type" or something.

In fact, it would be great to change "rule-engine-src" to something else.  It always bugged me there is a "src" folder when the whole opencog folder is already for source.

I am also not sure why "RuleEngine" is under "reasoning".  What would be "reasoning" without "RuleEngine"?  Is the whole "reasoning" already about the "RuleEngine"?  If so, "RuleEngine" stuff might need to move one folder up.

Maybe this structure like this would be better?

```
reasoning\pln-type
reasoning\rules
reasoning\engine
[...]

PR #1559, opened 2015-04-23
Sort pln documentations

ngeiswei:
This request contains a few PLN clean-ups, mostly

1. python/pln is renamed with python/pln_old
2. various READMEs pertaining to the new and old PLN are sorted so it's clear to the user what is old and what is new.


ngeiswei:
This partially addresses issue #1539

Issue #1530, opened 2015-04-15
Multiple atomspace getAtom bug


williampma:
The following code, assuming _as is just the normal atomspace everyone started with
```
AtomSpace* temp = new AtomSpace(_as);   // temp has _as parent

Handle h1 = temp->addAtom(createNode(CONCEPT_NODE, "abc"));

AtomSpace indie;
Handle h2 = indie.addAtom(h1);
Handle h3 = temp->getAtom(h2);

if (h3 == Handle::UNDEFINED)
    logger().debug("!!!!!!!!!!!!!!!!!!!FUBAR!!!!!!!!!!!!!!!!!!!!!");
```

FUBAR is printed, even though it should not!

linas:
Yep. OK. Took me a moment to understand what you are doing here. h3 should have been identical to h1.   I'll try to look at this "real soon now".

linas:
pull req #1561 fixes this for me.

linas:
please close if this is fixed.

williampma:
Ah right.  I will check it right away as soon as I figured out what happened to everything...  was gone for 2 weeks and everything changed...

PR #1561, opened 2015-04-23
Bugfix for multi-atomspace bug #1530

linas:
Bugfix for multi-atomspace bug #1530 and unit test expansion to look for it.


williampma:
It is fixed, thanks!

PR #1562, opened 2015-04-23
backing store testing update

AmeBel:
* Enable testing of backing store when postgres database is running on a remote server. 
* Testing while using docker containers are configured as defined in https://github.com/opencog/docker/blob/master/opencog/docker-compose.yml will be possible.
* This will be enabled on buildbot
* Documentation changes for consistency with examples and opencog-test.conf


linas:
Any reason why test-persist was renamed to opencog_test.

AmeBel:
The rename is to make it consistent with the examples @ https://github.com/opencog/opencog/blob/master/examples/guile/persist-example.scm

PR #1563, opened 2015-04-24
Reduct in the AtomSpace

linas:
Pursuant to the email discussion ... this gives Nil yet one more thig to think about: a cheap hack proof-of-concept for reduction in the atomspace.  Its not elegant code; nor is it fast. Rather it shows that its possible,  in a not-to-hard way.

So, for example:
```
(cog-reduce!
   (TimesLink
      (VariableNode "$y")
      (NumberNode 0.5)
      (VariableNode "$x")
      (PlusLink (NumberNode 1) (NumberNode 1))))
```
 reduces to `(TimesLink  (VariableNode "$y")  (VariableNode "$x"))`

We could every easily set this up so that  reduction runs automatically when an atom is added to the atomspace, thus making it impossible to insert non-reduced expressions.

I can't say that this is immediately useful for anything; its more ov a conversation-starter.


nilaidyia:
Interesting! So it performs a limited form of reduction (partial evaluation) but still. Of course I strongly advocate implementing something relying on algebraic properties rather than concrete types, cause it will make reduction of GSN and GPN way easier then!!! But since it's just a proof-of-concept.

I had the plan to implement it in Haskell once we have a binding to use Haskell pattern matching. Ultimately I don't mind using C++, as long as the code is elegant enough. I also had this idea of coding it in Felix http://felix-lang.org/ would be a way to use Felix pattern matching and have some very efficient C++ code out of it. But maybe what we should do is coding it Atomses! Using OpenCog pattern matcher, and have the Rule Engine perform those reductions!

ngeiswei:
Interesting! So it performs a limited form of reduction (partial evaluation) but still. Of course I strongly advocate implementing something relying on algebraic properties rather than concrete types, cause it will make reduction of GSN and GPN way easier then!!! But since it's just a proof-of-concept.

I had the plan to implement it in Haskell once we have a binding to use Haskell pattern matching. Ultimately I don't mind using C++, as long as the code is elegant enough. I also had this idea of coding it in Felix http://felix-lang.org/ would be a way to use Felix pattern matching and have some very efficient C++ code out of it. But maybe what we should do is coding it Atomese! Using OpenCog pattern matcher, and have the Rule Engine perform those reductions!

linas:
Semi-agree on algebraic properties-- I created a FoldLink to implement both plus and times, so that reduction for PlusLink is just reducation for FoldLink with (0, +) as the operators, and TimesLink is the Fold with (1,*) as the operator.

I want to get away from "generic", black-box GPN/GSN because who the heck knows what `(ExecutionOutputLink ( GSN "py:summation_of_two_atoms") A B)` does, whereas PlusLink is "obvious",  The problem is that any formulas encoded as `py:` or `scm:` in GPN/GSN are black boxes, and one cannot do reduction on black boxes.

It seems that it would not be hard to recreate much of comboreduct in the atomspace.

The code here does not use the pattern matcher at all. Using the pattern matcher for reduction seems like overkill.  Reduction is just not that hard (well, its horrid when using combo trees, but that's mostly because the combo-tree syntax is really hard to understand and use; reduction in the abstract is not that complicated.)

ngeiswei:
One could shed some light in those boxes by representing in the AtomSpace their types and properties, and the Atomese reduct engine would use that to perform reduction. For example
```
SubsetLink
   GSM "py: summation_of_two_atoms"
   ConceptNode "schema_property: symmetric"
```
etc.

PR #1567, opened 2015-04-25
Variable declarations are now optional

linas:
You can now use the BindLink and the SatisfactionLink without bothering to do variable declarations. If no variables are declared, then the body of the link is searched for all free variables, and these are grounded.

### DIFF
--- a/opencog/atoms/bind/BindLink.cc
+++ b/opencog/atoms/bind/BindLink.cc
@@ -31,7 +31,7 @@ using namespace opencog;
 
 void BindLink::init(void)
 {
-	ScopeLink::init(_outgoing);
+	extract_variables(_outgoing);
 	validate_body(_body);
 	unbundle_clauses(_hclauses);
 
--- a/opencog/atoms/bind/ConcreteLink.cc
+++ b/opencog/atoms/bind/ConcreteLink.cc
@@ -27,16 +27,18 @@
 #include <opencog/atomspace/ClassServer.h>
 #include <opencog/atomutils/FindUtils.h>
 #include <opencog/atomspace/Node.h>
+#include <opencog/atoms/reduct/FreeLink.h>
 
 #include "ConcreteLink.h"
 #include "PatternUtils.h"
+#include "VariableList.h"
 
 using namespace opencog;
 
 void ConcreteLink::init(void)
 {
 	_pat.redex_name = "anonymous ConcreteLink";
-	ScopeLink::init(_outgoing);
+	extract_variables(_outgoing);
 	unbundle_clauses(_body);
 	validate_clauses(_varlist.varset, _pat.clauses);
 	extract_optionals(_varlist.varset, _pat.clauses);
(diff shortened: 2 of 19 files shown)


bgoertzel:
Nice one, thanks ;)

On Sun, Apr 26, 2015 at 4:23 AM, Linas Vepštas <notifications@github.com>
wrote:

> Merged #1567 <https://github.com/opencog/opencog/pull/1567>.
>
> —
> <https://github.com/opencog/opencog/pull/1567#event-290065180>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

linas:
Hi Ben,

The merge after this also makes SatisfactionLinks completely optional,
and mabye BindLinks will become optional, too.  What this means is
that you can give the pattern matcher almost any link that has
variables in it, and it will just try to find groundings for them.

In essance, you can now (more or less) give the pattern matcher almost
any link at all, and it automatically just finds the satsifying set
for it.

I say "more or less" because the And Or, NotLinks play a special role
when combined with GPN's. The ImplicationLink is confusing, too:
should it be matched, or should it be treated like a re-write link?
Should I invent a new ReWriteLink, to be used by/instead of BindLink?

All this almost (but not quite) makes the "univeral rule engine"
almost not needed, its almost a built-in part of the pattern matcher.
But not quite.  There are many confusing issues that remain... but I'm
excited by the general direction.

Have you looked at the "fininste state machine" demo, and the "markov
chain" demo? I'm thinking that the URE could almost work like one or
the other of these two demos, and that the PLN forward chainer could
almost work like a generalized markov-chain demo.

anitzkin:
Linas,

You were a bit ambiguous there about whether BindLinks are becoming
optional or not; I'd like to know whether I should continue to
use the variable declarations and BindLinks in writing the scheme R2L rules.

Aaron

linas:
Hi Aaron,

Well, first of all, if you want to do what a BindLink does, you still have to use it.  Its not optional yet.

If I (get around to) make it optional, you'll still be able to use it. In fact, you would get a performance boost, since by explicitly using BindLink, some intermediatte results are cached.  BindLink uses, under the covers, SatsifactionLink to perform that caching.

For general ad-hoc usage, SatsifactionLink is now completely optional. if you do use it, you get a performance boost, because it caches some intermediate results.  But you don't have to explictly state it; it gets used automatically, under the covers, anyway (but without the caching benefits)

williampma:
Does this change mean that variables are now even more implicit instead of explicit?

I also noticed that BindLink and SatisfyingLink no longer inherit from ScopeLink.  What would be the proper way to determine which variables are free variables and which variables are bound variables?  Previously I assumed everything inside ScopeLink are bound (as implemented in https://github.com/opencog/atomspace/blob/master/opencog/atoms/bind/PatternUtils.cc#L185 ) but now that BindLink & SatisfyingLink no longer part of ScopeLink, this is no longer true.

linas:
If you explicitly specify variables, then those variables become bound, just as before; that behavior has not changed, and the distinction between free and bound is as before.  If you fail to specify any variables, then all variables inside the expression become de-facto bound.  

Note however that, due to code structure, you will get better performance if you explicitly specify the variables. (I might try to fix this someday...)

This change was prompted by Nil's rejection of the definition of SatisfyingSetLink as a kind-of Scoped Link. Because of how ScopeLink works, I had to break the dependency in the code.  All this is very confusing. The atomspace is not lambda calculus, nor is it a functional programming language, it is almost a logic language (e.g prolog, minikanren) but not quite. The scoping of variables is confusing.

However, all of the changes should be backwards-compatible; things should work as before.

PR #1569, opened 2015-04-27
Moses segregation

linas:
Moses has been moved to github.com/opencog/moses  If you  clone and pull that repository, you can build run teests and install moses.

-- this merge is step one for removing moses from opencog.  A later merge will remove it completely


ngeiswei:
But why????!!!! MOSES is finally being incorporated to the rest of OpenCog, I really don't understand this move...

ngeiswei:
Well, it's still possible to have both inter-operate of course, but, still, why?

ngeiswei:
Oh, I see it's being discussed in thread
```
Architecture for running multiple mind agents and entire experiment sessions from Python
```
which I have entirely skipped... I'll read it then...

bgoertzel:
Well, Mike Duncan has been asking for a while to have the ability to
download new versions of MOSES without having to download all of OpenCog
each time....


I was not pushing for this by any means -- but it makes some sense to me
since MOSES is being used separately from the rest of OpenCog right now, in
practice...


I definitely am still in favor of having MOSES and the Atomspace
interoperate more and more !! ;)

On Mon, Apr 27, 2015 at 1:41 PM, ngeiswei <notifications@github.com> wrote:

> Well, it's still possible to have both inter-operate of course, but,
> still, why?
>
> —
> <https://github.com/opencog/opencog/pull/1569#issuecomment-96513262>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

ngeiswei:
After having read the second part of the email thread I do understand the move. I just was surprised to see that happening without any prior discussion, but that's cause I missed it!

linas:
Sorry, we've discussed this repeatedly over the years. Basic standard
software design practice says that one should never dump everything into on
giant repo, which is what we've been doing with opencog.  it should have
been 4 or 5 repos all along.

A clean split does not prevent interoperability; it typically makes interop
easier, by making clear what the API's really are, When everything is in
one repo, it has a way of turning into spaghetti code, with poor design,
poor structure.  (e.g. the mess in embodiment as a great example of this)

I've just sort of hit my pain threshold; where it seems less painful to
split things apart into cleanly delineated modules, than to continue as we
are, pulling along all the baggage.

ngeiswei:
It's all good.

PR #1570, opened 2015-04-27
Removal of moses

linas:
This removes moses entirely, and starts a port of the embodiment code to the new moses location.

If you don't have moses installed, opencog should build and tests pass.

If you do have moses moses installed ... it won't build yet ... working on it, though.


ngeiswei:
Linas, you say "starts a port of the embodiment code to the new moses location", you mean "to a new location", right?

linas:
what i meant is that the embodiment code has to compile with moses taken from /usr/local/include header files, and the moses3 namespace.

The header files are easy, but the namespace issues are driving me nutes.

ngeiswei:
I see. I might some experiences with that. Feel free to commit your work so far before going to sleep, and I might finish the work today.

linas:
Hi Nil,

I've committed everything I've got.  It would be great if it worked by
tomorrow.

I forked util, so there are now two copies of this.  I suspect that this is
a bad idea. The right thing would be to split of util into its own project,
and make both opencog and moses depend on it.

If you decide to split it, grab the version in moses: it has some clearnup
w.r.t file paths that the opencog version does not have.

Otherwise, I can split off util tomorrow morning

ngeiswei:
I've done a fair bit of progress but I'm still blocked by a linking error involving boost and DSO, I have no idea what is the problem and I'm not very strong with linking errors. I'll try to get passed that but he can have a look at it you might go through faster than I.

What remains BTW is splitting util into its own library. I can do it but I'd like to have that linking error fixed first.

keyvan-m-sadeghi:
@linas nice. It seems new versions of git have a standard way of giving the
removed code a new home, see:
http://stackoverflow.com/questions/359424/detach-subdirectory-into-separate-git-repository/17864475#17864475

linas:
Hi, yes, I tried that, Basically, it erased the entire history. I did not
want to erase the entire history, so I forked instead.

--linas

PR #1571, opened 2015-04-27
Fix moses segregation

ngeiswei:
Fixed a bunch of compile errors to support the new standalone MOSES.

I'm still struggling with linking error:

```
/usr/bin/ld: ../../../util/libcogutil.so: undefined reference to symbol '_ZN5boost9re_detail12perl_matcherIN9__gnu_cxx17__normal_iteratorIPKcSsEESaINS_9sub_matchIS6_EEENS_12regex_traitsIcNS_16cpp_regex_traitsIcEEEEE5matchEv'
/usr/lib/gcc/x86_64-linux-gnu/4.8/../../../x86_64-linux-gnu/libboost_regex.so: error adding symbols: DSO missing from command line
```
I don't understand yet but I'm merging already so others can look at it if they want (it was still breaking anyway).

### DIFF
--- a/opencog/cython/opencog/pymoses.pxd
+++ b/opencog/cython/opencog/pymoses.pxd
@@ -1,3 +1,3 @@
-cdef extern from "moses/moses/main/moses_exec.h" namespace "moses3::moses":
+cdef extern from "moses/moses/main/moses_exec.h" namespace "opencog::moses":
     int moses_exec(int argc, char **argv) except +
 
--- a/opencog/embodiment/AvatarComboVocabulary/AvatarComboVocabulary.h
+++ b/opencog/embodiment/AvatarComboVocabulary/AvatarComboVocabulary.h
@@ -57,7 +57,6 @@ static const definite_object all_agents = "all_agents";
 // and allow several vocabularies to coexist 
 namespace AvatarCombo {
 
-using namespace moses3::combo;
 using namespace opencog::combo;
 
 //return a pointer to class base instance
--- a/opencog/embodiment/AvatarComboVocabulary/avatar_action_symbol.cc
+++ b/opencog/embodiment/AvatarComboVocabulary/avatar_action_symbol.cc
@@ -24,7 +24,7 @@
 #include "avatar_action_symbol.h"
 #include <moses/comboreduct/type_checker/type_tree.h>
 
-using namespace moses3::combo;
+using namespace opencog::combo;
 using namespace avatar_action_symbol_properties;
 
 avatar_action_symbol::avatar_action_symbol() { }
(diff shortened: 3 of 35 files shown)


linas:
I've resolved this. I've also split out cogutil from moses and put it in its own project in github. I've mostly split it out of opencog as well, -- everything compiles, but, due to a variety of hard-coded paths and etc a large fraction of unit tests fail because assorted modules don't load.  The module load path stuff is a mess.  Care to take stab at it?  The current state of opencog is here:   https://github.com/opencog/opencog/branches  its the branch called "util-remove".

linas:
err, looks like I accidentally pushed to master. I need to revert that. I'm getting tired.

linas:
OK, I reverted all the damage that I accidentaly did, so `master` on opencog should compile and test great with and without moses.  The not-quite working versions with cogutil removed is as mentioned before: here  https://github.com/opencog/opencog/branches its the branch called "util-remove".

ngeiswei:
@linas I didn't manage to make progress on that (just fixed a compile error related to MOSES embodiment). I'm resuming work on it. Have you done any progress?

ngeiswei:
Well, actually I'm making progress (it's crazy how a little nap can expand your consciousness), I'm pretty sure it has to do with DEFAULT_MODULE_PATHS, the CMAKE_INSTALL_PREFIX is not the same anymore as it's using cogutils'...

linas:
OK,  I just pushed somethig to cogutils that extends the search patchs pull
now

linas:
I changed the search paths in cogutils/opencog/util/files.cc and now all
opencog unit tests pas just fine.

So I thought I would try again, this time with moses installed, and ... now
I am getting the weird shared library DSO message again, which went away
yesterday. ...

ngeiswei:
Regarding DSO just pull the last util-remove.

ngeiswei:
Regarding DEFAULT_MODULE_PATHS, maybe it's better to move it to where modules are compiled (currently in the opencog repo), rigth?

linas:
Yes. I hacked the solution last night while I was tired and ran out of
energy for anything better .

ngeiswei:
OK, I'll move DEFAULT_MODULE_PATHS to CogServer.{h,cc}, where I think it makes more sense.

linas:
Yes, except half of the unit tests don't use the cogserver

linas:
and the goal is to depend on the cogserver less and less

ngeiswei:
But the cogserver does load the modules, right?
```
CogServer::loadModule
```

ngeiswei:
though of course if the plan is to kill the cogserver (better do it now while it's not self-aware), then I guess it matters little to have DEFAULT_MODULE_PATHS defined in it. So the question is what is gonna load the modules after?

ngeiswei:
Anyway, all tests pass I guess we can merge util-remove to the master.

linas:
if it works, then merge

ngeiswei:
Merging... (have to figure out how to deal with the delete conflicts).

PR #1573, opened 2015-04-28
Util remove

ngeiswei:


ngeiswei:
Addresses issues discussed in #1571

PR #1580, opened 2015-04-30
Module move fixes

ngeiswei:
The following unit tests were failing due to not having the right module paths

RuleUTest
AtomSpacePublisherModuleUTest
PythonModuleUTest


AmeBel:
Some failing tests http://buildbot.opencog.org:8010/builders/opencog_master_trusty/builds/522/steps/test/logs/stdio

PR #1582, opened 2015-04-30
Fix cog-bind human-implies-animal-stv

ArvinPan:
Bug found when trying to run:
guile> (cog-bind human-implies-animal-stv)

ExecutionLink should be ExecutionOutputLink.

By the way, the document which describes the pattern matcher need to be update:)
https://docs.google.com/a/singularityu.org/presentation/d/15C2lheJNV6UtrIYcJ2zt5UOui_PhxmMNf2OzwGS5oq8/edit?pli=1#slide=id.p
page 9, Example 2

(cog-bind human-implies-animal) should be (cog-bind human-implies-animal-stv)

### DIFF
--- a/examples/pattern-matcher/simple.scm
+++ b/examples/pattern-matcher/simple.scm
@@ -67,7 +67,7 @@
            (VariableNode "$H")
            (ConceptNode "human")
         )
-        (ExecutionLink
+        (ExecutionOutputLink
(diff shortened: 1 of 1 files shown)


linas:
thanks!

FYI, though, you arrive just at the tail-end of a large reorganization.

These examples have just been moved to github.com/opencog/atomspace

PR #1583, opened 2015-04-30
More module path fixes so that it compiles on the builbbot

ngeiswei:
Should fix the errors pointed out by AmeBel on #1580


AmeBel:
Thanks, tests all pass now. I will add new builders for the new repos, soon.

PR #1585, opened 2015-05-01
Move pln r2l back to reasoning

ngeiswei:
PLN doesn't work because default_cpolicy.json is not loaded. The code that controls that is still on the atomspace. Obviously this is bad, we need a more flexible way to control that.

### DIFF
--- a/opencog/reasoning/default_cpolicy.json
+++ b/opencog/reasoning/default_cpolicy.json
@@ -1,0 +1,92 @@
+{
+	"rules": 
+	[
+		{
+			"name": "pln-rule-deduction",
+			"file_path": "reasoning/pln/rules/deduction.scm",
+			"priority": 1,
+			"mutex": 
+			[
+				
+			],
+
+			"category": "PLN"
+		},
+
+		{
+			"name": "pln-rule-modus-ponens",
+			"file_path": "reasoning/pln/rules/modus-ponens.scm",
+			"priority": 1,
+			"mutex": 
+			[
+				
+			],
+
+			"category": "PLN"
+		},
+
+		{
+			"name": "pln-rule-contextualize-inheritance",
+			"file_path": "reasoning/pln/rules/contextualize.scm",
+			"priority": 1,
+			"mutex": 
+			[
+				
+			],
+
+			"category": "PLN"
+		},
+
+		{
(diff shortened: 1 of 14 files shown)


linas:
!?

I struggled a bit with the json path loading yesterday, I thought I'd fixed it all -- the unit tests were passing.   The issue was that...

Oh. Hey. this issue is this: if you are using PLN in opencog. then .. the default_cpolicy.json files are not installed to /usr/local/share/opencog  and so are not being found/loaded ... I am guessing that is the bug ...

linas:
... and, as a result of this commit,  'make test' in the atomspace now fails:
```
The following tests FAILED:
	 58 - RuleUTest (Failed)
	 62 - JsonicControlPolicyParamLoaderUTest (Failed)
```
I'm pretty sure that if we installed the json files, everything would have worked.

ngeiswei:
Linas, keep in mind that the atomspace should no longer rely on default_cpolicy.json because this refers to PLN rules, and PLN rules have been moved to opencog.

It's kinda weird neither RuleUTest nor JsonicControlPolicyParamLoaderUTest fail for me.

linas:
When I run RuleUTest, I get this error:

ERROR: In procedure apply-smob/1:
ERROR: Unbound variable: pln-rule-not
ABORT: unbound-variable

no surprise, because earlier is this error:

[WARN] Failed to load file reasoning/pln/rules/not.scm: 2 No such file or
directory

So RuleUTest seems to be trying to load the files that you just removed ...

-- Linas

linas:
perhaps, for RuleUTest only, we could copy a tiny subset of the pln rules
into the /tests/rule-engine directory, just enough to allow it to pass the
unit test.

ngeiswei:
Indeed, I'm puzzled as to why it passed on my system (I did remove the reasoning directory from atomspace). I've added example rules under examples/rule-engine/rules, I think those could be used.

Note I'll be on vacation next week, I may not have time to look into this but I should be able to check my emails regularly.

linas:
with a split like this, old garbage remains in the build directory, so when I forget to clear out build, weird stuff happens. No hurry, as long as we don't forget.

linas:
I created https://github.com/opencog/atomspace/issues/5 to track this

PR #1594, opened 2015-05-05
Added johnny example and modified pln NotRule

prateeksaxena2809:
Added example for first syllogism from List of Test Syllogisms mentioned [here] (http://wiki.opencog.org/w/List_of_Test_Syllogisms#Initial_test_list).

Also modified NotRule to be generic and handle nested NotLinks.


linas:
Could you  move this to /examples/pln/johnny ? OpenCog contains other things besides PLN

prateeksaxena2809:
@linas : Sure.

linas:
I think you can do this in-place: just do git mv  and then push to this same branch, and this same pull request will then show the new work.

opencog/reasoning/pln/rules/not.scm
```diff
@@ -8,39 +8,33 @@
...
-(define (pln-formula-not A)
-  (cog-set-tv!
-   (NotLink A)
-   (pln-formula-not-side-effect-free A))
-)
+(define (pln-formula-not nA A)
+  (if (eq? (cog-type A) 'NotLink)
+	(cog-set-tv!
+		(gar A)
+		(pln-formula-not-side-effect-free A))
```
AmeBel: Not of Not A = A, here it is Not A.

examples/pln/johnny/Assertions.scm
```diff
@@ -0,0 +1,50 @@
...
+(NotLink (stv 0.99 1)
+	(EvaluationLink
+		(PredicateNode "every@111")
+        (ListLink
+			(ConceptNode "day@222")
+			(EvaluationLink
+				(PredicateNode "eating@333")
+				(ListLink
+					(ConceptNode "Johnny@444")
+					(ConceptNode "Sweets@555"))))))
```
AmeBel: The above ImplicationLinks are not the output of running 

```
(nlp-parse "If Johnny is eating sweets every day, he is placing himself at rish for diabetes.")
(nlp-parse "Johnny does not eat sweets everyday.")  
```
Which means we can't do inference without human intervention, which is not the desired goal for r2l. @anitzkin what do you suggest ?

opencog/reasoning/pln/rules/not.scm
```diff
@@ -8,39 +8,33 @@
...
+	(cog-set-tv!
+		nA
+		(pln-formula-not-side-effect-free A))))
 
 (define (negate x)
```
AmeBel: rename the function to negate-strength, as there is a negate function in guile https://www.gnu.org/software/guile/manual/guile.html#Procedure-Properties

opencog/reasoning/pln/rules/not.scm
prateeksaxena2809: if A is ```NotLink```, ```(gar A)``` removes the ```NotLink``` which is what we want by doing Not of Not A.

AmeBel: my bad

examples/pln/johnny/Assertions.scm
AmeBel: @anitzkin as requested, the r2l outputs for the two sentences 
https://gist.github.com/AmeBel/1ed86e75240013cb9f4f
https://gist.github.com/AmeBel/f5bd8c73f5ec7dd60370

anitzkin: Lol.  There's so much that's right and so much that's wrong about those!
This is why there were quite a few of the old
R2L rules that I didn't even bother porting at all.  Like the not-rule that
was there.  I mean, what?  I will have to look at
both the LG and relex output also to figure out exactly what it's doing,
here.  I mean it looks to me, like the LG or relex output
for those sentences is probably not what it should be to interface
correctly with the R2L rules . . .  It's like trying
to put together one of those puzzles made of wooden blocks, where as soon
as you fit one piece in, three of the ones
that were in before, fall out . . .

AmeBel: and the universe explodes :smile:

linas: > . . .  It's like trying
> to put together one of those puzzles made of wooden blocks, where as soon
> as you fit one piece in, three of the ones
> that were in before, fall out . . .
>
>
Well, that is exactly the optimization problem I keep talking about. You
cannot just add on new rules, without also going back to modify the rules
you learned previously.  The older rules were "correct" when dealing with
the smaller corpus, but, when enlarging your corpus, you find that some of
the old rules result in bad outcomes, bad deductions.

You have to go back and modify the old rules.  This is like Ben's email
where he asks "what N can I choose for language" and the answer is you
can't: its neither global nor static. You worked with a small value of N at
first, and learned some confident rules for a limited corpus.

As you start hearing, handling, processing new classes of sentences, one of
two things happen: either some of the existing rules are confirmed, in
which case your confidence of those rules get *even bigger*, their status
as bedrock rules is further strengthened.  The other possibility is that
the existing rules get modified.  To find the right modification, you have
to go to deeper bedrock, and solve the global problem of (deeper_bedrock +
modified_rules + new_corpus) is minimally conflicted, inconsistent.

This makes the learning process hierarchical, and buries some rules deeply,
where they can't be modified without tearing down the whole edifice. Other
rules remain superficial and malleable, and don't harden into a fixed form
until a lot of corpus experience is built up.

-- Linas

anitzkin: Okay, I've examined the output for this more carefully now.  Actually, it's
almost all right. The two errors appear to be the interpreting of "risk" as
an indirect object of placing, which is what you all wanted originally, but
is an error in my mind, that is a result of one of the R2L rules for SVIO
-- that one I can correct right away.  The other error is in what R2L is
doing with the "if" which has to do with the last new rules I added, which
were the first rules we had for such clauses, so that may take some
fiddling . . .

I don't think anything else is wrong with it.  It just looks odd, because
it has all of the dependency relations treated separately, as we've been
discussing -- either new post-processing or PLN is required to put them
together into more wholistic structures.

anitzkin: And . . . okay, I've fixed the SVIO rule and added a rule to take care of
that "if clause.  However, I can't seem to push those for the moment,
because I get a 403 web error when I do . . .

prateeksaxena2809: Hello @anitzkin, have you been able to push your changes?

anitzkin: Nope. I keep getting the 403 error and I have not been able to correct it
yet. Will keep trying.

anitzkin: Okay, that's done now . . . just waiting to be merged, which may take a
while, because there were some commits stacked up there.

AmeBel:
@prateeksaxena2809 this can't be merged now, please update. Thanks

PR #1602, opened 2015-05-07
Assorted python/cython fixes.

linas:
The primary goal of these fixes, and those from yesterday, is to enable using python with the cogserver. In particular, a goal was to get python to use the cog-servers atomspace, instead of working in it's own private atomspace.  It now does this, and /examples/python/README.md now explains how to actually make this work.

It works for me, so this should be the last of the changes for a while.


linas:
p.s. this needs matching fixes to the atomspace, in order to get the full function.

PR #1605, opened 2015-05-08
A bunch of R2l rules ported to atomese -- all tested, all work

anitzkin:
The first versions of these were written before I learned about ChoiceLink, so I didn't bother re-doing them that way.  The SV rule uses a bunch of Absent Links to prevent it from being invoked in the cases of the many other rules it could overlap with; the rest of them, except for svo, don't appear to require anything in the way of exclusions or priorities; they contain uniquely identifying conditions.   I need to add one AbsentLink to svo to exclude _iobj.


anitzkin:
Um, only the last two of these commits count.  I closed the pull request on the rest of them.  Is there any way to get rid of all of those earlier commits? Or the whole shebang and then I'll re-commit the latest versions?

linas:
If its really true that you only want the the last two commits, then you can use 'git cherry-pick' ... however, cherry-picking is a kind-of more advanced git function, it assumes you are already adept at the basics.  It also requires that the last two commits really are independent of what came before.

If you are desperate, I can do this for you, but it is usually best to avoid getting into these situations to begin with.  The whole point of using branches in git is that you can work in one, or another, toggle between them, ignore, eventually discard branches you don't like, merge the ones you do like. So if you knew you were going down the wrong way at thee time, you should have back-tracked, and created a new branch.

PR #1611, opened 2015-05-12
Copy rule files to build dir

misgeatgit:


williampma:
Umm.. why?  Didn't we undo this before? I thought the .json is supposed to look at the DEFAULT_MODULE_PATH.

misgeatgit:
It wasn't loading for me.

williampma:
Then shouldn't it be fixed in the loading code?  It just seems a bad idea to copy these stuff to the build directory.

misgeatgit:
OK, Appending opencog to the existing path seems to work. I will revert this one.

williampma:
Ok, thanks!

PR #1613, opened 2015-05-12
add include of formulas.scm to deduction rule

eddiemonroe:
I think this include needs to go here if I am understanding things correctly.

### DIFF
--- a/opencog/reasoning/pln/rules/deduction.scm
+++ b/opencog/reasoning/pln/rules/deduction.scm
@@ -4,7 +4,7 @@
 ;
 ; AND(Inheritance A B, Inheritance B C) entails Inheritance A C
 ; -----------------------------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


opencog/reasoning/pln/rules/deduction.scm
```diff
@@ -4,7 +4,7 @@
 ;
 ; AND(Inheritance A B, Inheritance B C) entails Inheritance A C
 ; -----------------------------------------------------------------------------
-
+(include "formulas.scm")
 (define pln-rule-deduction
```
linas: Although include probably works here, you probably want to use load-from-path instead.

scheme include is kind-of-like a c-pre-procosseor #include while load-from-path  is like a dynamic library load.   So, here, you do not need to inline the formulas, you just need to make sure they are linkable.

PR #1615, opened 2015-05-14
Okay . . . can we get this sorted out already

anitzkin:
So, as far as I can tell, because of the byzantine nature of github, its best just to merge ALL the commits including the most recent one that corrects all the errors in the previous ones, as well as adding a lot of work, that I've been doing during the days when I couldn't figure out what to do with github.  

Anyway, the most recent commit is the way things ought to be -- all the files in the right place, and all of the new rules truly tested and working, so just merge it for god's sake.


AmeBel:
can't merge there is a merge conflict.

anitzkin:
Yes, well, what would you suggest I do about that.  I wanted to just cancel the earlier commits, but apparently it's impossible to do something that simple . . .

AmeBel:
you can remove earlier commits using `git cherry-pick`.

anitzkin:
Yes, I've read about it twice now and as far as I can understand the documentation, which is almost not at all, it doesn't seem to mention cancelling commits.  I also can't figure out why I can't push anything to github from my relex repositories. I suppose I should spend a day studying github, which I've avoided since I got here, because it seems retarded to have to learn such a complex system just to upload file changes.

linas:
If you think github is complicated, then clearly you have never used SVN or CVS or bazaar or ...pretty much anything that came before.  In the space of a few years, it picked up something over 75% market share and essentially eliminated multiple competitors. Why? Ease-of-use.  its light-years ahead of the competition.

Anyway, I'm guessing that perhaps 'git rebase' will maybe rescue you from the situation you are in.

linas:
Ah, I see part of the problem. You somehow are using sub-projects. Never ever ever ever never ever use subprojects.

linas:
Rather than watch aaron squirm in pain, I just cherry-picked the commits and applied them in pull requeest #1617

linas:
Aaron, at this point, what you want to do is this:
```
git checkout master
git pull
```
This will get you all the latest code, and reset your github repo into a sane state.  In the future, please avoid using `git merge` until you have studied how git works, and have an absolutely burning need for it.  In general, you should avoid using `git merge`; during day-to-day ordinary development, there is almost never any reason to use it.

PR #1619, opened 2015-05-15
new commits -- thought the old ones were resolved, drat it!

anitzkin:
I thought that the previous mess had been resolved this morning before I added these commits. Sorry.


anitzkin:
So it is now everything in the last 3 commits that we want to keep.  These three were made on a different branch than the others. Don't know why they're being appended to the earlier ones.

linas:
please stop doing this

linas:
aaron, your code has been checked in.  Get someone over there to show you how to set your git branch to master, to get pull, and then how to create a new branch for you to work in.  If there are problems with what is curretnly, you should work to resolve those problems, instead of continuing to pile on with the mistakes and errors.

anitzkin:
Linas, I do all that.  I committed my new work because I thought the
previous commits had been cherry picked and resolved.

PR #1620, opened 2015-05-15
Attempt to fix openpsi utests

ngeiswei:
This branch attempt to fix the openpsi utests. I made progress but couldn't get the test completed fixed. I'm merging now because we might not bother keeping those tests after all, and this branch contains a bunch of cleanings that are probably worth having.


AmeBel:
thanks

PR #1630, opened 2015-05-21
Rearrange SpaceServer

chenesan:
Do some rearrangement:
* Since there should be multiple scenes in the new embodiment code, the user may want to call function on different space map. In the previous code we can only operate on the single curMap in some functions, so I add a spaceMapHandle parameter in these functions and change the implementation, allowing user to input maphandle to operate on different map:

    addSpaceInfo

    removeSpaceInfo

    setAgentHeight

    findAllBlockEntitiesOntheMap

    addBlockEntityNodes

    addBlockListPredicatetoEntity

    addBlockEntityProperties

    updateBlockEntitiesProperties

  Also, I change the code in PAI.cc and MockOpcHCTest.cc to fit the change of the interface.
* Add a cloneSpaceMap function to clone any map in spaceMaps. 
* comment out some unused data members and function because they were only used in the commented function. 
* rearrange comment format in SpaceServer.h
I'm not sure if it's OK to delete those commented functions in SpaceServer.*. They make reading code much harder...

PR #1634, opened 2015-05-30
Add SpaceServerUTest

chenesan:
It's the test for #1630 , Can @ngeiswei comment on this?
Just test some basic operation: add/get map, add/remove space info in map.
I accidentally commited the cython binding of spaceserver (that's unfinished), so the commit history is a little messy..


tests/spacetime/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+LINK_LIBRARIES(
+	spacetime-types
+	spacetime
+)
+
+ADD_CXXTEST(SpaceServerUTest)
```
linas: why create a new directory? why not just put it into the spatial directory?  There already is a spaceTest there, and a time test.

chenesan: Because the SpaceServer code is put in the opencog/spacetime/, so I think I should put the test in the directory with the same name, e.g. tests/spacetime. But I think it's OK to put it in the tests/spatial.

PR #1644, opened 2015-06-11
add module for running cog-execute and related commands

AmeBel:
Fixes https://github.com/opencog/atomspace/issues/87

### DIFF
--- a/lib/opencog.conf
+++ b/lib/opencog.conf
@@ -81,7 +81,8 @@ MODULES               = opencog/server/libbuiltinreqs.so,
                         opencog/embodiment/AtomSpaceExtensions/libembodiment-types.so,
                         opencog/embodiment/AtomSpaceExtensions/libAtomSpaceExtensions.so,
                         opencog/modules/python/libPythonModule.so,
-                        opencog/modules/libRuleEngineModule.so
+                        opencog/modules/libRuleEngineModule.so,
+                        opencog/modules/libExecModule.so
 
 # Optional modules, not enabled by default
 #                        opencog/modules/benchmark/libbenchmark.so,
--- a/opencog/modules/CMakeLists.txt
+++ b/opencog/modules/CMakeLists.txt
@@ -112,3 +112,18 @@ IF (HAVE_GUILE AND HAVE_LINK_GRAMMAR)
 		LIBRARY DESTINATION "lib${LIB_DIR_SUFFIX}/opencog/modules"
 	)
 ENDIF (HAVE_GUILE AND HAVE_LINK_GRAMMAR)
+
+# ----------------------------------------
+# Build the execute module
+
+IF (HAVE_GUILE)
+	ADD_LIBRARY (ExecModule SHARED ExecModule)
+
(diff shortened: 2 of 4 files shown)


linas:
Can you retry, without your patch, and instead just do this, inside the cogserver:
```
(add-to-load-path "/usr/local/share/opencog/scm")
(use-modules (opencog exec))
```

The above seems to work for me, and if the above works for everyone, we should probably do that instead.   And if the above works, we should probably get rid of some of the other opencog modules as well, like libQueryModule.so  (under the theory of "less code" means "less confusion about how to configure" and fewer bugs to track down.)

AmeBel:
It works for me too.

AmeBel:
tried  (opencog query)  after removing libQueryModule.so, it also works

Got some errors, like
```
Error: Backtrace:
In ice-9/boot-9.scm:
 157: 13 [catch #t #<catch-closure 24a7040> ...]
In unknown file:
   ?: 12 [apply-smob/1 #<catch-closure 24a7040>]
In ice-9/boot-9.scm:
 157: 11 [catch #t #<catch-closure 30dc1c0> ...]
In unknown file:
   ?: 10 [apply-smob/1 #<catch-closure 30dc1c0>]
   ?: 9 [call-with-input-string "(load \"/usr/local/share//opencog/scm/core-docs.scm\")" ...]
In ice-9/boot-9.scm:
2320: 8 [save-module-excursion #<procedure 3345060 at ice-9/eval-string.scm:65:9 ()>]
In ice-9/eval-string.scm:
  44: 7 [read-and-eval #<input: string 3352c30> #:lang ...]
  37: 6 [lp (load "/usr/local/share//opencog/scm/core-docs.scm")]
In ice-9/boot-9.scm:
2320: 5 [save-module-excursion #<procedure 2459580 at ice-9/boot-9.scm:3961:3 ()>]
3966: 4 [#<procedure 2459580 at ice-9/boot-9.scm:3961:3 ()>]
In unknown file:
   ?: 3 [load-compiled/vm "/home/opencog/.cache/guile/ccache/2.0-LE-8-2.0/usr/local/share/opencog/scm/core-docs.scm.go"]
In /usr/local/share/opencog/scm/core-docs.scm:
 760: 2 [#<procedure 30e98e0 ()>]
In ice-9/boot-9.scm:
 102: 1 [#<procedure 2459640 at ice-9/boot-9.scm:97:6 (thrown-k . args)> unbound-variable ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 30dc180> unbound-variable ...]

ERROR: In procedure apply-smob/1:
ERROR: In procedure module-lookup: Unbound variable: cog-bind
ABORT: unbound-variable
```
i guess it is from SCM_PRELOAD

AmeBel:
If i am not mistaken https://github.com/opencog/atomspace/blob/master/opencog/scm/config.scm should be part of this repo, right? Or we can add a new config.scm that will add the paths.

linas:
yes,

cnfig.scm should probbly add the paths, and also do the (use-modules) thing.

PR #1646, opened 2015-06-15
New rules pln

prateeksaxena2809:
I have added/modified some boolean pln rules. Please provide feedback.


ngeiswei:
Looks like this hasn't been merged. I noticed you're still using the ImplicationLink inside the BindLink, this no longer the right format, instead use just a BindLink with 3 arguments, variables, pattern, rewrite. I guess that is why it hasn't been merged.

anitzkin:
No.  You removed all the ImplicationLinks from the rules I had previously written, which had been merged.  And now i wrote some new ones, in your new style, and all of them (new and old) are giving me this same error.

anitzkin:
Oh. Oops, you're talking about Prateek's work.  I responded because I just opened an issue about a problem with the revised relex rules not working . . . take a look at that issue!

PR #1647, opened 2015-06-15
Added/Modified some boolean rules

prateeksaxena2809:


prateeksaxena2809:
Its a group of three transformation rules named [create_boolean_transformation_rules](https://github.com/opencog/opencog/blob/master0/opencog/python/pln_old/rules/boolean_rules.py#L136). Also, how do I denote double implication in opencog?

AmeBel:
It is best to create 3 separate transformation rules.

@ngeiswei please comment on the double implication, thanks

AmeBel:
@moikle How is a double implication denoted in opencog?

moikle:
@amenbelayneh What do you mean exactly by a double implication?

AmeBel:
@prateeksaxena2809 correct me if I am wrong

@moikle I am assuming prateek's question is in line with https://math.stackexchange.com/questions/266563/double-implication

prateeksaxena2809:
Simply put, if 
``A=>B`` 
and 
``B=>A``
 then we can say that
``A <=>B``
meaning that A and B have a double implication relation . In the old pln, [the following rule](https://github.com/opencog/opencog/blob/master/opencog/python/pln_old/rules/boolean_rules.py#L139) transforms a double implication link into two implication links.

bgoertzel:
"Double implication" is called "Equivalence"

On Thu, Jun 18, 2015 at 10:36 PM, Prateek Saxena <notifications@github.com>
wrote:

> Simply put, if
> A=>B
> and
> B=>A
> then we can say that
> A <=>B
> meaning that A and B have a double implication relation . In the old pln, the
> following rule
> <https://github.com/opencog/opencog/blob/master/opencog/python/pln_old/rules/boolean_rules.py#L139>
> transforms a double implication link into two implication links.
>
> —
> <https://github.com/opencog/opencog/pull/1647#issuecomment-113175026>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

prateeksaxena2809:
@bgoertzel :- So, 
```
EquivalenceLink
        A
        B
```
is how the relation will be defined in PLN?

AmeBel:
yap

opencog/reasoning/pln/rules/and-breakdown-rule.scm
```diff
@@ -0,0 +1,45 @@
...
+(define pln-rule-and-breakdown
+  (BindLink
+   (VariableList
+    (VariableNode "$A")
+    (VariableNode "$B"))
+   (AndLink
+    (AndLink
+     (VariableNode "$A")
+     (VariableNode "$B"))
+    (VariableNode "$A"))
```
AmeBel: the VariableNode $A is redundant

opencog/reasoning/pln/rules/and-simplification-rule.scm
```diff
@@ -0,0 +1,51 @@
...
+
+(define pln-rule-and-simplification
+  (BindLink
+   (VariableList
+    (VariableNode "$A")
+    (VariableNode "$B")
+    (VariableNode "$C"))
+   (AndLink
+     (VariableNode "$A")
+     (AndLink
```
AmeBel: It would be best if the AndLink is a VarialbleNode restricted to an AndLink in the variable declartion. That way it would work with n-ary AndLink

AmeBel: never mind I see the problem with my suggestion for the backward chainer

williampma:
How is the AndBreakdownRule and AndEliminationRule supposed to work together?

AndEliminationRule is trying to set A's TV & B's TV to be based on the AndLink's TV, while AndBreakdownRule is trying to set B's TV based on AndLink's TV & A's TV.

So given the same AndLink, B's TV could potentially be different depends on which rule is applied?  (both rules will fire since both rule's input will always match)

AmeBel:
These are just  a direct port of the old rules. 

In the old pln, there wasn't any concept of a default control-policy, which meant that you had to define that macro rules that would be applied on a case by case bases. But the issue you pointed out is a relevant one. 

The use case I can see for these rules is in a situation when the insertion of an AndLink with stv is made with its outgoing set merely having default stvs, and we would like to propagate that tv over the outgoing set. Now how would the system identify such type of Andlinks, even if we have either one?

PR #1651, opened 2015-06-19
more new rules

anitzkin:


anitzkin:
Bunch o' new rules.  This sort of illustrates the problem I was talking about with the current way we do R2L -- lots o' rules -- but they should do the job . . .

AmeBel:
* Would you please add an example sentence on each rule or add a comment on each rule refering to the equivalent entry in https://github.com/opencog/opencog/tree/master/opencog/nlp/relex2logic
* Please add your suggestion for improvements in the README.md too

AmeBel:
did you get to wirte the function/control-policy that will execute these rules on newly parsed sentences?

anitzkin:
I haven't.  I mean, I  could write a simple script to just execute all of
the rules, but since that would be a temporary expedient rather than a
permanent part of the system it kind-of seems like a waste of time.   I
figured I should keep writing rules instead, since I seem to be only one
who has any idea of what the rules are that still need to be written (I
could write a lot more rules than we already have without exhausting what
can be done with the current relex output).   Why don't we have the
discussions necessary to decide how the control policy should be done in
the long run, not to mention discussing different approaches to the
rule-set to make a more flexible and efficient R2L, and what kind of post
processing may be necesary to make the output amenable to PLN?  If we plan
ahead a little, perhaps you won't need to redo the whole thing again in six
months?

AmeBel:
sure maybe latter today?

anitzkin:
k

PR #1652, opened 2015-06-19
Added/modified context rules

prateeksaxena2809:


AmeBel:
Don't know why we have the context-free-to-sensitive rule, it seems to be a wrong rule. It just takes 2  random atom and makes one a context defintion while the other a context observeration. 

Merging, as  context-free-to-sensitive rule was pre-existing and the edit on it is harmless.

PR #1660, opened 2015-06-28
add sudo to make install

suriyadeepan:
execution of setup.sh stopped while building; made a minor change to ocpkg script to add sudo to make install...


AmeBel:
thanks

PR #1667, opened 2015-06-30
Minor SuReal update

leungmanhin:
Add a new argument to SuReal so that we have control over the number of results we can get from SuRealPMCB

### DIFF
--- a/opencog/nlp/sureal/SuRealPMCB.cc
+++ b/opencog/nlp/sureal/SuRealPMCB.cc
@@ -28,6 +28,8 @@
 
 #include "SuRealPMCB.h"
 
+#define MAX_RESULTS 20
+
 
 using namespace opencog::nlp;
 using namespace opencog;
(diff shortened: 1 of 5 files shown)


opencog/nlp/sureal/surface-realization.scm
```diff
@@ -30,13 +30,22 @@
...
+	(case-lambda
+		((setlink) (sureal-main setlink 20))
+		((setlink thoroughness) (sureal-main setlink thoroughness))
+	)
+)
+
 ; A SetLink is the input because it is assumed that the output of the micro-planner
 ; is unordered.
-(define (sureal a-set-link)
+(define (sureal-main a-set-link thoroughness)
```
williampma: Instead of splitting `sureal` into `sureal` and `sureal-main`, I think they can be in the same function (they are short enough, and will be less confusing).

Also, you have default as 20.  I think the default should be "search completely" (ie. 100%), meaning there should be some way to allow searching 100% (which I don't think is possible with the current implementation yet).

opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -335,38 +336,58 @@ bool SuRealPMCB::initiate_search(PatternMatchEngine* pPME)
...
+    for (auto& c : sCandidate)
     {
-        logger().debug("[SuReal] Loop candidate: %s", c->toShortString().c_str());
+        logger().debug("[SuReal] Loop candidate: %s", c.handle->toShortString().c_str());
 
-        if (pPME->explore_neighborhood(bestClause, bestClause, c))
+        if (pPME->explore_neighborhood(bestClause, bestClause, c.handle))
             return true;
+
+        if (m_results.size() == m_thoroughness) return true;
```
williampma: If some grounding add 2 more more results at a time, it would skip over your m_thoroughness.  Better make it `>`.  Allowing complete search would also be handled here.

opencog/nlp/sureal/SuRealPMCB.h
```diff
@@ -46,7 +46,7 @@ class SuRealPMCB :
     public DefaultPatternMatchCB
 {
 public:
-    SuRealPMCB(AtomSpace* as, const std::set<Handle>& vars);
+    SuRealPMCB(AtomSpace* as, const std::set<Handle>& vars, size_t thoroughness = 20);
```
williampma: Default param is no longer need I think.

leungmanhin:
Thanks @williampma 
I've updated the code accordingly, please have a look

williampma:
Thanks!

PR #1677, opened 2015-07-15
add time domain in the TimeServer

chenesan:
Why to do this: 

Sometimes in the embodiment we may have different time zone. For example, when we play the game we know that the time in game is in the morning and the time in the real world may be 9pm. We have different time domain concepts. And in the GSoC Minecraft embodiment we find that we may want to save both the ROS timestamp and Minecraft timestamp. Because the unit of Minecraft timestamp is too large(1 sec per tick), we need ROS timestamp to help us save the order of event; but the Minecraft timestamp is still necessary for knowing the world(for example, judging if the monster will appear or not). So we want to keep those two time domain in the TimeServer. I think we can regard this change as the same one in SpaceServer, to save multiple spaceMap in SpaceServer.

What changed:

*Add a map<timedomain, temporaltable*> in the time server, and get the table by finding it in this map
*Add function: 
getTimeDomain()
getTimeDomains()
existTimeDomain()
to get/query time domain in the TimeServer.
Notice that the getTimeDomain() is used for old code which assume there's only one time domain in the system. 
*Change the interface of TimeServer. Add a time domain param.
*fix all the code using the TimeServer by adding the timedomain in the argument(mostly just call the timeServer().getTimeDomain())

I haven't added unit test to test the multiple time domain. but I have checked that I can fix and pass the unit tests (without anaphora, pythonmodule, plnrule, which is broken in my environment before this change.)under these changes.


linas:
Nil should review and merge; I've not very really looked into code in this area.

ngeiswei:
I don't mind to review it, but I won't have time to do that today or tomorrow.

I wonder if the right trick would be to replace octime_t by [boost::datetime](http://www.boost.org/doc/libs/1_58_0/doc/html/date_time.html), but anyway, I should first review it...

chenesan:
Hi @ngeiswei 
Sorry for the absence in last week..
Is there anything more I can do in this pull request?

ngeiswei:
Hi @chenesan I'm now available to review that. Expect a feedback from me within the next couple days.

ngeiswei:
OK, rereading the comment in your pull request I understand that time domain is not equal to time zone. Is it really all you want? Real time vs virtual world time?

I'm not pleased with the extra complexity your code adds, especially since I suspect you can avoid it and get more functionality in return (a way to manipulate temporal knowledge from different time domain, if that makes sense, maybe you don't want that anyway).

We probably should have a chat...

ngeiswei:
@chenesan I couldn't find you on Slack or G+, yet you seem to be real.

chenesan:
Hi @ngeiswei :
I'm YiShan. It seems that the github account name misleaded you. This name (chenesan) I have used for a long time. 
Thanks for the review, and I'm sorry that I didn't format/indent consistently in the code.
The time domain is for real time and virtual time. And I'm sorry that I use the time domain name "utc" and "gametime" in the test. It misleads you that I want to add a time zone in TimeServer..But I think it can also be used for expressing different time zone so I wrote it in the test. 
I agree it will make thing clearer to have a chat on slack. I will send you slack message later!

chenesan:
Hi @ngeiswei :
I have fixed the problem you suggested in the comment:
* fix the indent as 4-space, and obey the coding format in existing code
* using plain temporalTable object
* add DEFAULT_TIMEDOMAIN and put the timeDomain as optional argument. Except the addTimeInfo and removeTimeInfo, I put it in the last parameter. As for the addTimeInfo() because I think the truth value is seldom used so I put time domain in front of it. Though this will affect the code under dynamics/openpsi.( we have to pass DEFAULT_TIMEDOMAIN to the addTimeInfo)
* remove almost all the time domain dependence in client code of TimeServer (except the code under dynamics/openpsi) so it's clean as before. As for SpaceServer because new embodiment will use it so I add the time domain as the default argument in some functions.

opencog/embodiment/AtomSpaceExtensions/AtomSpaceUtil.cc
```diff
@@ -1605,14 +1599,13 @@ Handle AtomSpaceUtil::setPredicateValue( AtomSpace& atomSpace,
 Handle AtomSpaceUtil::addPropertyPredicate(AtomSpace& atomSpace,
         std::string predicateName,
         Handle object,
-        TruthValuePtr tv,
-        bool permanent,
-        const Temporal &t)
+	TruthValuePtr tv,
+	bool permanent,
+	const Temporal &t)
```
ngeiswei: Here is a tab, while the indentation style of that file seems to be 4-whitespace.

opencog/embodiment/AtomSpaceExtensions/AtomSpaceUtil.cc
```diff
@@ -2122,7 +2115,7 @@ std::string AtomSpaceUtil::getObjIdFromName( AtomSpace& atomSpace,
 
 Handle AtomSpaceUtil::getMostRecentPetSchemaExecLink(AtomSpace& atomSpace,
         unsigned long timestamp,
-        bool schemaSuccessful)
+	bool schemaSuccessful)
```
ngeiswei: Is a tab while the indentation style of that file seems to be 4-ws.

opencog/embodiment/AtomSpaceExtensions/AtomSpaceUtil.cc
```diff
@@ -2419,9 +2412,9 @@ Handle AtomSpaceUtil::getMostRecentAgentActionLink( AtomSpace& atomSpace,
 
 
 Handle AtomSpaceUtil::getMostRecentAgentActionLinkWithinTime(AtomSpace& atomSpace,
-        const std::string& agentId,
-        unsigned long t1,
-        unsigned long t2 )
+							     const std::string& agentId,
+							     unsigned long t1,
+							     unsigned long t2 )
```
ngeiswei: The indentation style of that file seems to be 4-ws.

opencog/embodiment/AtomSpaceExtensions/AtomSpaceUtil.cc
```diff
@@ -2430,8 +2423,8 @@ Handle AtomSpaceUtil::getMostRecentAgentActionLinkWithinTime(AtomSpace& atomSpac
 }
 
 Handle AtomSpaceUtil::getMostRecentAgentActionLinkAfterTime(AtomSpace& atomSpace,
-        const std::string& agentId,
-        unsigned long timestamp )
+							    const std::string& agentId,
+							    unsigned long timestamp )
```
ngeiswei: The indentation style of that file seems to be 4-ws.

opencog/embodiment/AtomSpaceExtensions/AtomSpaceUtil.h
```diff
@@ -750,8 +752,8 @@ class AtomSpaceUtil
      *         if there is no grabbed object
      */
     static Handle getHoldingObjectHandleAtTime(AtomSpace& atomSpace,
-            const std::string& holderId,
-            unsigned long time);
+					       const std::string& holderId,
+					       unsigned long time);
```
ngeiswei: Should be 4-ws.

opencog/embodiment/Learning/Filter/EntityRelevanceFilter.cc
```diff
@@ -290,7 +290,7 @@ const agent_to_actions EntityRelevanceFilter::getAgentActions(AtomSpace& as,
     //  t = Temporal(t.getLowerBound()+1, t.getUpperBound()-1);
 
     std::list<HandleTemporalPair> htp;
-    timeServer().getTimeInfo(back_inserter(htp),
+    timeServer().getTimeInfo(back_inserter(htp),
```
ngeiswei: Looks like you need to learn to use a real editor, like emacs or vim. ;-)

opencog/spacetime/SpaceServer.cc
```diff
@@ -539,39 +544,35 @@ bool SpaceServer::addSpaceInfo(Handle objectNode, Handle spaceMapHandle, bool is
     opencog::spatial::BlockVector pos(objX, objY, objZ);
     SpaceMap* theSpaceMap=spaceMaps[spaceMapHandle];
 
-    if (entityClass == "block")
-    {
+    if (entityClass == "block"){
```
ngeiswei: better put a space in-between `){`

ngeiswei:
@chenesan Thanks for the improvements. There are still some minor formatting issue (ws vs tab, trailing ws). The coding standard http://wiki.opencog.org/w/Development_standards allows both whitespace and tab (smart-tab to be precise) as long as the style is consistent within the same file. Please fix that and I'll merge. Thanks.

chenesan:
@ngeiswei Thanks for your review of the format!
I've used emacs for several months. But after this pull request I just start to learn how to set the indent and some tools to formatting the code. And I just found out that I set a non-MonoSpace font as default font long time ago So I don't see the "true" whitespace length in my code.....;-(
Because the indnet of function declaration in AtomSpaceUtil.* is messy (somewhere is 4 ws, somewhere is lined up with the first parameter..) So I mark the whole file and call the indent-region to do the indent.

I'm curious about how do you prevent the trailing whitespace automatically in emacs? That's useful since I usually forget this after delete the parameter;-(

ngeiswei:
`M-x delete-trailing-whitespace` ;-). Also I'm using magit, the diff (type d in the magit-status page) shows the trailing spaces in red.

opencog/spacetime/TimeServer.cc
```diff
@@ -107,113 +115,178 @@ octime_t TimeServer::getLatestTimestamp() const
...
-    return addTimeInfo(h, nodeName, tv);
+    return addTimeInfo(h, nodeName, timeDomain, tv);
 }
 
-Handle TimeServer::addTimeInfo(Handle h, const Temporal& t, TruthValuePtr tv)
+Handle TimeServer::addTimeInfo(Handle h, const Temporal& t, const TimeDomain& timeDomain, TruthValuePtr tv)
 {
     OC_ASSERT(atomspace->is_valid_handle(h),
-            "TimeServer::addTimeInfo: Got an invalid handle as argument\n");
+	      "TimeServer::addTimeInfo: Got an invalid handle as argument\n");
```
ngeiswei: Should be ws.

opencog/spacetime/TimeServer.cc
```diff
@@ -107,113 +115,178 @@ octime_t TimeServer::getLatestTimestamp() const
...
     for (std::list<HandleTemporalPair>::const_iterator itr = existingEntries.begin();
-            itr != existingEntries.end(); ++itr) {
-        Handle atTimeLink = getAtTimeLink(*itr);
+		 itr != existingEntries.end(); ++itr) 
+    {
+        Handle atTimeLink = getAtTimeLink(*itr, timeDomain);
         DPRINTF("Got atTimeLink = %lu\n", atTimeLink.value());
         if (atomspace->is_valid_handle(atTimeLink)) {
-            Handle timeNode = atomspace->get_outgoing(atTimeLink, 0);
+	    Handle timeNode = atomspace->get_outgoing(atTimeLink, 0);
```
ngeiswei: ws

opencog/spacetime/SpaceServer.cc
```diff
@@ -99,7 +99,10 @@ void SpaceServer::atomRemoved(AtomPtr atom)
 {
     Type type = atom->getType();
     if (classserver().isA(type, OBJECT_NODE)) {
-      removeSpaceInfo(atom->getHandle(),curSpaceMapHandle);
+        std::vector< std::string> timeDomains=timeser->getTimeDomains();
+        for(auto timeDomain: timeDomains){
```
ngeiswei: ){

opencog/spacetime/TimeServer.h
```diff
@@ -305,51 +357,64 @@ class TimeServer
...
-                                    octime_t startMoment, octime_t endMoment) const
+		OutputIterator getMapHandles( OutputIterator outIt, 
+                                              octime_t startMoment, 
+                                              octime_t endMoment,
+                                              const TimeDomain& timeDomain = DEFAULT_TIMEDOMAIN) const
     {
-        Temporal t(startMoment, endMoment);
-        std::vector<HandleTemporalPair> pairs;
+	Temporal t(startMoment, endMoment);
+	std::vector<HandleTemporalPair> pairs;
```
ngeiswei: ws

ngeiswei:
@chenesan There's still some tabs instead of ws in about ten places + minor cruft. It's not really important, I'm merging in case that is holding you back. But try to fix that in a later commit. Thanks.

chenesan:
@ngeiswei 
Thank you!

PR #1680, opened 2015-07-16
Temp

rodsol:
Add sentence-matching.scm to opencog.conf file and clean bad indentation


linas:
I'm thinking that it would be better to move the scheme file to a new directory, say /opencog/nlp/nlu  that will contain all of the new processing code. That way, the "chatbot" directory only contains the IRC interface code.   I'll try doing this and some other cleanup real soon now.

PR #1681, opened 2015-07-16
Cleanup old chatbot interfaces

linas:
Move all of the old, unused code to the old-chatbot directory.

Move the IRC-handling bits to the IRC directory.

Misc whitespace cleanup


linas:
Amen please review and merge. I think this is all harmless.

AmeBel:
thanks

PR #1682, opened 2015-07-17
forward chaining example

yantrabuddhi:
scheme example of how to use forward chaining with opencog


AmeBel:
:+1: 

suggestions.
1. it would be great to have the facts and the configuration separated to different files
2. and would be great to have the full set of rules and examples from the wikipedia article
3. edit the markdown as per https://help.github.com/articles/markdown-basics/ for better readiability

misgeatgit:
I am not sure if  this example should live in opencog or atomspace project. @ngeiswei  has recently pushed some examples related with forward chaining in the atomspace repo; perhaps this should also be moved there.

williampma:
Yeah, FC example should be in the atomspace repo.

AmeBel:
got it, the examples should be in the repository where the features are implemented. 

@yantrabuddhi you can port this to atomspace and include the suggestions there

ngeiswei:
If it was about PLN it could have been here, but I see that it includes its own rule set so it should be in the atomspace. Also I guess it should rather be in a subfolder of rule-engine, like
```
examples/rule-engine/forward-chaining
```

PR #1692, opened 2015-07-24
revert the delete of nlp modules

AmeBel:
related with https://github.com/opencog/atomspace/pull/166


AmeBel:
merging as https://github.com/opencog/atomspace/pull/166 has been accepted

PR #1701, opened 2015-08-05
Microplanning: Added imperative sentence form

williampma:
Assuming "imperative" uses the structure of using "you" as the subject, than imperative is just like declarative and can be microplanned by looking for subject-verb combo.

### DIFF
--- a/opencog/nlp/microplanning/sentence-forms.scm
+++ b/opencog/nlp/microplanning/sentence-forms.scm
@@ -42,5 +42,15 @@
 			)
 		)
 	)
+	
+	(InheritanceLink
+		(ConceptNode "ImperativeUtterance")
+		(OrLink
+			(EvaluationLink
+				(PredicateNode "MicroplanningVerbMarker")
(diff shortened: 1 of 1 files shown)


anitzkin:
Yeah, nice thought :-) but, realistically, imperatives almost never include an explicit you.  In fact, the only time people say "you" with the imperative is for special emphasis.   

However, LG and therefore Relex do detect imperatives a high proportion of the time, so you should be able to use the imperative "tense" atoms created by Relex (as an aside, 'imperative' is not a tense, but it is in relex,so . . . ).  Also note that I rigged relex to sort-of create implicit "you" subjects for imperatives, because that was needed for R2L.  So, currently, Relex assigns the subject of the imperative to be the "left wall" and calls it "you" -- and ugly hack but the best we could do without being able to add constituents to the parses given to relex by LG.  Anyway, so you should just know that the output of relex for imperatives may include this "you" subject as the referent of the left wall.

PR #1713, opened 2015-08-11
Make conceptnet converter produces Python script

kim135797531:
* Supports the output file in Scheme and Python both.
* Improve the parsing speed by using Numpy library.
* Avoid from error about link types that related with spacetime library.
* Change a README file to markup file.

Unit test result:
```
The following tests FAILED:
	 12 - AnaphoraTest (Failed)
	 14 - PLNRulesUTest (Failed)
	 15 - PythonModuleUTest (Failed)
	 19 - RestApiTest (OTHER_FAULT)
```


AmeBel:
Is there any reason not to use https://github.com/opencog/test-datasets/tree/master/conceptnet, other than we not specifying that this code isn't maintained?

AmeBel:
What is a python output? Why is it needed?

misgeatgit:
I guess dongmin didn't realize that there exists a port of conceptnet 5. @kim135797531 you can find a recently ported one [here](http://buildbot.opencog.org/downloads/) by the name cnet5.3-scm-port.tar.bz2

amirouche:
I think it's better to work with msgpack since it's [the intermediate format](https://github.com/commonsense/conceptnet5/wiki/Changelog#edges-in-msgpack-format) which will allow to save a few (a lot actually) CPU cycles if for instance you build a custom conceptnet with extra sources.

IIRC concepnet does:

```
[sources: wiktionary, wordnet, ...] -> {concepnet pipeline} -> [list of edges as .msgpack] -> [list of edges as .csv]
```

I have made [a script in python to iterate over all the files and unpack each line in lazy way an load conceptnet inside a database](https://git.framasoft.org/python-graphiti-love-story/AjguGraphDB/blob/master/example/conceptnet/load.py).

kim135797531:
Hi @AmeBel, 'python output' is a script file includes statements with python syntax. Like 'scheme output' file, it adds ConceptNet atoms to AtomSpace.
* Python binding can evaluate scheme script by `scheme_eval()`, but it needs scheme binding obviously and needs more statements for make proper scheme environment.
* Also I'm not sure the reason but when I try to load with scheme ConceptNet script in python, it tooks very long time to complete(tried compile & non-compile mode both). Debugging scheme code in python is also painful.
* So if we have python ConceptNet script file, user that uses OpenCog python binding will import ConceptNet in easily.

Next, I removed some types such as `RetroactiveImplicationLink` because it seems removed a long time ago. But some types are still difficult to use because of current implementation in python binding.
* Currently python binding can't add new custom type automatically after initialize types variable. If user want to use `DuringLink` type, then user must import `spacetime` library before import `types` in atomspace library. It seems weird so maybe user will meet error `Can't find DuringLink~~` when converting or using ConceptNet in python.
* In fact, we can load current script(has type defined in spacetime) in python by calling scheme function... It is not only strange but also dangerous because some python program will encounter the error when they met the link has unknown type(65535).

@misgeatgit, You are right. Now changed code can control ConceptNet 5. But I didn't know about `cnet5.3-scm-port.tar.bz2` file, is there exists information about the contents of http://buildbot.opencog.org?

@amirouche, Thank you for advice and great script. I'll try them after finish my work.

amirouche:
@kim135797531 I'm wondering why concepnet is not loaded directly in the database instead of going through "atomspace interpreter". Would it be faster (and better) to load concepnet in the backend programmatically?

amirouche:
IIUC you rewrite the csv file in scheme and python to be later imported inside the atomspace. This is not efficient.

linas:
Writing anything directly into the SQL database is discouraged, mostly because you will like make multiple errors in the required structure.  Also, the SQL format will change (we have to add the GTV someday soon). The variant for mysql if/when it is created, will be different (due to limitations in the mariadb types). The neo4j format is not even SQL. It would be foolish to write a direct-to-neoj importer.

Anyway, load speed is not very important: you do it once, you are done. Far far far more important is speed of PLN when it operates on the thing.

PR #1715, opened 2015-08-12
Graphing tools written in Python

rohit12:
This pull request contains three files that I had written to help draw graphs. The libraries that would be needed to draw graphs are numpy and matplotlib. The files are written in a fairly generic manner. However, they are hardcoded. To extend them, the requisite variable values, like pointers to files, etc would need to be changed.


AmeBel:
I think it would be best to have this in https://github.com/opencog/external-tools, but I will live the decision to @linas

linas:
Rohit,  Can you add a README that describes how to use this?

Amen,  These tools are very specific to this project, and the particular data format that it uses.  They are not in any way general. Maybe they could someday be generalized, but right now, its more important to continue the experiment, than to pollish some ad-hoc graphing tools.

rohit12:
Yes, I'll add a README soon. In fact, I have added comments in the files
themselves. I'll still add a README though.

PR #1716, opened 2015-08-12
SuReal updates

leungmanhin:
Wrote C++ code for getting LG connectors instead of calling the word-inst-get-source-conn scheme function. Also added a map to store the disjuncts of the nodes in the pattern.

### DIFF
--- a/opencog/nlp/sureal/SuRealPMCB.cc
+++ b/opencog/nlp/sureal/SuRealPMCB.cc
@@ -139,6 +139,7 @@ bool SuRealPMCB::variable_match(const Handle &hPat, const Handle &hSoln)
         HandleSeq qN1 = get_neighbors(hLeftWordInst1, false, true, WORD_SEQUENCE_LINK);
         HandleSeq qN2 = get_neighbors(hLeftWordInst2, false, true, WORD_SEQUENCE_LINK);
 
+        // get the NumberNode and compare their word sequences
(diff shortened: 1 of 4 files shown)


opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -99,20 +99,123 @@ bool SuRealPMCB::variable_match(const Handle &hPat, const Handle &hSoln)
...
+
+    // helper for checking if a ListLink has a LgLinkInstanceNode as a neighbor
+    auto hasNoLgLinkInstanceNode = [](Handle& h)
+    {
+        HandleSeq qN = get_neighbors(h, true, false, EVALUATION_LINK);
+        return not std::any_of(qN.begin(), qN.end(), [](Handle& hn) { return hn->getType() == LG_LINK_INSTANCE_NODE; });
+    };
+
+    // remove any ListLink that doesn't have a LgLinkInstanceNode as a neighbor
+    qSolnListLinks.erase(std::remove_if(qSolnListLinks.begin(), qSolnListLinks.end(), hasNoLgLinkInstanceNode), qSolnListLinks.end());
```
williampma: Is this doing the same as https://github.com/opencog/atomspace/blob/master/opencog/atomutils/AtomUtils.h#L131

williampma:
If `word-inst-get-source-conn` is no longer used anywhere else, might be better to remove it.

leungmanhin:
Yes you are right. I've updated it to use the get_predicates now.

PR #1722, opened 2015-08-14
Added pln inheritance rules

prateeksaxena2809:


opencog/reasoning/pln/rules/modus-ponens-rule.scm
```diff
@@ -0,0 +1,50 @@
...
+    (let
+        ((sA (cog-stv-strength A))
+         (cA (cog-stv-confidence A))
+         (sAB (cog-stv-strength AB))
+         (cAB (cog-stv-confidence AB))
+         (snotAB 0.2)
+         (cnotAB 1))
+        (cond
+            [(and
+                (= (gar AB) A))
```
williampma: Humm... I see you are going the path of hiding even more the rule's output inside Scheme function.  It's not possible to tell that $A need to be inside the link $C just from the rule's definition anymore...

I know you are working around the limitation of the Pattern Matcher, where one cannot specific additional matching criteria within a link type VariableNode.  I wondering if we need a VariableLink... or some better way to define patterns.

williampma:
Are the `abduction-rule.scm`, `deduction-rule.scm`, etc. supposed to be separate from the existing `abduction.scm`, `deduction.scm`, etc.?

opencog/reasoning/pln/rules/modus-ponens-rule.scm
williampma: However, in this case since there're only 3 types... maybe you could just use an `OrLink`.

bgoertzel:
Conceptually, what is the difference between e.g. deduction.scm and
deduction-rule.scm supposed to be?

ben

On Mon, Aug 17, 2015 at 12:19 PM, William Ma <notifications@github.com>
wrote:

> Are the abduction-rule.scm, deduction-rule.scm, etc. supposed to be
> separate from the existing abduction.scm, deduction.scm, etc.?
>
> —
> <https://github.com/opencog/opencog/pull/1722#issuecomment-131675328>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

prateeksaxena2809:
deduction-rule.scm, abduction-rule.scm are a general rules which works on all links. But since, the backward chaining creates issues in the links created outside the bind link, i thought of retaining the old rules in case it is needed for implementation in some of the syllogisms. Should I remove them?

bgoertzel:
Yes, it's best to retain the old rules for now, since I think the new ones
are still problematic

It's not going to work, really, to have the output of the rule specified
only deep inside some Scheme function, as is done in the new versions
(looking at the new modus ponens rule as William noted)

The only way to see that the output of the rule is B is to look at the
Scheme function and the "gdr" there, but the backward chainer can't do
that...   Somehow we need the input of the rule to look like

A
LinkType A B

not just like

A
C

generically... and somehow we need the output to look like

B <tv>

with the tv set by the pln-formula-modus-ponens,

rather than the output Atom *and* its truth value being set inside the
pln-formula-modus-ponens Schema file...

In the current form the rule is only useful for forward chaining and not
for backward chaining.  This may be useful for short-term experimentation
with forward chaining but is definitely not a correct design for OpenCog
going forward...

William, you know the current limitations of the PM.   Is there really no
way to do this in a backward chainer friendly way using the PM in its
current form?   If not obviously  we need to either

-- extend the PM appropriately

or

-- introduce deep types

ASAP so we can do this stuff right...

thx
ben

On Mon, Aug 17, 2015 at 2:57 PM, Prateek Saxena <notifications@github.com>
wrote:

> deduction-rule.scm, abduction-rule.scm are a general rules which works on
> all links. But since, the backward chaining creates issues in the links
> created outside the bind link, i thought of retaining the old rules in case
> it is needed for implementation in some of the syllogisms. Should I remove
> them?
>
> —
> <https://github.com/opencog/opencog/pull/1722#issuecomment-131699539>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

williampma:
Wait, I am confused.  So is this new set of rules meant to be not used by the BackwardChainer, and that's why there's both `deduction.scm` and `deduction-rule.scm`.  If so then everything is fine (though I imagine PM will be quite slow for query like

```
        (AndLink
            (VariableNode "$A")
            (VariableNode "$B")
            (VariableNode "$C")
            (VariableNode "$D")
            (VariableNode "$E"))
```

even if some of the VariableNodes are typed).

On the other hand, if you want to make also BC friendly.... ah, I see there's really no way to do that in one rule.

Even if 
```
            (TypedVariableLink
                (VariableNode "$C")
                (TypeChoice
                    (TypeNode "InheritanceLink")
                    (TypeNode "ImplicationLink")
                    (TypeNode "SubsetLink"))))
```
is pulled out of the VariableList and the following is added to the pattern
```
(OrLink
   (InheritanceLink $A $B)
   (ImplicationLink $A $B)
   (SubsetLink $A $B)
)
```
There's no way to know which of the link in the OrLink is actually used in the Pattern Matcher output part... and that is needed to get the TV of the link to calculate the TV of B.

Not in one rule anyway...  the only way around it, it seems, is like what we did with evaluation-to-member-rule, and split this into 3 rules, one of each link type.... :(

bgoertzel:
Splitting into 3 rules is not a disaster, for now...

On Mon, Aug 17, 2015 at 3:54 PM, William Ma <notifications@github.com>
wrote:

> Wait, I am confused. So is this new set of rules meant to be not used by
> the BackwardChainer, and that's why there's both deduction.scm and
> deduction-rule.scm. If so then everything is fine (though I imagine PM
> will be quite slow for query like
>
>         (AndLink
>             (VariableNode "$A")
>             (VariableNode "$B")
>             (VariableNode "$C")
>             (VariableNode "$D")
>             (VariableNode "$E"))
>
> even if some of the VariableNodes are typed).
>
> On the other hand, if you want to make also BC friendly.... ah, I see
> there's really no way to do that in one rule.
>
> Even if
>
>             (TypedVariableLink
>                 (VariableNode "$C")
>                 (TypeChoice
>                     (TypeNode "InheritanceLink")
>                     (TypeNode "ImplicationLink")
>                     (TypeNode "SubsetLink"))))
>
> is pulled out of the VariableList and the following is added to the pattern
>
> (OrLink
>    (InheritanceLink $A $B)
>    (ImplicationLink $A $B)
>    (SubsetLink $A $B)
> )
>
> There's no way to know which of the link in the OrLink is actually used in
> the Pattern Matcher output part... and that is needed to get the TV of the
> link to calculate the TV of B.
>
> Not in one rule anyway... the only way around it, it seems, is like what
> we did with evaluation-to-member-rule, and split this into 3 rules, one of
> each link type.... :(
>
> —
> <https://github.com/opencog/opencog/pull/1722#issuecomment-131717821>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

prateeksaxena2809:
So, for now should I split all the rules with pattern matching issues?

bgoertzel:
I think that's best ..

On Mon, Aug 17, 2015 at 5:16 PM, Prateek Saxena <notifications@github.com>
wrote:

> So, for now should I split all the rules with pattern matching issues?
>
> —
> <https://github.com/opencog/opencog/pull/1722#issuecomment-131740421>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

linas:
Slightly off-topic: William wonders:
> I know you are working around the limitation of the Pattern Matcher, where one cannot specific additional matching criteria within a link type VariableNode.

It is worth noting that the pattern described in https://github.com/opencog/atomspace/issues/211 does work to bind a variable to a link and at the same time, bind variables to components inside of it.  It won't be very efficient, in the current implementation, but it shoudl work. (well, it does work in BinDlink form, but is oddly broken in GetLink form, thus the bug report  https://github.com/opencog/atomspace/issues/211 )

PR #1723, opened 2015-08-14
Added predicate rule

prateeksaxena2809:


opencog/reasoning/pln/rules/evaluation-implication-rule.scm
```diff
@@ -0,0 +1,65 @@
...
+                (cog-stv-strength A)
+                (cog-stv-strength C)
+                (cog-stv-strength AB)
+                (cog-stv-strength AC))
+            (min
+                (cog-stv-confidence B)
+                (cog-stv-confidence A)
+                (cog-stv-confidence C)
+                (cog-stv-confidence AC)
+                (* 0.9 (cog-stv-confidence AB))))))
```
williampma: I think there are multiple discount factors, not just one.  Looking at the old python code it seems to be

`evalCB = 0.9 * 0.9 * (min (0.9 * evalAB), implAC, A, C)`

According to the comment on the python code (perhaps you should add that comment to here), it is from Chapter 10: Higher Order Extensional Inference.  It seems it is online at http://download.springer.com/static/pdf/392/chp%253A10.1007%252F978-0-387-76872-4_10.pdf?originUrl=http%3A%2F%2Flink.springer.com%2Fchapter%2F10.1007%2F978-0-387-76872-4_10&token2=exp=1439788690~acl=%2Fstatic%2Fpdf%2F392%2Fchp%25253A10.1007%25252F978-0-387-76872-4_10.pdf%3ForiginUrl%3Dhttp%253A%252F%252Flink.springer.com%252Fchapter%252F10.1007%252F978-0-387-76872-4_10*~hmac=ca35e8abf0f639b73dccf63f69e37962fd7bd29ab0d15e8291d6684ebb45fa7f and it also mentions the multiple discounts.

prateeksaxena2809: I can see why one 0.9 should be added as a discount factor from [the old pln formula code](https://github.com/opencog/opencog/blob/master/opencog/python/pln_old/rules/formulas.py#L557). I am not sure about the other 0.9 . I will add the comments from the old pln to the code. Should I do the same for other codes?

williampma: If the comment are useful.

One of the 0.9 I see is from https://github.com/opencog/opencog/blob/master/opencog/python/pln_old/rules/formulas.py#L555
and another is from
https://github.com/opencog/opencog/blob/master/opencog/python/pln_old/rules/formulas.py#L554

opencog/reasoning/pln/rules/evaluation-implication-rule.scm
```diff
@@ -0,0 +1,110 @@
...
+                (cog-stv-strength AC))
+            (*
+                0.9
+                (min
+                    (cog-stv-confidence B)
+                    (cog-stv-confidence A)
+                    (cog-stv-confidence C)
+                    (cog-stv-confidence AC)
+                    (* 0.9 (cog-stv-confidence AB)))))))
+
```
williampma: There should be 3 discount factors right?  

mem2InhFormula has one.
deductionIndependenceBasedFormula has one over the min( )
inh2MemFormula, which is just mem2InhFormula, has one again.

prateeksaxena2809: Yes, sorry my bad.

williampma:
Thanks!

PR #1724, opened 2015-08-15
Code to create Decoded Disjuncts

rohit12:
This PR contains two files. The DecodedConnector type has been added to the atom_types.script file.
The grammar.scm file contains code to create decoded disjuncts as well as creating those disjuncts for all connections for a word.


AmeBel:
@linas please reveiw

linas:
Hi Rohit,

I just expanded, slightly, on the documentation here. All of the routines are missing a clear indication of what structures they take as input, and what structures they generate as output. This makes the code very difficult to understand.  Could you add this?  Something like this:
```
; For example, this function expects as input 
;
;     MSTBlahLink
;         MSTFooNode "asdf"
;
;  Given the above, it returns as a result
;
;   MSTConnectorLink
;       LgDirectionLink "+"
;       MSTFooNode "asdf"
```

rohit12:
Yes, I'll add the documentation.

PR #1725, opened 2015-08-17
minor fix

rodsol:


anitzkin:
Hmm. Probably this fix is correct, but it occurs to me, it is possible that
I coded it this way because in a "which indirect object question" the
indirect object must be represented as the object of the preposition "to",
a predicate.  I would just check the relex output for the sentence, "To
which department did you send the email?"  just to check that you don't get
_iobj(send, to) -- in case the scheme helper was not in error.

best,
Aaron

PR #1726, opened 2015-08-18
aiml rule converter

yantrabuddhi:


yantrabuddhi:
aiml converter for opencog , missing parentheses for scheme input

PR #1732, opened 2015-08-19
Move the disjuncts matching to clause match for SuReal

leungmanhin:


opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -97,212 +86,268 @@ bool SuRealPMCB::variable_match(const Handle &hPat, const Handle &hSoln)
...
+        Handle hSolnWordInst = m_as->get_handle(WORD_INSTANCE_NODE, sSoln);
 
-    // get the LG connectors for those in the qLGInstsLeft
-    for (Handle& hEvalLink : qLGInstsLeft)
-    {
-        const Handle& hLinkInstNode = LinkCast(hEvalLink)->getOutgoingSet()[0];
+        // sometimes when the pattern node is not a variable, it may match to
+        // a solution node that does not have a corresponding WordInstanceNode,
+        // so we should accept it here, e.g. (PredicateNode "possession")
+        if (hSolnWordInst == Handle::UNDEFINED)
```
williampma: Almost thought this is wrong.  Can you add an explanation saying something like "if the node is a variable, it would have matched to a node with corresponding WordInstanceNode by variable match", and "if the node is not a variable, then the standard node_match would have made it matched to itself".

leungmanhin:
Oh yes, I've updated the comment, please have a look, thanks!

williampma:
Thanks!

PR #1735, opened 2015-08-21
Octomap feature PR On the 3DSpaceMap branch

chenesan:
Content is as same as the previous PR but on the 3DSpaceMap feature branch.


chenesan:
I'll update the doc and fix the formats in the next few days, and then you can see if the code is good enough to be merged in the feature branch.

AmeBel:
Sure.

When you guys ( @LucidBlue @chenesan ) are ready https://github.com/OC2MC/opencog-to-minecraft ownership could be transferred to opencog. Thanks :smile:

linas:
I strongly suggest merging this into master, and not leaving hanging in some "feature branch".  No one else is using the space server code right now, anyway, so no harm in breaking anything there.  

I don't want to merge myself, since I am a few steps too far away from the action, though.

bgoertzel:
If that is OK for you, it's OK for me --- the only other likely user of the
space-server in the near future is Hanson Robotics or Mandeep (for the Poly
U robot project) ... so if you and Mandeep don't mind the space-server
being broken for a little while, we can merge this stuff...

On Thu, Aug 27, 2015 at 11:23 AM, Linas Vepštas <notifications@github.com>
wrote:

> I strongly suggest merging this into master, and not leaving hanging in
> some "feature branch". No one else is using the space server code right
> now, anyway, so no harm in breaking anything there.
>
> I don't want to merge myself, since I am a few steps too far away from the
> action, though.
>
> —
> <https://github.com/opencog/opencog/pull/1735#issuecomment-135273730>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

linas:
OK, mandeep, are you there?

linas:
I just realized that this is a merge into a branch called 3Dspacemap. So that is a harmless merge.  We will next need to merge that branch into master...

AmeBel:
@yantrabuddhi your suggestion is needed.

bgoertzel:
I guess mandeep will not have time to do anything with the 3dspacemap till
October at least ... As of now he doesn't know that code much ... I would
say this is OK to merge as there are really no other users of the current
3dspacemap code ... And we know we want to replace the current code with
the new octomap version for general usage ...

bgoertzel:
Oh ... I just noticed the other thread on this ...

I am en route to the airport to fly to Kazakhstan and will likely be
offline for the next 20 hrs or so

Based on the other thread I now see there are complexities here，so my
current opinion on merging thus code is that I have no opinion till I
emerge online in KZ and can read all the emails in the other thread
carefully ...

I'm no github or process guru so I should leave those aspects to others,
but about this code I do have knowledge ... I don't really want people
building more code that relies on the old existing spacemap code that's now
in master; it is better for any new development effort to go into
completing and then using the new octomap based spacemap code

Ben

pennachin:
@bgoertzel My understanding is that this code is known to break stuff in master. Let me know if I'm wrong. But if I'm right, the breakage should be fixed before a merge. This is how one gets to a stable master, which we all seemed to agree is a worthwhile goal. 

Is anyone actually building more code that relies on the old spacemap now? If yes, then this merge would probably break someone else's work. If no, then why is this merge urgent enough to be done before issues are fixed?

bgoertzel:
I agree we want a stable master branch without anything broken in it ...

Since this seems controversial and not urgent and I'm about to board a long
flight, I suggest we let it sit till I emerge in Astana and can read all
the threads carefully ;p

Ben

pennachin:
Ok with me ;-)

chenesan:
Hi,

About this PR..Currently I'm still refactoring the spaceMap, to seperate the functionality of Octree3DMapManager to the OctomapOcTree(for saving block), EntityManager(for saving Entity, I'm working on this)and SpaceMapUtil(for calculation). I've make sure that the spaceMap, Space/Time server and cython binding can be built in each commit. But indeed the change of spaceMap will break the old embodiment and any other part using the spaceMap.

I'm wondering if we want to save the old embodiment code. If so, which specific parts we want to save or abandon?

bgoertzel:
I got online for a few minutes, in Beijing airport en route to Urumqi en
route to Astana ;p ...

 I took a little time to review this situation now, by reference to my
long-term memory ;p  ... rather than just replying hastily like I did
before…

Here are some comments, some for Yi-Shan and some for others' general
enlightenment...

0)
The OctomapOctree is definitely what we want for the next phases of OpenCog
development.  Octomap is good 3rd party code, it’s thread-safe, it’s easy
to customize — and Yi-Shan and Brad have been using it for their
OpenCog/Minecraft work…

1)
The major TODOs that I know about, that would need to be fixed to replace
the existing octree with the OctoMapOctree without losing any valuable
functionality inside Opencog, are:

— Remove Octree3DMapManager as Yi-Shan ahs suggested, replacing calls to it
with direct calls to the OctomapOctree

— Modify the code for assessing spatial relations between objects, to make
it work with OctomapOctree [this is called the Inquery server or something
like that, my bandwidth won't let me look at the code online right now and
I don't have it on my laptop]....

2)
There is some other code that worked with the previous 3DSpaceMap that we
probably DON'T need to keep around … mainly, as I recall:

- The code that identifies persistent entities in the Octomap (entities
composed of multiple blocks; e.g. identify that a bunch of adjacent wood
blocks are all the same entity)

This functionality is important but the existing implementation is not
scalable and will slow down tremendously when new blocks are added to an
even moderately large map.

3)
The main code I know of that uses the old 3DSpaceMap code is Shujing’s
planner.   I consider this planner “deprecated” because I don’t like the
way it uses its own rules coded in C++, inside its own rule engine, rather
than having its rules coded as cognitive content in the Atomspace.
However, it does work and has powered some demos and had a paper published
on it.   Viewed as a prototype to explore the integration of navigation and
more general planning, it serves a purpose..

We will build a new planner based on the URE and PLN, and incorporating the
insights learned from Shujing’s planning work. But we need to get PLN to
work OK on moderately complex non-temporal inferences first.

…

So I would guess if one simply merged the new OctomapOctree code now, it
would break Shujing’s planner, which is not desirable…

Off the top of my head, then, what is needed to merge the code without any
controversy or meaningful breakage would be

-- calculation of spatial relations between blocks and entities in the
OctoMap to still work

-- Shujing's planner to still work [now, fully validating that it still
works would require more testing than Yi-Shan can be expected to do....
But at least, all ocPlanner-related unit tests should still work...]

...

Given that nobody is really using the old embodiment (Shujing is not
currently using her planner either) and we want to replace it, I think the
above should be good enough... others may speak up if I seem to be leaving
something out...

-- Ben



On Thu, Aug 27, 2015 at 6:59 PM, Yi-Shan, Chen <notifications@github.com>
wrote:

> Hi,
>
> About this PR..Currently I'm still refactoring the spaceMap, to seperate
> the functionality of Octree3DMapManager to the OctomapOcTree(for saving
> block), EntityManager(for saving Entity, I'm working on this)and
> SpaceMapUtil(for calculation). I've make sure that the spaceMap, Space/Time
> server and cython binding can be built in each commit. But indeed the
> change of spaceMap will break the old embodiment and any other part using
> the spaceMap.
>
> I'm wondering if we want to save the old embodiment code. If so, which
> specific parts we want to save or abandon?
>
> —
> <https://github.com/opencog/opencog/pull/1735#issuecomment-135382860>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

chenesan:
Thanks Ben, this looks clearer.
I still have some subtle problems, though:
(1) Currently we only can get const spaceMap in SpaceServer, and we must write on it throught the SpaceServer interfaces. I know that it's good for encapsulation, but it's also make it inconvenient to use the Octomap. Once we want any new writing operation on Octomap we have to write a corresponding functions on SpaceServer(ex. laserScan/update occupancy of block...). I'm not sure if we should make user of SpaceServer able to write on the spaceMap or just write functions in SpaceServer.
(2) So I see that we need to fix the OCPlanner. But what about other parts under embodiment/, ex. Learning/Filter, Learning/LearningServer?

bgoertzel:
Hi Yi-Shan,

> (1) Currently we only can get const spaceMap in SpaceServer, and we must
> write on it throught the SpaceServer interfaces. I know that it's good for
> encapsulation, but it's also make it inconvenient to use the Octomap. Once
> we want any new writing operation on Octomap we have to write a
> corresponding functions on SpaceServer(ex. laserScan/update occupancy of
> block...). I'm not sure if we should make user of SpaceServer able to write
> on the spaceMap or just write functions in SpaceServer.
>

I am not the C++ guru, so I think I will ask Nil to answer this one...

I guess that, in theory, the SpaceServer interface could provide an
abstraction layer that lets the Atomspace call spatially-related functions
in a way that doesn't depend on the details of the Octomap.  Maybe that
encapsulation would make it easier to replace Octomap with something better
later on...?  But whether this theoretical benefit actually means anything
in reality I'm not sure and would rather let Nil opine on ;-)



> (2) So I see that we need to fix the OCPlanner. But what about other parts
> under embodiment/, ex. Learning/Filter, Learning/LearningServer?
>
>
>
Again I must defer to the great Dr. Nil Geisweiller...

As I recall, that code was broken by Shujing a long time ago.   Then Nil
was spending some time trying to get it working again, and I think he
*almost* succeeded and then got distracted by other things....

We definitely do NOT want to create any more broken unit tests in the
master branch -- we need to get a master branch where all unit tests
pass....  But, disabling unit tests in code that basically doesn't work
anyway, seems more of a judgment call....   I will leave this call to Nil
and Cassio, though (as Linas is super-busy with OpenCog/Hanson-Robotics
stuff just now)

Also, Nil may be able to quickly fix that code insofar as it needs to be
fixed.  He knows that code better than anyone I guess.   Whereas only
Shujing knows the ocPlanner....

-- Ben

PR #1737, opened 2015-08-22
add bindlink svio question

rodsol:
Who told you  that  information?

```
BindLink
      VariableNode $x
       EvaluationLink
            PredicateNode told@123
            ListLink
                VariableNode $x
                 ConceptNode information@345
                 ConceptNode you@678 
```

### DIFF
--- a/opencog/nlp/relex2logic/rule-helpers.scm
+++ b/opencog/nlp/relex2logic/rule-helpers.scm
@@ -177,12 +177,14 @@
 				(ImplicationLink (PredicateNode verb_instance) (PredicateNode verb))
 				(InheritanceLink (ConceptNode obj_instance) (ConceptNode obj_concept))
 				(InheritanceLink (ConceptNode iobj_instance) (ConceptNode iobj_concept))
-				(EvaluationLink
-					(PredicateNode verb_instance)
-					(ListLink
-						(VariableNode var_name)
-						(ConceptNode obj_instance)
-						(ConceptNode iobj_instance)
+				(BindLink (VariableNode var_name)
+					(EvaluationLink
+						(PredicateNode verb_instance)
+						(ListLink
+							(VariableNode var_name)
+							(ConceptNode obj_instance)
+							(ConceptNode iobj_instance)
+						)
 					)
 				)
 			)
@@ -193,12 +195,14 @@
 					(ImplicationLink (PredicateNode verb_instance) (PredicateNode verb))
 					(InheritanceLink (ConceptNode subj_instance) (ConceptNode subj_concept))
 					(InheritanceLink (ConceptNode iobj_instance) (ConceptNode iobj_concept))
-					(EvaluationLink
-						(PredicateNode verb_instance)
-						(ListLink
-							(ConceptNode subj_instance)
-							(VariableNode var_name)
-							(ConceptNode iobj_instance)
+					(BindLink (VariableNode var_name)
+						(EvaluationLink
+							(PredicateNode verb_instance)
+							(ListLink
+								(ConceptNode subj_instance)
+								(VariableNode var_name)
+								(ConceptNode iobj_instance)
+							)
 						)
 					)
 				)
@@ -210,12 +214,14 @@
 					(ImplicationLink (PredicateNode verb_instance) (PredicateNode verb))
 					(InheritanceLink (ConceptNode subj_instance) (ConceptNode subj_concept))
 					(InheritanceLink (ConceptNode obj_instance) (ConceptNode obj_concept))
-					(EvaluationLink
-						(PredicateNode verb_instance)
-						(ListLink
-							(ConceptNode subj_instance)
-							(ConceptNode obj_instance)
-							(VariableNode var_name)
+					(BindLink (VariableNode var_name)
+						(EvaluationLink
+							(PredicateNode verb_instance)
+							(ListLink
+								(ConceptNode subj_instance)
+								(ConceptNode obj_instance)
+								(VariableNode var_name)
+							)
 						)
 					)
 				)


linas:
I'm not sure, but given how you wrote this, I think you want to use GetLink here, not BindLink.  The getLink will simply return a set ov values for your variable.   See the wiki for details: http://wiki.opencog.org/w/GetLink

bgoertzel:
Rodi --

Yah, Linas is right.   I did not think of that because GetLink is new.  But
since there is now GetLink, we may as well use it instead of BindLink...

Sorry for the confusion ;p ;) ... basically you are writing programs in a
language ["Atomese"] that is being created as you program ;D

ben

On Mon, Aug 24, 2015 at 3:05 AM, Linas Vepštas <notifications@github.com>
wrote:

> I'm not sure, but given how you wrote this, I think you want to use
> GetLink here, not BindLink. The getLink will simply return a set ov values
> for your variable. See the wiki for details:
> http://wiki.opencog.org/w/GetLink
>
> —
> <https://github.com/opencog/opencog/pull/1737#issuecomment-133898754>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

AmeBel:
@rodsol why would you want a Get/Bind link as a an output of a helper function?

bgoertzel:
Amen,

Because we are moving toward a convention that all variables in the
Atomspace must be bound (to *some* link)

The output of a R2L rule that represents a question must have VariableNodes
in it, and so these VariableNodes must be bound somewhere or other....

-- Ben

On Mon, Aug 24, 2015 at 11:52 AM, Amen Belayneh <notifications@github.com>
wrote:

> @rodsol <https://github.com/rodsol> why would you want a Get/Bind link as
> a an output of a helper function?
>
> —
> <https://github.com/opencog/opencog/pull/1737#issuecomment-134012110>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

AmeBel:
ok, so how will PLN deal with this?

bgoertzel:
The BC can deal with this without too much trouble, it just has to strip
away the GetLink when it takes the Atom-construct as a target... we already
discussed w/ William in some other github thread as I recall..

On Mon, Aug 24, 2015 at 12:05 PM, Amen Belayneh <notifications@github.com>
wrote:

> ok, so how will PLN deal with this?
>
> —
> <https://github.com/opencog/opencog/pull/1737#issuecomment-134022294>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

linas:
If you meremly want to bind variables (scope them so that they don't leak out, and/or define exactly what the interface is), then you can use LambdaLink  http://wiki.opencog.org/w/LambdaLink I added C++ code for this just a few days ago.  So, for example, Getlink inherits from lambdaLink, and has the semantics "search for this pattern" -- other kinds of lambda can have other semantics.

bgoertzel:
Rodi: a practical suggestion occurs at the end of this long discursive
email ;)

Well Linas, the issue here is *what should be the output of the NL
comprehension pipeline when it interprets a question*?

The logic of questions says that a question can be equated to the set of
all statements that correctly answer the question.   (We can think about
this as a fuzzy set if we want to...)....   We don't have a link type that
means exactly this, but I'm not sure we need to...

GetLink makes some sense to because when executed it will try to find the
answer to the question....   So for instance, conceptually, if we wanted to
pattern match

**
$X really loves (Who tickled the elephant?)
***

the GetLink solution is basically interpreting this as

**
$X really loves (the result of get $Y so that: $Y tickled the elephant)
***

On the other hand, the SatisfyingSet solution is interpreting this as

**
$X really loves (the set of $Y so that: $Y tickled the elephant)
***

... the difference being that in the former case, the PM will actually try
to figure out who tickled the elephant, and in the latter case, the PM will
only return the set of people who are already known to have tickled the
elephant...

The LambdaLink solution is more like

**
$X really loves ($Y so that: $Y tickled the elephant)
***

without specifying whether $X loves one, many or all of the entities that
tickled the elephnt...

I would say that different NLP questions can have different senses ---
which would be an argument for a minimalist solution like LambdaLink

On the other hand, SatisfyingSetLink is basically equivalent to the formal
logic of questions, which is nice...

RODAS: a simple practical suggestion... since we are kinda confused about
what link type to use here, I suggest that you make a single Scheme
function such as

get-question-quantifier-type()

You can initially make it  return LambdaLink, and then if/when we want to
change that later on, we can just change the link type in one place.   In
general we should try to avoid cut and pasted code in the different R2L
functions (Aaron cut and pasted a lot, rather than abstracting things into
functions).....   So if we want to change LambdaLink to GetLink later,
there should just be one line of code to change, rather than having to go
through and change it in a lot of different functions...

-- Ben





On Tue, Aug 25, 2015 at 5:22 AM, Linas Vepštas <notifications@github.com>
wrote:

> If you meremly want to bind variables (scope them so that they don't leak
> out, and/or define exactly what the interface is), then you can use
> LambdaLink http://wiki.opencog.org/w/LambdaLink I added C++ code for this
> just a few days ago. So, for example, Getlink inherits from lambdaLink, and
> has the semantics "search for this pattern" -- other kinds of lambda can
> have other semantics.
>
> —
> <https://github.com/opencog/opencog/pull/1737#issuecomment-134384423>.
>



-- 
Ben Goertzel, PhD
http://goertzel.org

linas:
Sorry, did not mean to confuse the issue. For the question-answering example, using GetLink is fine, I guess.  I was merely reacting to an earlier remark of "all variables must be bound" and LambdaLink is a non-descript variable-binder.  It seems like it might not be appropriate here. I can't say.

AmeBel:
http://wiki.opencog.org/wikihome/index.php/Link_comparison helped me see the difference.

PR #1743, opened 2015-08-27
merge new 3d space map into master

linas:
This is part 2 of the merge discussion... merging the 3D spacemap into master. @yantrabuddhi you are the only one likely to be affected by this; any comments?


pennachin:
@linas the pull request you just merged didn't build under CircleCI. This really needs to be fixed (and it might be a CircleCI thing, let's see what happens to this pull request) before merging.

I think merging code with known issues onto master is a bad practice that should be avoided. I don't see the benefit of doing so in this case, why not keep the code in a feature branch until known issues are fixed?

pennachin:
CircleCI failed to build this PR as well...

linas:
But nothing ever builds on circle-ci; it hasn't starting working yet.... are you seeing something I'm not?

pennachin:
The master head builds and then CircleCI fails it because the unit tests fail, see:

https://circleci.com/gh/opencog/opencog/99

So CircleCI is working, and producing the results I get on my machine. For this PR (and the one you just merged into this branch) it fails to complete the build.

pennachin:
Why did you merge this? This code doesn't build, and now master doesn't build.

linas:
Cassio, 

1) if you have a problem building, open a bug report. Don't just revert. You can't try to take a high road about process, and then completely disreard the process when it suits you.  What you did was wrong. Just open a bug report.

2) We need to dump CircleC. You seem to strongly want some kind of status report --  find some kind of system that actually works. CircleCI is deeply and fundamentally flawed -- basically, it is broken in a large variety of ways.  I've lived with it for months, and every day, my hatred of the thing deepens.  It is a total piece of SHIT.

pennachin:
Linas,

This code didn't build for me, for Roman, and on CircleCI. It doesn't build on a fresh clone and building from scratch. This isn't *me* having a problem building. This is code that may compile on your local machine but doesn't compile on a fresh, standard build. 

I'm not disregarding the process. The process is "master should be stable". It wasn't, so now it's at least back to where it was before the merges. Whatever flaws CircleCI may have, "it builds on my machine hence it should be on master" is a worse idea. 

If you want to discuss CircleCI, happy to do it on the mailing list, where Nil has started a thread.

rTreutlein:
Actually it build fine on my Machine, its just that it now has ROS as a dependency.

PR #1744, opened 2015-08-27
Support load pln-config.scm from any location

ngeiswei:
Loading only worked when doing (load "pln-config.scm") from the current directory. Now you can do (load "../pln-config.scm"), etc.


williampma:
Thanks!

PR #1749, opened 2015-08-31
Seperate Entity functions in SpaceMap to EntityManager

chenesan:
The works is move the entity handling functions
* add/remove entity
* update entity location
to another class EntityManager. Because they're not related to the octree data structure.
After this change I have removed the Octree3DMapManager.* in the code base.

This change also affect the SpaceServer. So I change the HandleToSpaceMaps to 
map<Handle, pair<SpaceMap*, EntityManager*> >

I have some problems about the code. I need some advices about them:
(1)
The EntityManager makes SpaceServer a little messy. Since we have to handle two objects for each scene.
I'm thinking if it's better to make a Scene class, which represent the scene and own the SpaceMap and EntityManager, to replace the pair<SpaceMap*, EntityManager*>. 
Also we can move some environment parameter, ex. agent height, map(scene) name in the Scene object.
(2)
In the old Octree3DMapManager we count the number of unit blocks in the OcTree by mTotalUnitBlockNum. We handle this member in the add/removeSolidUnitBlock. But now we move the functionality to the OctomapOcTree. It's obvious that if someone add a block through the updateNode() + setBlock() function, the number of unit block will be wrong. 
I'm wondering if we must handle mTotalUnitBlockNum. It seems that it's not used in most of the use case.


bgoertzel:
Hi Yi-Shan,


> I have some problems about the code. I need some advices about them:
> (1)
> The EntityManager makes SpaceServer a little messy. Since we have to
> handle two objects for each scene.
> I'm thinking if it's better to make a Scene class, which represent the
> scene and own the SpaceMap and EntityManager, to replace the pair.
> Also we can move some environment parameter, ex. agent height, map(scene)
> name in the Scene object.
>

This sounds OK to me, though I haven't studied it carefully enough to be
certain it's a good idea...

In general the handling of Entities in the prior code was not adequate so
when someone wants to make code that recognizes and updates Entities
intelligently, they will have a bunch of work to do...


> (2)
> In the old Octree3DMapManager we count the number of unit blocks in the
> OcTree by mTotalUnitBlockNum. We handle this member in the
> add/removeSolidUnitBlock. But now we move the functionality to the
> OctomapOcTree. It's obvious that if someone add a block through the
> updateNode() + setBlock() function, the number of unit block will be wrong.
> I'm wondering if we must handle mTotalUnitBlockNum. It seems that it's not
> used in most of the use case.
>

It looks to me like the method

getTotalUnitBlockNum

is never called anywhere in the codebase....  So, I don't currently see the
importance of having the method there...

linas:
I merged because this is going into its own branch.

linas:
I'm slightly confused about the plans for this branch.  I just tried merging it with master, and can see that various things don't compile; for example ` error: ‘const SpaceMap’ has no member named ‘containsObject’` -- I guess the plan is to keep hacking here, until all these sorts of things get resolved, thus, eventually making this branch mergeable into master?

chenesan:
Yeah, just as your guess. My plan is --
(1) replace the old SpaceMap with Octomap
(2) replace all the functions call to old SpaceMap with new Octomap in the code base, so we can build the whole opencog and pass the unit test (can be merged in the master).
The function names in new SpaceMap may be different from old one. And some of redundant method/member may be removed. But basically we will not change the core functionality of SpaceMap.

linas:
OK sounds good to me!

PR #1752, opened 2015-08-31
Improve the memory use of Pattern Miner to 1/10

shujingke:


linas:
This doesn't compile.

linas:
I fixed one build break, but now there are a zillion of them:  
```
/home/linas/src/novamente/src/opencog-git/opencog/learning/PatternMiner/PatternMiner.cc:54:42: error: ‘class opencog::AtomSpace’ has no member named ‘getOutgoing’
     HandleSeq outgoingLinks = atomSpace->getOutgoing(link);
```

Nil changed the camel-case to snake-case quite a long time ago, now.

PR #1756, opened 2015-09-02
call the functions

rodsol:
calling the functions  which connect ConceptNode/PredicateNode with word instance 
from  some of the Helpers


williampma:
Will the code be eventually refactored to not use name-string and instead pass nodes around?

PR #1757, opened 2015-09-02
Add speech acts to the R2L outputs generated by nlp-fc

leungmanhin:
It supports declarative, imperative, truth query and interrogative speech acts generation.


opencog/nlp/relex2logic/declarative.scm
```diff
@@ -0,0 +1,50 @@
...
+            )
+            (TypedVariableLink
+                (VariableNode "$interp-node")
+                (TypeNode "InterpretationNode")
+            )
+        )
+        (AndLink
+            (LemmaLink
+                (VariableNode "$word-inst-node")
+                (WordNode ".")
```
AmeBel: Isn't there any other structure than a `.` to identify a declarative sentence? In a chat-bot context we can't be confident about punctuations, right? @ruiting @bgoertzel any suggestions.

PR #1759, opened 2015-09-04
improve to 'strict' standards, fix embedded double quotes 

kinoc:
Apply use strict.
Check mandatory variables are set, or provide default values. 
Provide proper scoping for all variables.
Convert embedded double quotes into single quotes for AIMLCODENode which is just the quoted XML of the template (for now).


AmeBel:
Thanks

PR #1760, opened 2015-09-05
Tmp call

rodsol:
call the functions from the remaining  r2l helpers


opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -1128,6 +1184,8 @@
 	[else 
     (list (InheritanceLink (ConceptNode var1_instance) (ConceptNode var1))
 	(InheritanceLink (ConceptNode var2_instance) (ConceptNode var2))
```
AmeBel: correct the square brackets above

williampma:
Do the non-instanced WordNode also need to link to the non-instanced ConceptNode/PredicateNode?  I guess that should be part of Word Sense Disambiguation.

AmeBel:
@williampma correct the syntax error please thanks

williampma:
where's the error? square brackets are allowed in scheme.

AmeBel:
i didn't know that, `[ConceptNode "what"]` works.

PR #1762, opened 2015-09-07
Readme for aim2oc

AmeBel:


AmeBel:
@kinoc just added a readme. It is what I have fished from emails, so it might not be correct. If there any corrections, please update it when you can. Thanks in advance :smile:

PR #1764, opened 2015-09-07
cmake: add modules directory to RPATH

AmeBel:
hopefully a lasting fix for https://groups.google.com/forum/#!topic/opencog/5dSwdxfucO4


AmeBel:
@linas / @ngeiswei living it to either of you to merge this.

ngeiswei:
Thanks @AmeBel. I'm not an expert in dynamic linking, I've read RPATH should be used with caution. All I can say is that it does fix the problem (and likely others as well). I'm for merging it but I'll let Linas have a say.

linas:
Since we are using RPATH already, then this change does not make anything worse.

Typically, only MacOS uses RPATH, so I was surprised to see it here. 

It would be better to set RUNPATH instead of RPATH

PR #1772, opened 2015-09-16
Fix AbsentLink in svo rule

leungmanhin:
The AbsentLink in this rule was not in correct form.


williampma:
There are probably other problems.  Some rules do not mirror the exclusion needed from the old rules.  For example, the SV rule will always fire even when SVO or SVIO is fired.

PR #1776, opened 2015-09-18
Improve induction rule

williampma:


ngeiswei:
I'm no scheme expert but I believe one could write a scheme function or macro to avoid duplicating the rules into *-{implication,inheritance,subset}.

williampma:
It has nothing to do with Scheme.  It's the inability to specific criteria inside a link-type VariableNode.

See https://github.com/opencog/opencog/pull/1722

ngeiswei:
Yes, I know, but instead of copy/pasting the rule and changing the type, you could code a function that does that for you ;-)

PR #1780, opened 2015-09-21
load missing modules for some atom types

AmeBel:


ngeiswei:
@AmeBel when getting rid of embodiment I also got it of the configuration files under it, as they contained so much obsolete content. In case you need don't hesitate to checkout https://github.com/opencog/opencog/releases/tag/old-embodiment_8-nov-2015

AmeBel:
@ngeiswei thanks for the reminder :-). 

FYI, I added the `libspacetime-types.so` b/c the nlp pipline uses `AtTimeLink` whlie I added `libembodiment-types.so` b/c it was just there.

PR #1784, opened 2015-09-25
Octomap feature

chenesan:
With some changes:

(1) Use Scene Object to save the EntityRecorder and OctomapOcTree, it's just a wrapper to prevent using pair inside the SpaceServer code.

(2) Add some cython binding of Octomap

(3) Seperate the SpaceMapUtil form SpaceMap since I think we should isolate Octomap data structure and calculation on it..


ngeiswei:
This cannot be right, @chenesan have you rebased your branch,  or something?

chenesan:
I accidentally rebase it after I push commit to my github fork..Is there anything I can do to fix this?

ngeiswei:
So what exactly happened? I guess you've rebased your branch with the upstream master and since the map 3dbranch haven't been updated for a while it considers all those commits as part of your branch. Let me look at this a bit closer... Maybe we want to chat about that in google hangout or slack...

ngeiswei:
Yeah, the SHAs seem to match so it's not that awful, although I would still recommend to redo a pull request so someone we can easily review it.

I forgot what justifies the existence the 3DMap branch. Since the embodiment has been removed, can't we just merge it?

ngeiswei:
@chenesan so, assuming I understand correctly what's going on, I think the best way to proceed is to

1. Merge the master into the 3DSpaceMap branch
2. Possible rebase again your commits with the master

this might make your PR looks much better. Of course, better would have been not to rebase in to the first place. Also I must ask you, did base your work on the 3DSpaceMap branch, or on the master branch? Obviously is you are then to merge to certain branch A, it makes more sense that you had forked A in the first place, then rebase (or better merge) A into your feature branch before doing the pull request. And I suspect A is 3DSpaceMap in your case, although I'm not entirely sure...

ngeiswei:
@chenesan ignore my previous comment and let's chat.

ngeiswei:
@bgoertzel @pennachin @shujingke @linas I'd like to understand whether you still need to keep the 3DSpaceMap branch around, or whether it is OK to merge it to the master, so I can properly advice @chenesan on what to do.

PR #1791, opened 2015-10-05
R2l 2

rodsol:


opencog/nlp/relex2logic/det.scm
```diff
@@ -0,0 +1,59 @@
...
+            )
+            (EvaluationLink
+                (DefinedLinguisticRelationshipNode "_det")
+                (ListLink
+                    (VariableNode "$noun")
+                    (VariableNode "$det")
+                )
+            )
+            (InheritanceLink 
+                (VariableNode "$noun")
```
AmeBel: I think this rule is for the a second iteration write? In other words it is a second-level rule. Thus `nlp-fc` should be modified or the rule-base has to be modified into two and the output of the first layer piped to the second. 

@williampma @leungmanhin @bgoertzel  comment pls?

AmeBel:
except the issue with modifying the `nlp-fc`, LGTM

opencog/nlp/relex2logic/det.scm
williampma: The rule itself doesn't look second iteration to me.  It's only matching WordInstanceNode and ParseNode.

And even if the rule is for 2nd iteration, FC would add the result back into the focus set and see them in the 2nd step right?  Or did something change I am not aware that makes R2L only run 1 step?

leungmanhin: I think the latest FC would apply all the R2L rules once only, so I guess there will be no 2nd step for R2L rules? Anyway this rule looks fine to me as well, relex can generate these atoms

misgeatgit: @williampma we agreed on iterating over all rules and apply them on the focus set in one call or step. This reminds me of the policy vs mechanism isolation principle that @linas mentioned on another thread. As @AmeBel  said this particular case can be handled by having N rule bases for the N phase and call FC N times which seems like a policy to my understanding.

rodsol: Hi guys ,
what is the second iteration  ??

AmeBel: @williampma there is an `InheritanceLink`, and (justed check with @misgeatgit ) the weights associated with the rules have no effect in giving priority to the application of the rules in the first and only iteration. Thus, no guarantee that this rule wil be run after the rest of the rules are applied, in the first iteration. In addition, the number of iterations has no effect as there is only a single iteration.

AmeBel: my bad, it inherits from a  `DefinedLinguisticConceptNode`

williampma: @misgeatgit @leungmanhin @AmeBel I am aware that for R2L we have agreed to apply all rules in one step, but I don't recall deciding to do this only once.  Post-processing was always in the pipeline and I always assume there will be more than one step.  ie.  all rules applied at step 1, then again all rules at step 2 (and maybe again at step 3 or 4 depends on how many level of post-processing is needed).  Maybe @bgoertzel will have some input for this?

Of cause it is possible to do `cog-fc` multiple times only different rule bases externally but that seems to be wasting the functionality of the chainer.  Maybe there some performance benefit for this, but it seems very minor.

PR #1797, opened 2015-10-08
Stop the SuReal search when a good enough solution is found

leungmanhin:


williampma:
This doesn't seem right.  Correct me if I am wrong.  From what I can tell, this is stopping the search when all binary links from the original semantic input are matched... meaning it doesn't really do anything, since any grounding would have matched the binary link to other binary link....

I thought the goal was that when a grounding is found, check the leftover semantic structure of the matched sentence, and if all the leftover are unary, then the grounding is good enough.

ie. it has nothing to do with the binary link count of the original semantic input.  You won't need to pass them into the PMCB.

leungmanhin:
I see, so looks like problem of determining whether a node is really a word needs to be fixed before doing this, otherwise it may keep returning nothing...

leungmanhin:
Oh maybe not. keep forgetting we now have reference to the word instance, let me do it in another PR

PR #1799, opened 2015-10-09
Modify SuReal to return as soon as a good enough solution is found

leungmanhin:


opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -426,17 +428,82 @@ bool SuRealPMCB::grounding(const std::map<Handle, Handle> &var_soln, const std::
         shrinked_soln[kv.first] = kv.second;
     }
 
-    // store the solution; all common InterpretationNode are solutions for this
-    // grounding, so store the solution for each InterpretationNode
-    for (auto n : qItprNode)
+    // if there are more than one common InterpretationNodes at this point,
+    // we should also have more than one SetLinks here, so examine them one by one
+    for (Handle hSetLink : qSolnSetLinks)
```
williampma: humm... isn't the set qItprNode smaller than qSolnSetLinks?  Where's the intersection check for qSolnSetLinks?

opencog/nlp/sureal/SuRealSCM.cc
```diff
@@ -149,111 +149,14 @@ HandleSeqSeq SuRealSCM::do_sureal_match(Handle h, int thoroughness)
...
-            // check all combinations of all the old mappings to the new
-            for (auto& em : existingMaps)
-            {
-                for (auto& am: appendMaps)
-                {
-                    // at this point, we know nothing in em & am would map the
-                    // same variable because they are from two disconnected
-                    // clauses.  Instead, we want to check to make sure no two
-                    // variables get mapping to the same node.
-                    auto checker = [&](const std::pair<Handle, Handle>& ekv)
```
williampma: Just curious, does handing the disconnected component handling to the PM code able to handle this?  ie. make sure no 2 variables map to the same word.

opencog/nlp/sureal/SuRealPMCB.cc
leungmanhin: These SetLinks are gathered earlier when we are retrieving the InterpretationNodes (L370), and it returns false if any of the pred_solns doesn't share a common InterpretationNode with other pred_solns (L395), so it seems reaching here means these SetLinks are the ones we want?

opencog/nlp/sureal/SuRealSCM.cc
leungmanhin: Hmm maybe not, but we also do this check this in grounding: https://github.com/leungmanhin/opencog/blob/su/opencog/nlp/sureal/SuRealPMCB.cc#L425

opencog/nlp/sureal/SuRealPMCB.cc
williampma: Doesn't seem right... qItprNode is shrinking on each iteration, but qSolnSetLinks is growing...

opencog/nlp/sureal/SuRealSCM.cc
williampma: Ah right.

opencog/nlp/sureal/SuRealPMCB.cc
leungmanhin: Oh yes, I think I understand what you mean now, let me fix this

leungmanhin:
I've pushed the fix for the qSolnSetLinks, do it this way may also be less confusing as well...

PR #1801, opened 2015-10-12
Forall instantiation rule

ngeiswei:
Supported features:
- Full instantiation (all variables in the forall).
- Partial instantiation (only one variable is instantiated).
- Type restrictions are taken into account

Unsupported features:
- Probabilistic formula (only support crisp for now)
- Term to instantiation are randomly picked by the rule application itself, the user has no control over it


williampma:
There's some scope violation here, part of the reason I had to rewrite member2eval rule and suggested why alpha-conversion is necessary.

You BindLink resolved as
```
(BindLink
   (VariableList
     (TypedVariableLink
        (VariableNode "$TyVs")
        (TypeChoice
           (TypeNode "TypedVariableLink")
           (TypeNode "VariableList")))
     (VariableNode "$B"))
  (ForAllLink
     (VariableNode "$TyVs")
     (VariableNode "$B")))
  ...
)
```

$TyVS is scoped by BindLink, but then it's re-scoped by ForAllLink.  So the $TvVs inside ForAllLink should actually be a different $TyVS then the one being referred to by the BindLink (ie. PM shouldn't ground it).

Modus Ponens rule is also broken the same way, though I don't know how to fix it.

ngeiswei:
@williampma actually this rule output isn't a ForAllLink because all variables are being instantiated.

The partial instantiation rule however may indeed violate the variable uniqueness per scope rule. But I'm not sure what it hurts in that case. Have we agreed on a variable uniqueness policy yet?

PR #1803, opened 2015-10-13
Use DefinedLinguisticPredicateNode for R2L predicates

leungmanhin:
Aims to make it easier to tell if a node is actually a word in a sentence, which will be useful in SuReal. Something similar will also need to be done for some ConceptNodes generated by R2L.


williampma:
Thansk for updating the old post-processing code and microplanner too.

PR #1808, opened 2015-10-14
ForAll Implication to higher order PLN rule

ngeiswei:
Turn a forall involving an implication into an implication over predicates (higher order).

### DIFF
--- a/opencog/reasoning/pln/rules/forall-implication-to-higher-order-rule.scm
+++ b/opencog/reasoning/pln/rules/forall-implication-to-higher-order-rule.scm
@@ -1,0 +1,77 @@
+;; =======================================================================
+;; ForAll Implication to higher order rule
+;; (TODO add wiki page)
+;;
+;; ForAllLink
+;;    V
+;;    ImplicationLink
+;;       P[V]
+;;       Q[V]
+;; |-
+;; ImplicationLink
+;;    LambdaLink
+;;       V
+;;       P[V]
+;;    LambdaLink
+;;       V
+;;       Q[V]
+;;
+;; where V is a variable, and P[V] and Q[V] are predicate bodies
+;; containing V.
+;; -----------------------------------------------------------------------
+
+(use-modules (srfi srfi-1))
+
+(use-modules (opencog exec))
+(use-modules (opencog logger))
+
+(define forall-implication-to-higher-order-variables
+  (VariableList
+     (TypedVariableLink
+        (VariableNode "$TyVs")
+        (TypeChoice
+           (TypeNode "TypedVariableLink")
+           (TypeNode "VariableList")))
+     (TypedVariableLink        
+        (VariableNode "$Impl")
+        (TypeNode "ImplicationLink"))))
+
+(define forall-implication-to-higher-order-body
+  (ForAllLink
+     (VariableNode "$TyVs")
+     (VariableNode "$Impl")))
+
+(define forall-implication-to-higher-order-rewrite
+  (ExecutionOutputLink
+     (GroundedSchemaNode "scm: pln-formula-forall-implication-to-higher-order")
+     (ListLink
+        (VariableNode "$TyVs")
+        (VariableNode "$Impl"))))
+
+(define pln-rule-forall-implication-to-higher-order
+  (BindLink
+     forall-implication-to-higher-order-variables
+     forall-implication-to-higher-order-body
+     forall-implication-to-higher-order-rewrite))
+
+;; TODO: only crisp for now
+(define (pln-formula-forall-implication-to-higher-order TyVs Impl)
+  (cog-set-tv!
+   (let* (
+          (impl-outgoings (cog-outgoing-set Impl))
+          (impl-t (car impl-outgoings))
+          (impl-d (cadr impl-outgoings)))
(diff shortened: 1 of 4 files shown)


williampma:
I am still of a mind to believe that PM matching into the variable declaration part of a scope link does not make sense.  You still have

```
 (BindLink
  ; bindlink-var-decl
  (VariableList
     (TypedVariableLink
        (VariableNode "$TyVs")
        (TypeChoice
           (TypeNode "TypedVariableLink")
           (TypeNode "VariableList")))
     (TypedVariableLink        
        (VariableNode "$Impl")
        (TypeNode "ImplicationLink"))))
  ; bindlink-pattern
  (ForAllLink
     ; forall-link var-decl
     (VariableNode "$TyVs")
     (VariableNode "$Impl"))

  bind-link-output-stuff

)
```

It does not make sense to ground (VariableNode "$TyVs") inside the ForAllLink since it's the variable declaration part of the ForAllLink (ie. a different variable from the one used by BindLink).  

It's not a variable-name uniqueness problem.  The structure just does not make much sense...

Unless you are saying the "pattern" inside a BindLink should not be consider real atom (so the ForAllLink isn't really scoping the variable)...

ngeiswei:
@williampma I don't understand what you see as problematic. Those forall rules work, in unit tests and in an example I'm building. I don't get it... Have you had a look at PLNRulesUTest.cxxtest?

linas:
William, Nil's code looks fine to me.  He is searching for a ForAll link of a very very specific kind:  It must have explicitly declared scoping variables in it.   Now, I don't see why it needs to be so very strict ....


One a tangent: There is now a mostly-working GlobNode which is  like a VariableNode but matches one-or-more atoms in a row.   Having an OptionalNode that behave like a VariableNode but matched zero or one instances could be useful.   Then Nil could handle ForAll links with either zero or one variable declarations in them....

linas:
Viz like this:
```
(TypedVariableLink
        (OptionalNode "$TyVs")
        (TypeChoice
           (TypeNode "TypedVariableLink")
           (TypeNode "VariableList")))
```

so that if it is present, then $TyVs must be typed as shown.. else if not present then .. nothing

linas:
A batter name than `OptionalLink` is solicited

ngeiswei:
@linas I think there is no useful way to use universal quantification over atoms without constraining them. Even worse, if you don't do so you get into paradoxes, like the universal quantification itself could be a substitution term.

AmeBel:
I think William is saying there are two scopes in play. One scope at the `BindLink` level, the other at the `ForAllLink` level. The way the pattern matcher works seems to overide the scoping in the `ForAllLInk` by the one found in the `BindLink` level. 

In the above case the `ForAllLink` isn't a bound-quantifier, thus the `BindLink` seems to be bounding it. But what if the `ForAllLink` came was a bound-quantifier, that is, if it had a `TypedVariableLink` for the same `VariableNode`s?

ngeiswei:
Sorry I still don't understand. Can you provide an example?

However I did realize something, while building this pattern I build a non-sensical ForAll

```atomese
  (ForAllLink
     (VariableNode "$TyVs")
     (VariableNode "$Impl"))
```

As currently written the existing ForAll rules wouldn't consider it anyway. However in the future one might write a ForAll rule that starts using it improperly. That's the only potential problem I can foresee.

linas:
Nil, Yes sure.  You could even constraint ForAll with the SignatureLink to get even more sepcific... but ... recall that ContextLink is a "better" way of having a "for-all" but constrained to a certain subset.  The wiki says more.

linas:
Amen, I think the confusion is that Nil left out a PrsentLink: he is just searching for the ForAll links, not evaluating or executing them.  Almost as if they are quoted ... 

So it should be this:
```
...
(PresentLink (ForAllLink ....
``` 

But yes, Nil, please realize that there is half-finished infrastructure in the pattern matcher such that, if it finds executable links in the body of the pattern, it will execute them first, and then do the matching on the result.  This could lead to confusion someday, if one wanted to not execute before  searching ...

ngeiswei:
@linas Oh, you mean that if ForAllLink becomes executable in the future, then ForAllLink pattern should be preceded by a PresentLink, right? I certainly like that. Although perhaps I don't need to preventively add the PresentLink today.

ngeiswei:
But of course if that's the confusion to others and the current pattern matcher will support having the PresentLink in front of the ForAllLink, I'm more than happy to add it.

AmeBel:
@ngeiswei  as Linas mentioned differentiating searching from application would simplify semantics of the rule, as Linas mentioned search vs execution.

ngeiswei:
@AmeBel I completely agree with that. However was it really the source of confusion here? (I guess @williampma would have to say).

williampma:
I was worry about the syntax, since the way the rule is written it's like assuming "pass-by-reference".

I guess PresentLink makes sense, since it sort of makes the stuff inside "virtual".

I still have to think some more on how this rule should be handled by BC.

ngeiswei:
@williampma it won't be properly handled by the BC. I've purposely let that aside for now. I obviously intend to tackle it soon enough, once I'm done building the whole inference trail of the example I'm working on.

PR #1812, opened 2015-10-15
Rename PLN rules to be consistent with the rest

ngeiswei:
PLN users beware!!! This is yet one of those useless cleanings Nil is obsessed about!

I've removed the pln prefixes as we know they are PLN rules (since there
are attached to the PLN rule-base-system), and their name is already
sufficiently long like that! (And relex2logic doesn't use that kind of
prefixing and seems to be doing fine).

So don't call your next PLN rule
```
pln-rule-<NAME>
```
instead call it simply
```
<NAME>-rule
```

Same thing for formula, do not use
```
pln-formula-<NAME>
```
instead use
```
<NAME>-formula
```


AmeBel:
:smile:  yes, the directory path is informative enough. Thanks

PR #1813, opened 2015-10-19
R2l

rodsol:
port gender rule

### DIFF
--- a/opencog/nlp/relex2logic/gender.scm
+++ b/opencog/nlp/relex2logic/gender.scm
@@ -1,0 +1,49 @@
+(define gender
+	(BindLink
+		(VariableList
+			(TypedVariableLink
+				(VariableNode "$a-parse")
+				(TypeNode "ParseNode")
+			)
+			(TypedVariableLink
+				(VariableNode "$word")
+				(TypeNode "WordInstanceNode")
+			)	
+			(TypedVariableLink
+				(VariableNode "$gtype")
+				(TypeNode "DefinedLinguisticConceptNode")
+			)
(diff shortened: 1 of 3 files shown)


opencog/nlp/relex2logic/gender.scm
```diff
@@ -0,0 +1,49 @@
...
+			(ListLink
+				(VariableNode "$word")
+				(VariableNode "$gtype")
+			)
+		)
+   )
+))
+
+(define (pre-gender-rule word gtype)
+ (cond  ((or (string=? (cog-name gtype) "masculine") (string=? (cog-name gtype) "feminine"))
```
AmeBel: the gender-rule does this checks, so why have it here? Are you thinking of having another rule for `person` gender-type?

rodsol: The PM is bounding other atoms beside DefinedLinguisticConceptNode "masculine"  and DefinedLinguisticConceptNode "feminine"  .. because of the patterns 
that is why i put the filter
please let me know if there is other way to do this

rodsol: and for "person" gender type i think we should change the label  from RelEx to unknown or something...

AmeBel: the other option is to make a `GroundedPredicateNode` out of the check and put it in the BindLink. It does the same thing but using the pattern macher.

AmeBel: Why `person` gender-type should be changed?

PR #1815, opened 2015-10-20
Remove unused (nonexistant?) library

linas:


AmeBel:
actually the library does exist https://github.com/opencog/atomspace/blob/master/opencog/persist/zmq/atomspace/CMakeLists.txt#L16

This removal have caused an issue reported @https://groups.google.com/d/msg/opencog/-lZO6xQDkMs/Uxhq-2MEBgAJ

If I got the issue right the build finishes successfully because the header files are found and the run fails because the loading of symbols from the shared-library fails, right???

Is there a means to check for such an issue using cmake configuration/build step? Ofcourse a start-stop run step could be added to circle-ci but if that is possible at the build-step it could be helpful for developers as it could possibly signal what could go wrong.

PR #1822, opened 2015-10-24
Add atomspace & cogutil include directories to compiler include paths

hedayat:
Needed when they are installed in non-standard paths


ngeiswei:
Works for me (although I didn't try for non standard paths).

PR #1823, opened 2015-10-26
Add eclipse project files to .gitignore

hedayat:


AmeBel:
thanks

AmeBel:
Assuming you are using Eclipse, would you please check http://wiki.opencog.org/wikihome/index.php/Using_Eclipse_CDT and if there is anything to update would you please point it out or update it?

hedayat:
Sure I will. :)
Thanks.

ngeiswei:
I knew there was a wiki page about Eclipse! @hedayat if you can have your Eclipse Section in the development standard refer to it, and possibly harmonize the 2 that would be great!

hedayat:
@AmeBel I tried, but since I use a different Eclipse/CMake workflow (I create makefiles with CMake and import the project into Eclipse as a Makefile project rather than creating Eclipse project with CMake), I cannot update the current guidelines (if needed). 

@ngeiswei OK! I will add a section about importing code style in the Eclipse page.

AmeBel:
Ok thanks

PR #1827, opened 2015-10-26
Implement equivalence transformation rule formula

eddiemonroe:
Implemented based on sim2inh formula:
(1+sB/sA)*simAB/(1+simAB)

### DIFF
--- a/opencog/reasoning/pln/rules/equivalence-transformation-rule.scm
+++ b/opencog/reasoning/pln/rules/equivalence-transformation-rule.scm
@@ -38,8 +38,26 @@
                     (VariableNode "$B"))))))
 
 (define (equivalence-transformation-formula AII EV)
-    (cog-set-tv!
-        AII (cog-tv EV)))
+    (let (
+            (IAB (gar AII))
+            (IBA (gdr AII)))
+         (cog-set-tv!
+            IAB
+            (equivalence-transformation-side-effect-free-formula IAB EV))
+         (cog-set-tv!
+            IBA
+            (equivalence-transformation-side-effect-free-formula IBA EV))
+    AII))
+
+(define (equivalence-transformation-side-effect-free-formula IAB EV)
+    (let* (
+            (A (gar IAB))
+            (B (gdr IAB))
+            (sA (cog-stv-strength A))
+            (sB (cog-stv-strength B))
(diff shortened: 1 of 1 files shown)


eddiemonroe:
Does it make sense that the conclusion Implication strength ends up being greater than the premise Equivalence strength? That seems strange to me...

ngeiswei:
@eddiemonroe, I'm finding another formula based on the fact that

```atomese
SimilarityLink <s>
   A
   B
```

is equivalent to

```atomese
InheritanceLink <s>
   OrLink
      A
      B
   AndLink
      A
      B
```

If I go through the math I get

```math
s = P(A and B | A or B)
s = P(A, B) / P(A or B)
s = P(A, B) / (P(A) + P(B) - P(A, B))

s = P(A, B) / (P(A) + P(B) - P(A, B))
P(A, B) = s * (P(A) + P(B) - P(A, B))
P(A, B) = s * (P(A) + P(B)) - s * P(A, B)
P(A, B) * (1 + s) = s * (P(A) + P(B))
P(A, B) = s * (P(A) + P(B)) / (1 + s)

P(B|A) = P(A, B) / P(A)
P(B|A) = (s * (P(A) + P(B)) / (1 + s)) / P(A)
P(B|A) = (P(A) + P(B)) * (s / ((1 + s) * P(A)))
```

which corresponds to

```math
(sA+sB)*(simAB / ((1+simAB) * sA)
```

in your notation.

ngeiswei:
I mean those should actually be extensional similarity and inheritance...

eddiemonroe:
@ngeisweil, the formula I'm using is from the PLN book and the wiki. And I think it's equivalent to the one you produced.

(sA+sB) * (simAB / ((1+simAB) * sA)
((sA+sB) * simAB) / ((1+simAB) * sA)
((sA+sB) *  simAB) /sA)  /  (1+simAB)
((1+sB/sA) *  simAB)  /  (1+simAB)

And exploring further, I can see why the resulting Implication strength is greater than the Equivalance. It just seemed counter-intuitive to me before.

ngeiswei:
Oh, there're equivalent, awesome! Yeah, it is counterintuitive, but I guess you saw it, since
```
P(A) < P(A or B)
```
then
```
P(A and B|A) > P(A and B|A or B)
```

PR #1829, opened 2015-10-28
Simplify deduction rule

ngeiswei:
PLN rules coders, please have a look at this, you don't need to duplicate over and over the rule code, use a rule generator instead.


opencog/reasoning/pln/rules/deduction.scm
```diff
@@ -1,102 +0,0 @@
...
-            (InheritanceLink
-                (VariableNode "$A")
-                (VariableNode "$B")
-            )
-            (InheritanceLink
-                (VariableNode "$B")
-                (VariableNode "$C")
-            )
-            ; To avoid matching (Inheritance A B) and (Inheritance B A)
-            (NotLink
```
williampma: This part is missing on the new deduction rule.

ngeiswei:
@linas maybe you can close this PR, I'm gonna do a grand PLN rule cleaning and I'll clean those variables in the process.

In meantime it would be good to find out while PLNRulesUTest fails on your system...

linas:
?? which PR?

ngeiswei:
Never mind, I'm also doing to many things in parallel ;-)

opencog/reasoning/pln/rules/deduction.scm
ngeiswei: Good catch, thanks!

ngeiswei: Hmm, and PLNRuleUTest didn't catch it, I understand why, it doesn't produce invalid outputs, just useless...

PR #1831, opened 2015-10-28
Remove pointless/useless clauses.

linas:
Repeating the variables that appear in the clauses is useless, pointless, ad potentially slows down the search.  Thus, I remove them.

Nil or William ... ca you review these, ad merge, if you agree?  I cannot run PLNRulesUTest -- it was failing before my changes, and its failing after them ... 

Also  BlendingTest and HebbianCreationModuleUTest are failing -- these were working just a few weeks ago...

### DIFF
--- a/opencog/reasoning/pln/rules/abduction-rule.scm
+++ b/opencog/reasoning/pln/rules/abduction-rule.scm
@@ -29,9 +29,6 @@
             (VariableNode "$B")
             (VariableNode "$C"))
         (AndLink
-            (VariableNode "$A")
-            (VariableNode "$B")
-            (VariableNode "$C")
             (InheritanceLink
                 (VariableNode "$A")
                 (VariableNode "$B"))
@@ -61,9 +58,6 @@
             (VariableNode "$B")
             (VariableNode "$C"))
         (AndLink
-            (VariableNode "$A")
-            (VariableNode "$B")
-            (VariableNode "$C")
             (ImplicationLink
                 (VariableNode "$A")
                 (VariableNode "$B"))
@@ -93,9 +87,6 @@
             (VariableNode "$B")
             (VariableNode "$C"))
         (AndLink
-            (VariableNode "$A")
-            (VariableNode "$B")
-            (VariableNode "$C")
             (SubsetLink
                 (VariableNode "$A")
                 (VariableNode "$B"))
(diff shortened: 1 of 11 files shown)


williampma:
Most are OK, except the modus ponens rule.  The $A is put outside on purpose so that it can backward chain with free var, instead of the implicit bound var inside the implication link...

ngeiswei:
Exactly for that kind of reasons it's not good to have the URE rule represented as BindLink. Instead a URE rule should have its own format (yet to be determined) and build BindLinks or GetLinks as necessary.

I don't mind to merge it (after the modus ponens has been fixed) though.

ngeiswei:
@linas it's strange that PLNRulesUTest fails for you. Are there other unit tests failing on your system besides Anaphora and RestApi?

ngeiswei:
Regarding the URE rule format I definitely indent to tackle it soon, but I'd rather finish my demo first and attempt to use the URE on it so my thinking can be as grounded as possible, so probably within 2 or 3 weeks.

linas:
I don't understad the comment about he modus-ponens rule.  I looked at it several times, and the Variable there seems to be extraneous. -- again, AndLink is unordered, so specifying the variable after the implication is no different than specifying it before he implication.

These unit tests fail, for me, in master:
```
The following tests FAILED:
	 10 - AnaphoraTest (Failed)
	 12 - PLNRulesUTest (Failed)
	 18 - BlendingTest (Failed)
	 22 - HebbianCreationModuleUTest (Failed)
```

ngeiswei:
@linas the reason is because the URE parses the BindLink to understand the rule. I think this is not the right approach, but removing that extraneous variableNode is gonna break backward chaining. Again, we'll fix that soon.

ngeiswei:
@linas could you parse your PLNRulesUTest stdout somewhere (maybe attach it in this issue or create a new one)?

linas:
Hmm. Its passing now.  I think I  accidentally tested some development branch; I'm doing too may things in parallel.

ngeiswei:
@linas I'm gonna take over that PR, because there is some conflicts to resolve with my latest commit.

ngeiswei:
I'm gonna do a grand PLN rule cleaning and I'll clean those variables in the process.

ngeiswei:
Hmmm, actually those dangling variable nodes might be relevant to the URE because they affect the formula... So again, yes it's stupid to have them in the BindLink because they make the rule more verbose and possibly slow down the search, but to the URE it might be useful to know that the output depends on the TV of those...

Anyway, I may still remove them, when we switch to a proper URE rule format we can re-introduce them.

PR #1832, opened 2015-10-28
Simplify many PLN rule (using generators)

ngeiswei:
+ Remove useless premises for the pattern matcher. Beware, it may be
that those premises are useful to the URE.

Most of those changed rules are not in the PLNRules utest, but I'm merging anyway, obviously very little care have been given when writing/porting those rules, I don't think anybody are currently using them, some might not even be useful, as a result I have kept my grand PLN rule cleaning to the minimum...


williampma:
Your modus ponens rules are also missing the $A outside the ImplicationLink.  It's is needed for the rule input part.

ngeiswei:
I know it's missing (I actually changed my mind while doing the cleaning), but why does the URE needs it?

williampma:
Partially grounded rule's input will generate new potential target.  The potential targets could contains variables that need to be solved, but if it's inside ImplicationLink the variables become implicitly bound.

Kind of hard to explain.  It's a problem encountered early on when I started working on the backward chainer.  Putting $A outside was a way to make the variables free and solvable.

https://groups.google.com/forum/#!topic/opencog/iIn0Dl-Ntac

ngeiswei:
@williampma I'm still not completely sure I understand the problem in this thread. Are you saying that having the ImplicationLink first then spawn the BC into recursively creating bigger and bigger ImplicationLinks?

ngeiswei:
The BC should pick up the inferences branches to explore randomly, so the order shouldn't matter.

williampma:
No, that was the old problem.

The point is that in the example in that discussion, the thing $A is matched to, need to be taken out of the ImplicationLink so that the variables inside can be grounded separately in a separate BC step.

ngeiswei:
But why would you need to ground $A separately from ImplicationLink $A $B? I mean the standalone $A and the $A in the ImplicationLink are the same, so...

williampma:
For example, if the atomspace has

```
ImplicationLink
  InheritanceLink X dog
  InheritanceLink X animal

InheritanceLink johnny dog
```

mapping `ImplicationLink $A $B` to the implication link cannot solve for X, so it need to taken out separately to find `InheritanceLink johnny dog` to know that X can map to johnny.

ngeiswei:
First of all I don't think we should keep using implicit scoping, instead of
```
ImplicationLink
   InheritanceLink X dog
   InheritanceLink X animal
```
we should use
```atomese
ImplicationLink
   VariableNode X
   InheritanceLink X dog
   InheritanceLink X animal
```
which is a shortcut for
```atomese
ImplicationLink
   LambdaLink
      VariableNode X
      InheritanceLink X dog
   LambdaLink
      VariableNode X
      InheritanceLink X animal
```
Now regarding unifiying X, I can't help but wonder whether it matters that X is a premise, since X is a variable in the term to unify...

williampma:
Not sure I understand you.
```
ImplicationLink
   VariableNode X
   InheritanceLink X dog
   InheritanceLink X animal
```
is not a rule in the rule base, so there's no unification.  The only unification is done on the Modus Ponens rule, and the extra $A in that rule allows BC to take out the implicant and treat it as standalone to resolve any variable inside.

PR #1844, opened 2015-10-30
Finished distributed version of Pattern miner

shujingke:


misgeatgit:
@shujingke  PatternMinerServer.* and PatternMinerClient.* are duplicated in LearningServer and PatternMiner dirs. Is it intentional?

PR #1848, opened 2015-11-02
R2L readme slightly cleanup

williampma:
and stuff.


leungmanhin:
Ah this cleanup reminds me two files that seem obsoleted or not needed:
https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/loader/gen-r2l-en-rulebase.scm.old
https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/loader/r2l-creator.sh.old

Maybe we can remove them as well?

williampma:
Perhaps?  I wasn't sure why that was kept in the first place.

PR #1851, opened 2015-11-03
Create the InterpretationLink before running R2L rules

leungmanhin:
The InterpretationLink is required for the R2L speech act rules, so it needs to be in the focus set.

### DIFF
--- a/opencog/nlp/scm/processing-utils.scm
+++ b/opencog/nlp/scm/processing-utils.scm
@@ -161,14 +161,14 @@
         )
     )
 
-    (define (run-fc parse-node)
+    (define (run-fc parse-node interp-link)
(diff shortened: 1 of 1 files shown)


AmeBel:
Why does the rules require an InterpretationLink? For example, isn't https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rules/imperative.scm#L24 enough?

leungmanhin:
I think it's because the speech act is in this form:
```
InheritanceLink
    InterpretationNode
    DefinedLinguisticConceptNode "...SpeechAct"
```
So we need to get the InterpretationNode from the InterpretationLink.

AmeBel:
ok

PR #1858, opened 2015-11-05
Fix FindAtomSpace dependency on HAVE_ZMQ

hedayat:
* HAVE_ZMQ is only defined if find_package(ZMQ) is done before
  find_package(AtomSpace), which is broken design. FindAtomSpace should
  either search for ZMQ itself, or not depend on it at all. In this
  patch, we only check for the existance of zmqatoms library.


AmeBel:
This patch results in `AtomSpacePublisherModuleUTest (SEGFAULT)`. Details @ https://circle-artifacts.com/gh/opencog/opencog/337/artifacts/0/home/ubuntu/opencog/build/tests/Testing/Temporary/LastTest.log

hedayat:
I have no idea. Also, I see that AtomSpacePublisherModuleUTest has printed OK. It works fine on my system and exits with exit code 0. There is no indication of SEGFAULT except the test marked as failed:

[2015-11-05 13:56:39:592] [INFO] Initializing AtomSpacePublisherModule.
.OK!
<end of output>
Test time =   5.32 sec
----------------------------------------------------------
Test Failed.


Anyway, the fix is so simple and it simply adds libzmqatoms to the list of ATOMSPACE_LIBRARIES if it exists, which is the expected behavior. Even if there is a bug, I'd argue that it is somewhere else!
Without this patch, zmqatoms is not added (while it should) and this is why people get undefined reference errors when they run cogserver with current code (and they remove check for HAVE_ZMQ as a workaround). 

Anyway, I have no idea why this patch causes a fail specially since I cannot reproduce the failure on my system. Any help is appreciated!

pennachin:
@hedayat I suggest you try running the tests on a Docker container (for instance, using the docker compose setup in https://github.com/opencog/docker/tree/master/opencog) which should be a perfect match for the circle CI environment. That may be helpful in figuring out this confusing issue...

AmeBel:
I just re-run on circle-ci it passes, it seems to be a random stuff.

AmeBel:
also tested on local machine

hedayat:
@pennachin, thanks, fortunately this time it wasn't necessary.
@AmeBel thanks! There is certainly a problem, but as I said this patch don't 'cause' the problem, it might have exposed it. And since in the first run this test has crashed during exit (after printing OK), I guess there is a problem when libraries are being unloaded. There might be a race condition (if the test runs in multi-threads?!) or something similar. BTW, it's hard to debug

PR #1865, opened 2015-11-10
R2l gender

rodsol:
gender rule

### DIFF
--- a/opencog/nlp/relex2logic/rules/gender.scm
+++ b/opencog/nlp/relex2logic/rules/gender.scm
@@ -1,3 +1,4 @@
+;check the atoms that should be grounded 
 (define (check-gender gtype)
     (if (or (string=? (cog-name gtype) "masculine") (string=? (cog-name gtype) "feminine"))
         (begin (stv 1 1))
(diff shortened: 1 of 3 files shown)


opencog/nlp/relex2logic/rules/gender.scm
```diff
@@ -0,0 +1,69 @@
...
+                (VariableNode "$word")
+                (DefinedLinguisticConceptNode "person")
+                
+            )
+        (EvaluationLink
+                (GroundedPredicateNode "scm: check-gender")
+                (ListLink
+                    (VariableNode "$gtype")
+                )
+            )
```
williampma: Is this not the same as
```
(OrLink
   (EqualLink
       (VariableNode "$gtype")
       (DefinedLinguisticConceptNode "masculine")
   )
   (EqualLink
       (VariableNode "$gtype")
       (DefinedLinguisticConceptNode "feminine")
   )
)
```
?

linas: yes I think that is the same, and is a bit easier to read (and would run faster)

PR #1871, opened 2015-11-13
Remove some hacks from the moses-pln-synergy

ngeiswei:
In particular most of what is related to manipulating lambda
expressions.

### DIFF
--- a/examples/moses-pln-synergy/scm/README.md
+++ b/examples/moses-pln-synergy/scm/README.md
@@ -76,16 +76,204 @@ And search for the following
    )
 ...
 
-scheme@(guile-user)> (cog-bind pln-rule-eliminate-neutral-element-hack)
-scheme@(guile-user)> (cog-bind pln-rule-eliminate-dangling-junctor-hack)
-scheme@(guile-user)> (cog-bind pln-rule-equivalence-hack)
-$6 = (SetLink
+scheme@(guile-user)> (cog-bind implication-lambda-distribution-rule)
+...
    (ImplicationLink (stv 1 0.99999982)
-      (PredicateNode "take-treatment-1" (stv 0.1 0.80000001))
-      (PredicateNode "take-compound-A" (stv 0.2 0.80000001))
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (AndLink
+            (EvaluationLink
+               (PredicateNode "take")
+               (ListLink
+                  (VariableNode "$X")
+                  (ConceptNode "treatment-1")
+               )
+            )
+            (EvaluationLink (stv 1 0.99999982)
+               (PredicateNode "contain")
+               (ListLink
+                  (ConceptNode "treatment-1")
+                  (ConceptNode "compound-A")
+               )
+            )
+         )
+      )
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "compound-A")
+            )
+         )
+      )
+   )
+...
+scheme@(guile-user)> (cog-bind and-lambda-distribution-rule)
+...
+   (AndLink
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "treatment-1")
+            )
+         )
+      )
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink (stv 1 0.99999982)
+            (PredicateNode "contain")
+            (ListLink
+               (ConceptNode "treatment-1")
+               (ConceptNode "compound-A")
+            )
+         )
+      )
+   )
+...
+scheme@(guile-user)> (cog-bind lambda-grounded-construction-rule)
+$5 = (SetLink
+   (LambdaLink (stv 1 0.99999982)
+      (TypedVariableLink
+         (VariableNode "$X")
+         (TypeNode "ConceptNode")
+      )
+      (EvaluationLink (stv 1 0.99999982)
+         (PredicateNode "contain")
+         (ListLink
+            (ConceptNode "treatment-1")
+            (ConceptNode "compound-A")
+         )
+      )
    )
 )
 
+scheme@(guile-user)> (cog-bind implication-construction-rule)
+...
+   (ImplicationLink (stv 1 0.99999982)
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "treatment-1")
+            )
+         )
+      )
+      (LambdaLink (stv 1 0.99999982)
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink (stv 1 0.99999982)
+            (PredicateNode "contain")
+            (ListLink
+               (ConceptNode "treatment-1")
+               (ConceptNode "compound-A")
+            )
+         )
+      )
+   )
+...
+
+scheme@(guile-user)> (cog-bind implication-implicant-distribution-rule)
+...
+   (ImplicationLink (stv 1 0.99999982)
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "treatment-1")
+            )
+         )
+      )
+      (AndLink
+         (LambdaLink
+            (TypedVariableLink
+               (VariableNode "$X")
+               (TypeNode "ConceptNode")
+            )
+            (EvaluationLink
+               (PredicateNode "take")
+               (ListLink
+                  (VariableNode "$X")
+                  (ConceptNode "treatment-1")
+               )
+            )
+         )
+         (LambdaLink (stv 1 0.99999982)
+            (TypedVariableLink
+               (VariableNode "$X")
+               (TypeNode "ConceptNode")
+            )
+            (EvaluationLink (stv 1 0.99999982)
+               (PredicateNode "contain")
+               (ListLink
+                  (ConceptNode "treatment-1")
+                  (ConceptNode "compound-A")
+               )
+            )
+         )
+      )
+   )
+...
+scheme@(guile-user)> ;; If X takes treatment-1 then X takes compound-A
+scheme@(guile-user)> (cog-bind deduction-implication-rule)
+...
+   (ImplicationLink (stv 0 0.80000001)
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "treatment-1")
+            )
+         )
+      )
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
+         (EvaluationLink
+            (PredicateNode "take")
+            (ListLink
+               (VariableNode "$X")
+               (ConceptNode "compound-A")
+            )
+         )
+      )
+   )
+...
 scheme@(guile-user)> ;; Infer that being well hydrated speeds up recovery 
 scheme@(guile-user)> (cog-bind pln-rule-average-hack)
 scheme@(guile-user)> (cog-bind pln-rule-modus-ponens)
--- a/examples/moses-pln-synergy/scm/pln-config.scm
+++ b/examples/moses-pln-synergy/scm/pln-config.scm
@@ -47,6 +47,11 @@
   (list (string-append pln-rules-dir "deduction-rule.scm")
         (string-append pln-rules-dir "modus-ponens-rule.scm")
         (string-append pln-rules-dir "implication-instantiation-rule.scm")
+        (string-append pln-rules-dir "implication-lambda-distribution-rule.scm")
+        (string-append pln-rules-dir "and-lambda-distribution-rule.scm")
+        (string-append pln-rules-dir "lambda-grounded-construction-rule.scm")
+        (string-append pln-rules-dir "implication-construction-rule.scm")
+        (string-append pln-rules-dir "implication-implicant-distribution-rule.scm")
         "pln-rules/implication-or.scm"
         "pln-rules/hack.scm"
         )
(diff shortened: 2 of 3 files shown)


examples/moses-pln-synergy/scm/README.md
```diff
@@ -76,16 +76,204 @@ And search for the following
...
+                  (ConceptNode "compound-A")
+               )
+            )
+         )
+      )
+      (LambdaLink
+         (TypedVariableLink
+            (VariableNode "$X")
+            (TypeNode "ConceptNode")
+         )
```
linas: Well, this is a valid but somewhat funny way of saying things. The normal predicate-logic way of saying this is "for all x, if x takes treatment-1 and treament-1 contains compound-A then x takes compound-A"  There is a specific rewrite rule (I forget what its called, and I don't have my book on logic here with me) that allows you to say  "double-turnstile forall-x p(x) -> p(y) " (double-turnstile symbol because this is a semantic consequence, not a logical consequence).     

In ordinary predicate-logic p(y) always means lambda.y p(y) -- that is implicilty understood. So you converted a for-all rule into a specific rule.  Seeing the lambda there is .. correct, but unexpected and surprising, somehow.  It could use some articulation.

examples/moses-pln-synergy/scm/README.md
```diff
@@ -76,16 +76,204 @@ And search for the following
...
+         (EvaluationLink (stv 1 0.99999982)
+            (PredicateNode "contain")
+            (ListLink
+               (ConceptNode "treatment-1")
+               (ConceptNode "compound-A")
```
linas: Huh??? I can't parse this at all. Is this a bug?

examples/moses-pln-synergy/scm/README.md
```diff
@@ -76,16 +76,204 @@ And search for the following
...
+            )
+         )
```
linas: This is just wrong, you slipped a parenthesis somewhere .... as written with the parens above, it says "if X takes treatment-1 then treatment-1 contains compound A" which can be either true or false, but is certainly not universally true.  (unless it is true that treatement-1 actually does contain compound-1 ... does it ???)

examples/moses-pln-synergy/scm/README.md
```diff
@@ -76,16 +76,204 @@ And search for the following
...
+                  (ConceptNode "treatment-1")
+                  (ConceptNode "compound-A")
+               )
+            )
+         )
+      )
+   )
+...
+scheme@(guile-user)> ;; If X takes treatment-1 then X takes compound-A
+scheme@(guile-user)> (cog-bind deduction-implication-rule)
```
linas: well, again, this seems very awkward, because X is free in the english-language phrase, whereas in the below, there are two different X's that are both lambda-bound below. --  there are two distinct x's which can be alpha-converted to any letter, say, y and z.   So, below, what is being said is that  "double-turnstile (its a semantic consequence that) [(lambda.y) p(y) -> lambda.z q(z) ]"  where p(y) is  "y takes treatment-1" and q(z) is "z takes compound-A".  By using the same symbol x in both lambdas, its confusing and misleading.

By contrast, in the statement above, "If X takes treatment-1 then X takes compound-A" the variable x is free, and not bound!   The only way to  make it free is to to say  "double-turnstile (its a semantic consequence that) [(lambda.y) p(y) -> lambda.z q(z) ] (x)"  which can then be beta-reduced to "  "double-turnstile (its a semantic consequence that) [p(x) ->q(x)]" so that x is now free.

Alternately you can turn this back into a combinator by binding x again: "double-turnstile (its a semantic consequence that) lambda.x [p(x) ->q(x)]"   

To summarize, the english-language explanation is a predicate-logic statement with free variables, but the actual code is some combinators with no free variables.  This is confusing because there's a lot of computation and juggling you have to do in your head to convert the one to the other.   Better to write it out ...

examples/moses-pln-synergy/scm/README.md
ngeiswei: That is where PLN differs from predicate logic. In PLN the implication is a higher order operator over the domain of predicates. The notation

```
ImplicationLink
   X
   P[X]
   Q[X]
```

is just a shorthand for

```
ImplicationLink
   Lambda
      X
      P[X]
   Lambda
      X
      Q[X]
```

where P[X] and Q[X] indicates an expression containing X.

examples/moses-pln-synergy/scm/README.md
ngeiswei: No, it's not a bug. AndLink has 3 versions (overloaded). A version over predicates, concepts, and TVs (or rather atoms that are neither predicate nor concept but still have TVs). This rule just perform the transformation from a AndLink over TVs to an AndLink over predicates. Distributing the Lambda in order to shorten the scoping of the variables.

examples/moses-pln-synergy/scm/README.md
ngeiswei: That is exactly what it is saying :-)

Since treatment-1 does contain compound-1, we can infer that anything implies that treatment-1 contains compound-1.

examples/moses-pln-synergy/scm/README.md
ngeiswei: Just to be sure, note that the scopes of X do not overlap. Then, again PLN is not predicate logic, and Implication is not universally quantified, it is universally conditionally quantified, and to express this the Implication operates directly on predicates.

examples/moses-pln-synergy/scm/README.md
linas: OK, but please understand that what you wrote is not syntactically correct.   You should be writing this:
```
ImplicationLink
   Lambda
      Y
      P[Y]
   Lambda
      Z
      Q[Z]
```

Using the same variable X for both lambdas is confusing.  Again: every lambda can be (and should be) alpha-converted. Failing to do so leads either to confusion, or worse, incorrect calculations.

ngeiswei: It is syntactically correct.

linas: The variable X appears twice. It should not.  In textbooks, this would be considered a syntax error, although currently the atomspace and PLN allow this.

ngeiswei: In textbooks??!!! What textbooks?

linas: ? Textbooks on logic?  My library here has many shelves of them. I took the effort to read one recently; Dirk van Dalen, "logic and structure" or something like that. Its aimed ad undergraduates.

ngeiswei: And so you found in there that variables under different non overlapping scopes must be uniquely named?

linas: No, look, I'm just saying that this example is confusing due to a large variety of issues, listed above and below.  I worked out what I think was the intended meaning of the stuff here, but it was hard, I had to think a log about it, I had to do some calculations with pencil and paper to verify the results.  ... and I am already familiar with OpenCog!  So I think these examples are much too hard for the beginner, who is trying to learn the material for the first time.  They should be clearly explained, and the notionional difficulties should be fixed.

ngeiswei: Ah, OK. Actually I spent half of yesterday adding explanations to make it clearer. But it's by no mean meant to the beginner, it is meant to me, so I can understand what the URE needs to be to deal with complex PLN inference chains like this one.

linas: Well, that's an interesting statement. Naively, in ordinary predicate logic, this would seem to be a fairly straight-forward thing to derive (not sure, I have not tried)  but here there is extra complexity, and I can't tell why the complexity is needed. You state that you want to use higher-order logic, implying that the variables could range over sets.  But then you convert everything to combinators, which have no free variables in them at all. -- that's not higher-order logic, that's just combinators.... I can't tell why this step is needed: it just seems to add complexity. 

Why not just use free variables, and a for-all link? Why all the fiddle-faddle with lambda?  What does the lambda allow you to do that ordinary predicate logic does not?

ngeiswei: In predicate logic you can write
```predicate_logic
(ForAll x P(x)) -> (ForAll x Q(x))
```
it's possible, although usually people prefer to move all the quantifiers at the top level. Here we *cannot* do that, because
```atomese
ImplicationLink <TV>
   P
   Q
```
is not convertible to a single universal quantification. Instead it is a conditional quantification, the condition predicate is P, and the quantified predicate is Q. The formula is
```math
Sum_x P(x) and Q(x) / Sum_x P(x)
```
You cannot turn that into a universally quantification like in predicate logic, or even average quantification. So you cannot move all the quanfiers at the top level of the proposition.

Because of that you cannot do anything useful in PLN with free variables, cause, OK you've got a proposition with free variables (i.e. universally quantified), which means something that is not even useful to calculate the TVs of what matters to you, i.e. conditioned quantifications (i.e. ImplicationLinks).

So yes it make inference more complicated than with predicate logic, but I don't see another way, and I actually think it really is the simplest way.

By higher order I mean the ImplicationLink, not the domain, sorry for the confusion. The ImplicationLink works over sets, not individuals, just like map or fold, in this sense I mean higher order.

linas: What I meant was this: In predicate logic, if one has that
```
ForAll x (P(x) -> Q(x))
```
then one can deduce that it holds for any single choice of `x` that is, that 
```
P(x) -> Q(x)
```
In the above, x is free.   You can make it bound, by writing
```
lambda y (P(y)->Q(y))
```
Lets go back to the first expression, lets call it `A`  that is `A` is the expression `ForAll x (P(x) -> Q(x))`  Now, in classical predicate logic, `A` would be either true or false. In Bayesian logic, yes, assigning a 'probability' of  `Sum_x (P(x) and Q(x)) / Sum_y P(y)` to `A` is the correct answer.  (Please observe my careful use of the bound variable `y` in the above.) Note that this assignment of this truth-value to `A`  predates the arrival of either you or I or Ben on this planetary surface. I believe that this formula can be found in even old texts; its "obvious".  

Now suppose that `L` stands for `lambda y (P(y)->Q(y))` -- the question then is: what is the truth value of `L`?  You seem to be stating that it has exactly the same truth value as `A` does !??  That is not at all clear to me. 

Consider now the expression `M` which is `P(x)->Q(x)`  What is the truth value for `M`?  It seems plainly clear to me that both `L` and `M` have the same truth value, but you seem to disagree about that too.  By the way, observe that `M = L(x)`  which is hopefully obvious, and needs no explanation, right? So of course both `L` and `M` should have the same truth value ... how could they not?  Yet you reject `M` and embrace `L`, and I don't understand why.  They are just two sides of the same coin.

Finally, consider the opencog expression 
```
ImplicationLink <TV>
   P
   Q
```
You seem to claim that this is identical to `L`  which is fine by me, but then you insist that it has a truth-value that is identical to `A`.   And then you claim that `M` cannot even have a truth value.   This is just one of many, mulltiple confusions.

ngeiswei: @linas in my view, in Atomese/PLN neither `A` nor `L` nor `M` are correct. At best they are confusing, at worse they are ill-defined.

The only basically correct notations is:

```
ImplicationLink
   P
   Q
```

where P and Q are PredicateNodes or something equivalent to it, like a defined or grounded predicates, or lambdas.

Another notation exists due to its convenience and terseness:

```
ImplicationLink
   <variables>
   <P-body>
   <Q-body>
```

which is equivalent to

```
ImplicationLink
   LamdaLink
      <variables>
      <P-body>
   LambdaLink
      <variables>
      <Q-body>
```

as explained in the wiki page about ImplicationLink.

Can we do more than that and use free variables? I think we can, but for starter it's safer to avoid implicit scoping, that is why I've chosen to clutter the demo with lambdas (ultimately they could be avoided if you use more defined predicates along the way, but I wanted to see how it looks if I manipulate lambda expressions directly).

Now regarding instantiation: please note that there is no universal quantification over ImplicationLink, instead it is condition quantification, P being the condition. When you perform instantiation with an ImplicationLink expression, you don't end up with an instantiated ImplicationLink, you directly end up with an instantiated implicand. Why? Because an instantiated ImplicationLink isn't even defined anyway, and you don't need it, it's more efficient to get what matters, the instantiated implicand.

The ImplicationLink wiki doesn't go enough in depth, I'll try to remedy that soon.

linas: But what you are saying seems to be inconsistent with the TV you want to assign to it!  The only way taht I can understand this:
```
ImplicationLink
   LamdaLink
      <variables>
      <P-body>
   LambdaLink
      <variables>
      <Q-body>
```
is as `(lambda.x P(x)) -> (lambda.y Q(y))`   which is the same thing as `lambda.x lambda.y (P(x)->Q(y))` and the TV of that is (obviously)  `sum_x sum_y (P(x) and Q(y)) / sum_z P(z)`  but that is not the TV that you want to use.  I'm quite thoroughly confused.

ngeiswei: But I don't understand

```
sum_x sum_y (P(x) and Q(y)) / sum_z P(z)
```

isn't how probabilities are usually calculated.

```
P(B|A) = P(B, A) / P(A)
```

the sums involved in estimating the probabilities of P(B, A) and P(A) are
```
P(B, A) = sum_x B(x) and A(x) / <sample size>
P(A) = sum_x A(x) / <sample size>
```
And so you end up with
```
P(B|A)= [sum_x B(x) and A(x)] / [sum_x A(x)]
```
note that I've used square brackets to delimit the scope of sum_x.

I'm sure you agree with all this, right?

ngeiswei: Ah, I understand the source of your confusion! :-)

No, `(lambda.x P(x)) -> (lambda.y Q(y))` is not the same thing as `lambda.x lambda.y (P(x)->Q(y))`.

linas: Huh? in what way is it possible that  `(lambda.x P(x)) -> (lambda.y Q(y))` isn't identical to `lambda.x lambda.y (P(x)->Q(y))`  ???

ngeiswei: There could be more than one reason, but the primary one is that `P(x) -> Q(y)` is not well defined. Only P -> Q is well-defined, where P and Q are predicates.

Likewise, you wouldn't be surprised that
```scheme
(map (lambda (x) (foo x)) my-list)
```
isn't identical to
```
(lambda (x) (map (foo x) list))
```
and may in fact be ill-defined (depending here on what (foo x) is)..

ngeiswei: Well you *can* somewhat define P(x)->Q(x) by considering P(x) as a predicate over all contexts in which P(x) is true (like Ben likes to do), but you can see that you are shitting from the meaning of P->Q, it means something else, and there are implicit steps to involved come up with this definition of P(x)->Q(x) that are for now best to avoid. Let's makes everything as explicit as possible for now. Once everything works that way we can start adding sugar all over the place if we want to.

ngeiswei: One more remark as to why `P(x)->Q(x)` is ill-defined. `P(x)->Q(x)` is essentially in PLN `Prob(Q|P)`. Now how would you define `Prob(Q(x)|P(x))`? A probability is over events (power-sets), but Q(x) and P(x) are not power-sets. OK, let's make a notation abuse and say that `P(x)` is in fact the singleton `{P(x)}`, and let's assume that Omega is discrete so that `{P(x)}` has a well-defined probability. What would be `Prob(Q(x)|P(x))`? It would either 1, if `Q(x)` and `P(x)` denotes the same element in Omega. Or undefined if `Q(x)` and `P(x)` denotes 2 different elements. How useful is that? It isn't... You can wrap a lambda around that, but it's not equal to `Prob(Q(x)|P(x))` (which is what matters to us).

In other words, yes you may try to come up with ways to define `P(x)->Q(x)`, but they are not useful anyway!

linas: Yes, I completely agree that P(x)->Q(y) is completely weird, but you were the one who wrote it with two lambdas, not me.  Having two different lambdas seems quite wrong to me.  More tomorrow after some sleep.

ngeiswei: Oops, my previous message contains typos, {x} is the probabilitized singleton not {P(x)}, you get the idea anyway.

PR #1873, opened 2015-11-17
Remove more hacks and improve comments

ngeiswei:


examples/moses-pln-synergy/scm/README.md
```diff
@@ -164,6 +179,10 @@ $5 = (SetLink
    )
 )
 
+### (5) Build the tautology that if X takes treatment-1, then
+    treatment-1 contains compound-A. Semi-formally
+    (\x take(x, treatment-1)) -> (\x contain(treatment-1, compound-A))
```
linas: well, this is still wrong; one is x, the other is y; you are using the variable x twice, which is just confusing.  These cannot be the same x, yet you use the same symbol, twice.

examples/moses-pln-synergy/scm/README.md
```diff
@@ -242,7 +264,118 @@ scheme@(guile-user)> (cog-bind implication-implicant-distribution-rule)
       )
    )
 ...
-scheme@(guile-user)> ;; If X takes treatment-1 then X takes compound-A
+
+### (7) Factorize the lambda in the implicand of (6). Semi-formally
+    (\x take(x, treatment-1)) and (\x contain(treatment-1, compound-A))
+    -> (\x (take(x, treatment-1) and contain(treatment-1, compound-A)))
```
linas: ditto -- the variables should be changed to x y and z, instead of 3 x's in a row. They're combinators.

examples/moses-pln-synergy/scm/README.md
ngeiswei: As long as the variables do not have overlapping scopes I don't understand what is confusing. Do you find confusing to write
```c
int inc(int x) { return x + 1; }
int dec(int x) { return x - 1; }
```
?

linas: People are used to reading C code. But PLN and atoms are all new, and the exact meaning is not new if one is a beginner. Its hard to tell if the variable name choice is significant or not (ditto, if you did not know C).

ngeiswei: I consider the meaning of a variable in Atomese/PLN to be the same as in lambda calculus. And lambda has the same meaning as well, it is an abstraction, it exposes a variable to be substituted in a body, that's all, as soon as there are no variable name clash, variables can share the same name at multiple places.

linas: exposes->hides.  Free variables are exposed, bound ones are hidden. lambdas convert expressions into combinators.

ngeiswei: I understand what you mean. You are right, saying that a lambda exposes the bound variables to the user is incorrect, what it exposes is an abstraction, and that abstraction contains a variable (which name is irrelevant from the user's view) and a body. I guess from the outside (the user) using an abstraction or combinators would make no difference, so yes I agree.

PR #1874, opened 2015-11-20
SuReal updates

leungmanhin:
This is for handling disconnected clauses in SuReal, only search within the set of atoms got matched previously


leungmanhin:
@williampma I've updated a few things here, please have a look, thanks!

PR #1878, opened 2015-11-23
Prelim AIML translator and interactor

kinoc:
Update to translate subset of AIML into OpenCog format, along with simple console REPL interpreter.


linas:
thank you!  will review over the coming week(s)

PR #1879, opened 2015-11-23
More updates in sureal grounding

leungmanhin:


williampma:
Looks like you should write a unit test for testing "good enough" too.

leungmanhin:
Ok I'll add that as well

PR #1884, opened 2015-11-25
Tense

rodsol:


opencog/nlp/relex2logic/rules/tense.scm
```diff
@@ -0,0 +1,64 @@
+;Both link grammar and relex output are selected because of the pattern 
+;Example : "He gave me the book."
+;(DefinedLinguisticConceptNode "past") and (DefinedLinguisticConceptNode ".v-d")
+;I just want to avoid the Link Grammar output. There might be another way to do this
+(define (check-tense tense)
+    (if (string-contains (cog-name tense) ".")
```
AmeBel: have you tried AbsentLink

rodsol: No, But I just checked it now (http://wiki.opencog.org/w/AbsentLink) . From the explanation on that page it doesn't seems a solution for this case.
Because here  two DefinedLinguisticConceptNode 's are present in the atomspace… one from link grammar and the other from relex and they are selected because of there pattern. I want the pattern matcher not to select the DefinedLinguisticConceptNode  from link grammar

linas: You could do something like this, maybe:
```
(NotLink (EqualLink (VariableNode "$tense") (DefinedLinguisticConceptNode ".v-d")))
```

Note, however, relex is broken and should be fixed: it is generating
```
; tense (run, past)
(InheritanceLink (stv 1.0 1.0)
   (WordInstanceNode "ran@6bba3d6e-3b62-40d2-a24b-634352abc6c3")
   (DefinedLinguisticConceptNode "past")
)
; subscript-TAG (run, .v-d)
(InheritanceLink (stv 1.0 1.0)
   (WordInstanceNode "ran@6bba3d6e-3b62-40d2-a24b-634352abc6c3")
   (DefinedLinguisticConceptNode ".v-d")
)
```
and that is just wrong; one of these should be TenseLink, and the other should be SubscriptLink.  The inheritanceLink is useless, because it gloms two completely different categories together into one.

rodsol: But there is no TenseLink or SubscriptLink in current atome type ,right?

linas: right. The would need to be added. Adding them is not hard, you edit nlp-types.script and add them.

AmeBel: and update relex to create those link types.

linas:
Hi Rodas, could you take a look at pull request #1910 ?

I made several simplifications:
-- removed the pre-tense-rule helper
-- changed tense-rule so that it takes atoms, not strings
-- changed tense-rule to return a ListLink

If you add new rules in the future, could you follow this format? It makes the code slightly simpler & faster.

linas:
Also pull req #1911 adds a new TenseLink type.  In order to actually get it, you will need to pull and build a new version of Relex.

rodsol:
Alright

PR #1886, opened 2015-11-26
get chatbot running, again, after precious changes

linas:


opencog/nlp/relex2logic/rules/definite.scm
```diff
@@ -18,24 +19,23 @@
...
             )
         )
-      )
     )
 )
 
-; This is function is not needed. It is added so as not to break the existing
-; r2l pipeline.
+; This is function is not needed. It is added so as not to break the
+; existing r2l pipeline.  Huh ??? How can it not be needed?  It is
```
AmeBel: The wording is misleading I guess. What is meant is `definite-rule` could be modified to accept a node instead of the name of the node. But doing that at the time of adding this rule was cumbersome, as it meant modifying the java code that generated the call in the previous r2l-implementation. In short it is a "temporary hack"

linas: OK, well, this seems to be done in multiple places in the code, and it does not seem to be obvious in any way.  The fix isn't clear to me.

AmeBel: Since we have moved to ure based r2l implementation, there is no need to keep backward compatability, and rules that are added should update the helper functions to accept nodes and not names of the nodes. Then we can avoid the `pre-whatever-helper-function-name` indirections.

PR #1892, opened 2015-11-26
Added Ros as a requirement for spatial/spacetime

rTreutlein:
this should fix the failing build.


pennachin:
As this still didn't fix the build, I've reverted the recent merges. Once we figure out what causes the build to fail, we'll remerge the 3DSpaceMap code. We should also figure out if the version that should be merged is the one on this repo, the one on the Minecraft repo, or some combination of both.

rTreutlein:
Will Try to fix it in the 3DSpaceMap branch.

linas:
Cassio, I think you are mis-understanding the situation. This back and forth is not useful or productive.

PR #1895, opened 2015-11-27
Fixed Ros dependency

rTreutlein:
I had to merge my master into this, that is why there are so many commits.


pennachin:
There should be a way to cherry pick commits or something like that, I hope we don't have to merge 509 commits.

rTreutlein:
@pennachin Well most of those are in master already so it shouldn't be to big an issue. I tried cherry-picking but there was other stuff that was failing in this branch.

linas:
Gah, Well, I had taken a lot of time and effort to cherry-pick exactly the right stuff but Cassio revered all of it.

linas:
Basically, you got this mess because (a) cassio reverted, and (b) he he did it wrong and then (c) you commited this without review  and (d) you did it wrong.

I'm pissed off at the fucking going around here cassio Get fucking rid of the goddamned circle CI stuff and quit making everyone's lives miserable.

linas:
Sorry. I'm just in a bad mood, I guess.

PR #1919, opened 2015-12-02
Fix anaphora unit test

leungmanhin:
This and the python path fix in #1917, should fix the anaphora unit test, let's see


pennachin:
Yay! Thanks!

PR #1924, opened 2015-12-02
Remove pointless/useless clauses.

linas:


linas:
never mind clicked on the wrong thing

PR #1925, opened 2015-12-02
Change truthvalue include and library paths

ngeiswei:
Companion PR for https://github.com/opencog/atomspace/pull/512


linas:
OK by me.

PR #1935, opened 2015-12-04
Move cython binding for moses to the moses repo

AmeBel:
Associated with https://github.com/opencog/moses/pull/30.


linas:
Thanks!

PR #1938, opened 2015-12-08
Fix deduction consistency

ngeiswei:


ngeiswei:
Some old bad commit found there way here, let me close that.

PR #1951, opened 2015-12-21
Fuzzy matcher for sentence-matching

leungmanhin:
This is mainly for the question-answering part of the chatbot, it needs opencog/atomspace#541 to be merged first in order to be functional


linas:
According to CircleCI, this does not compile:
```
 /opencog/opencog/nlp/fuzzy/Fuzzy.h:27:45: fatal error: opencog/query/FuzzyPatternMatch.h: No such file or directory
 #include <opencog/query/FuzzyPatternMatch.h>
```

leungmanhin:
Yes the updates in opencog/atomspace#541 should solve that problem

opencog/nlp/fuzzy/FuzzySCM.cc
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @param pat        The input pattern
+ * @param rtn_type   The return type (the type of atom we are looking for)
+ * @param excl_list  The exclude-list, containing a list of atoms that we
+ *                   don't want in the result set
+ * @return           A list of solutions and their similarity scores
+ */
+Handle FuzzySCM::do_nlp_fuzzy_match(Handle pat, Type rtn_type,
+                                    const HandleSeq& excl_list,
+                                    bool dup_check)
```
linas: ahh .. I see. This is the complicated API that pull req https://github.com/opencog/atomspace/pull/546 is for.  

It would be better if "exclude_list" as actually part of the pattern, by using AbsentLink.

I also really really need a "must-have" list, of terms that must appear in the pattern, exactly as they are, no fuzziness. It would be best if this was a part of the pattern, as well.  Not sure how that could work.   I'm wriing an email now, to suggest an idea...

leungmanhin: Yes I think it's a good time to use the AbsentLink instead

linas: OK, started thinking about how to do this., Its not as easy as I first thought.

PR #1952, opened 2015-12-21
Fix cogserver so it uses same atomspace as everyone else..

linas:
The guile bindings for the cogserver were broken. 

Previously, the cogserver always created a new atomspace, and used it, even if there was an existing atomspace being used.  This made existing atomspace contents invisible to the cogserver.

This requires new code added to the atomspace!  So go `git pull; make; sudo make install` that first.


linas:
circle-CI is failing because it is not pulling in the current atomspace. This results in a build break.

PR #1957, opened 2015-12-22
openpsi cleanup

AmeBel:


linas:
thanks. 

who is working on pln? Can we move that out of the "reasoning" directory?

AmeBel:
I think Nil .

PR #1958, opened 2015-12-23
Fix no module test_helpers error in cogserver

misgeatgit:


opencog/python/preload_functions/test_functions.py
```diff
@@ -1,6 +1,11 @@
 # 
 # Test Functions preloaded into the CogServer
 
+# Modify PYTHON_PATH so that test_helpers module gets registered.
+from sys import path 
+from os.path import realpath
+path.append(realpath('..')+"/opencog/python")
```
linas: well, the `cogserver.conf` file should already be doing this.  That is, the correct path should already be searched, so this hack should not be needed.  Maybe cogserver.conf is broken?   But whatever. Path stuff in opencog is a horrible mess.

PR #1967, opened 2015-12-26
Fix Cogserver loading issue

AmeBel:
due to the recent directory restructuring a clean build and run of cogserver, had issues. This fixes the issue and updates the path related documentation as well

### DIFF
--- a/lib/cogserver.conf
+++ b/lib/cogserver.conf
@@ -73,29 +73,29 @@ ANSI_ENABLED	       = true
 
 # Cogserver in OSX will automatically change .so extension to .dylib
 # if .so exists.
-MODULES               = opencog/modules/libbuiltinreqs.so,
-                        opencog/modules/libnlp-types.so,
-                        opencog/modules/libattention-types.so,
-                        opencog/modules/libPLNTypes.so,
-                        opencog/modules/libembodiment-types.so,
-                        opencog/modules/libPersistModule.so,
-                        opencog/modules/libPersistZmqModule.so,
-                        opencog/modules/libQueryModule.so,
-                        opencog/modules/libLGDictModule.so,
-                        opencog/modules/libSuRealModule.so,
-                        opencog/modules/libattention.so,
-                        opencog/modules/libAtomSpaceExtensions.so,
-                        opencog/modules/libPythonModule.so,
-                        opencog/modules/libpy-shell.so,
-                        opencog/modules/libscheme-shell.so,
-                        opencog/modules/libRuleEngineModule.so
+MODULES               = opencog/cogserver/modules/libbuiltinreqs.so,
+                        opencog/cogserver/modules/libnlp-types.so,
+                        opencog/cogserver/modules/libattention-types.so,
+                        opencog/cogserver/modules/libPLNTypes.so,
+                        opencog/cogserver/modules/libembodiment-types.so,
+                        opencog/cogserver/modules/libPersistModule.so,
+                        opencog/cogserver/modules/libPersistZmqModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libLGDictModule.so,
+                        opencog/cogserver/modules/libSuRealModule.so,
+                        opencog/cogserver/modules/libattention.so,
+                        opencog/cogserver/modules/libAtomSpaceExtensions.so,
+                        opencog/cogserver/modules/libPythonModule.so,
+                        opencog/cogserver/modules/libpy-shell.so,
+                        opencog/cogserver/modules/libscheme-shell.so,
+                        opencog/cogserver/modules/libRuleEngineModule.so
 
 # Optional modules, not enabled by default
-#                        opencog/modules/libatomspacepublishermodule.so
-#                        opencog/modules/libbenchmark.so,
-#                        opencog/modules/libhebbiancreation.so
-#                        opencog/modules/libdimensionalembedding.so
-#                        opencog/modules/libviterbi.so
+#                        opencog/cogserver/modules/libatomspacepublishermodule.so
+#                        opencog/cogserver/modules/libbenchmark.so,
+#                        opencog/cogserver/modules/libhebbiancreation.so
+#                        opencog/cogserver/modules/libdimensionalembedding.so
+#                        opencog/cogserver/modules/libviterbi.so
 
 # Load order of the SCM modules:
 # First, the user-defined config file.
--- a/lib/opencog-chatbot.conf
+++ b/lib/opencog-chatbot.conf
@@ -22,19 +22,19 @@ PROMPT                = ""
 
 # These file paths assume that you have done `make install` after
 # compiling; these modules are found in `/usr/local/lib/opencog`.
-MODULES               = opencog/modules/libbuiltinreqs.so,
-                        opencog/modules/libscheme-shell.so,
-                        opencog/modules/libpy-shell.so,
-                        opencog/modules/libattention-types.so,
-                        opencog/modules/libembodiment-types.so,
-                        opencog/modules/libnlp-types.so,
-                        opencog/modules/libspacetime-types.so,
-                        opencog/modules/libLGDictModule.so,
-                        opencog/modules/libPersistModule.so,
-                        opencog/modules/libPythonModule.so,
-                        opencog/modules/libQueryModule.so,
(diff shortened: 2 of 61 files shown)


linas:
Holy cow! I did not think of `make examples` sorry.

lib/cogserver.conf
```diff
@@ -73,29 +73,29 @@ ANSI_ENABLED	       = true
...
+                        opencog/cogserver/modules/libPersistZmqModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libLGDictModule.so,
+                        opencog/cogserver/modules/libSuRealModule.so,
+                        opencog/cogserver/modules/libattention.so,
+                        opencog/cogserver/modules/libAtomSpaceExtensions.so,
+                        opencog/cogserver/modules/libPythonModule.so,
+                        opencog/cogserver/modules/libpy-shell.so,
+                        opencog/cogserver/modules/libscheme-shell.so,
+                        opencog/cogserver/modules/libRuleEngineModule.so
```
linas: This is not right. Modules are installed in the same location as before: `/usr/local/lib/opencog/modules`  The above change will cause the modules to not be found; the cogserver will just fail to start, I think.

lib/opencog-chatbot.conf
```diff
@@ -22,19 +22,19 @@ PROMPT                = ""
...
+                        opencog/cogserver/modules/libattention-types.so,
+                        opencog/cogserver/modules/libembodiment-types.so,
+                        opencog/cogserver/modules/libnlp-types.so,
+                        opencog/cogserver/modules/libspacetime-types.so,
+                        opencog/cogserver/modules/libLGDictModule.so,
+                        opencog/cogserver/modules/libPersistModule.so,
+                        opencog/cogserver/modules/libPythonModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libRuleEngineModule.so,
+                        opencog/cogserver/modules/libSuRealModule.so
```
linas: same remark as above. Doing this will bork the chatbot.

opencog/learning/PatternMiner/DistributedPatternMinerClient.cc
```diff
@@ -23,8 +23,8 @@
  */
 
 #include "DistributedPatternMinerClient.h"
-#include <opencog/server/CogServer.h>
-#include <opencog/server/Factory.h>
+#include <opencog/cogserver/server/CogServer.h>
+#include <opencog/cogserver/server/Factory.h>
 #include <opencog/util/Logger.h>
```
linas: strange .. is this not built by default? I did not see this break, when I made the changes.

opencog/nlp/learn/misc-scripts/opencog-en.conf
```diff
@@ -32,11 +32,11 @@ ANSI_SCM_PROMPT       = "[0;34mguile-en[1;34m> [0m"
...
-MODULES               = opencog/modules/libbuiltinreqs.so,
-                        opencog/modules/libPersistModule.so,
-                        opencog/modules/libQueryModule.so,
-                        opencog/modules/libscheme-shell.so,
-                        opencog/modules/libnlp-types.so
+MODULES               = opencog/cogserver/modules/libbuiltinreqs.so,
+                        opencog/cogserver/modules/libPersistModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libscheme-shell.so,
+                        opencog/cogserver/modules/libnlp-types.so
```
linas: Please don't!  Seriously, this completely breaks the server!  You need to test before making these changes.

scripts/createSymLinksOCPKG.sh
```diff
@@ -95,17 +95,17 @@ if ! [ -z $GITROOT ]; then
 		echo "Creating folders for symbolic links..."
 
 		# Let's create the folders first
-		mkdir -p bin/opencog/server
+		mkdir -p bin/opencog/cogserver/server
```
linas: "bin" ??? do you mean "build"?   There is no "bin" directory, there is a build directory, where the source code gets built.

lib/cogserver.conf
AmeBel: no, it will load

scripts/createSymLinksOCPKG.sh
AmeBel: previously bin used to be used. actually this script must be removed

lib/opencog-chatbot.conf
AmeBel: i haven't tried installing it and running the cogserver, so yes their might be some issue. 

I will make a pr for https://github.com/opencog/opencog/issues/1968, I am testing now

lib/cogserver.conf
AmeBel: ok it works when running from the build directory but not on installation. Reverse of the scheme loading issue

AmeBel: I guess the optonal fix is to use `opencog/cogserver/modules` for running from build directory and for tests and use `opencog/modules` for installation. . I wil add a separate conf for developement.

linas: argh. Right. what a mess. When I run "production" servers, e.g. for the behavior trees, the build directory is nowhere nearby, so everything has to run from the install dir.   For unit testing, it should test what is in the build dir, else things get confusing.

linas:
How did you test this stuff?  Docker only?  Did you try to test it in a stand-alone environment?  I spent about the last 8 hours basically undoing much of what this commit does, piece-meal, because basically, nothing at all worked for me. I had to debug a dozen different scripts that all got broken by these changes.  

What's more, most of these changes don't even make sense: they point to directories that simply don't exist, on my system.   When you test,  you realize that you need to clean out the contents of /usr/local so that old files located in crazy old install locations are gone, right?  Cause otherwise, its like chasing ghosts.

lib/opencog-chatbot.conf
linas: well, and it did bork the chatbot, and it took me all day to find this, and all the other breakages.  I'm kind of miffed that this pull request was merged, despite the line notes that said "don't merge this".

AmeBel:
- Yes I test in docker. What do you mean in a stand-alone environment?
- The undoing was already done in  https://github.com/opencog/opencog/pull/1969, you aproved. As I said  in the #1969 I might have missed something. 
- Yes, the changes don't make sense if you look at them in the install paths that `opencog.conf` and `cogserver.conf` use. But, from the build directory it makes total sense. In any case #1969 fixed the issues related with opencog.conf and cogserver.conf. What I didn't know was that `opencog-chatbot.conf` was being installed, and there is no way of telling that from https://github.com/opencog/opencog/blob/master/lib/CMakeLists.txt#L19, so didn't revert the changes in opencog-chatbot.conf assuming it was being used from build directory. Maybe a wrong assumption, what am I missing? In addition, so as to avoid conflict with you and other developers as well as make it easier for noobs I added [this](https://github.com/opencog/opencog/blob/master/lib/development.conf)  and [this](https://github.com/opencog/opencog/blob/master/opencog/cogserver/scm/config.scm)
- I suppose you do have to clean contents in /usr/local, but I am always using `docker-compose run --rm dev` which  effectively means I start from a clean cogutils and atomspace installs every time I start work. And when there are any breaking or significant changes in atomspace or cogutils I rebuild every image, with the exception of the base dependency image `opencog/opencog-deps`. And rebuild the `opencog/opencog-dev:cli` image (aka the development image for this repo) atleast once a week.

From you experience, what additional step do you suggest I add to minimize conflict with you and other developers?

AmeBel:
Is there an integration test for the chatbot that one could use for testing the whole pipeline? Not that I am making changes to the chatbot, just curious.

linas:
The chatbot(s) are under development, its too early to do testing, except of a "hello world" type nature.

linas:
Still confused. Just now, I went to the build directory, and started the cogserver from the bash prompt, as `opencog/server/cogserver` without specifying any config file.  Normally, this would cause it to pick some default config file, after which the cogserver would run.  Whatever it is doing, it is failing to load up the modules, either from `/usr/local/lib/opencog` or from the build directory: the default config file is unable to find either set of .so modules.  I don't care which it uses, since I did a `make install` prior to starting the cogserver.

linas:
I'm opening a different bug to diiscuss this, since I don't understand what I'm witnessing.

PR #1985, opened 2016-01-07
Use DefinedSchemaNode in PLN rule names.

linas:
This is the other half of pull req opencog/atomspace#578


ngeiswei:
Oh, too bad, I was about to commit that as well.

PR #1988, opened 2016-01-10
r2l rules use ure function for definition

AmeBel:


.gitignore
```diff
@@ -1,6 +1,5 @@
 *.a
-opencog.log
-opencog-python.log
+*.log
```
linas: It would be nice if we stopped writing out forward-chainer.log. Its always huge, and it pollutes the filesystem.

lib/development.conf
```diff
@@ -97,45 +97,52 @@ MODULES               = opencog/cogserver/server/libbuiltinreqs.so,
...
+                        nlp/scm/file-utils.scm,
+                        nlp/scm/nlp-utils.scm,
+                        nlp/scm/disjunct-list.scm,
+                        nlp/scm/processing-utils.scm,
+                        nlp/relex2logic/r2l-utilities.scm,
+                        nlp/relex2logic/rule-utils.scm,
+                        nlp/relex2logic/rule-helpers.scm,
+                        nlp/relex2logic/post-processing.scm,
+                        nlp/lg-dict/lg-dict.scm,
+                        nlp/sureal/surface-realization.scm
```
linas: In the long run, this should probably be replaced by a `(use-modules (opencog) (opencog cogserver) (opencog microplanning) ...etc.)`   I know that this doesn't quite always work as expected, due to various weirdnesses, but that should be the long-term goal.

linas: well, but you already know this, so never mind.

PR #1989, opened 2016-01-11
add function for setting ip/port of relex server

### LINKED ISSUE #1987: using r2l over the network fails
### Steps taken
```scheme
cd /opencog/build
sudo make install                   
cd ../opencog/nlp/relex2logic/      
guile -l run-r2l.scm                
scheme@(guile-user)> (define relex-server-host "172.17.0.2")                           
scheme@(guile-user)> (nlp-parse "this is a test")                                      
ERROR: In procedure connect:                                                           
ERROR: In procedure connect: Connection refused                                        
                                                                                       
```
Previously (before the modulearization) setting `relex-server-host` sufficed when running from the cogserver.

AmeBel:
resolves #1987

### DIFF
--- a/opencog/nlp/scm/nlp.scm
+++ b/opencog/nlp/scm/nlp.scm
@@ -7,6 +7,29 @@
 (define relex-server-host "127.0.0.1")
 (define relex-server-port 4444)
 
+(define-public (relex-server-host! ip)
+"
+  Set the relex-server's ip. By default it is 127.0.0.1.
+
+  ip:
+  - A string specifying the ip
+"
+    (if (not (string? ip)) (error "Expected the ip to be a string, got: " ip))
+    (set! relex-server-host ip)
+)
+
+(define-public (relex-server-port! port)
+"
+  Set the port number for the relex-server. By default it is 4444
+
+  port:
+  - A number specifying the port number
+"
+    (if (not (number? port))
+        (error "Expected the ip to be a number, got:" port))
+    (set! relex-server-port port)
+)
+
 ; Load various parts....
 (load "nlp/nlp-utils.scm")
 (load "nlp/disjunct-list.scm")


opencog/nlp/scm/nlp.scm
```diff
@@ -7,6 +7,29 @@
 (define relex-server-host "127.0.0.1")
 (define relex-server-port 4444)
```
linas: it probably would have been enough to just say `define-public` above, right?  That way, you would not need to create wrapper routines?

AmeBel:
Did try `define-public` didn't work. Just tried `set!` , after reseting to previous commit, doesn't work.

linas:
Can't quite be right.  Try this:   `cd opencog/scm/opencog; vi dummy.scm` and create this file:
```
(define-module (opencog dummy))
(define-public xxx 42)
(define-public (show)  (display "urf ") (display xxx) (newline))
```

Then start guile and run
```
(add-to-load-path  "/whereever/your/opencog/scm")
(use-modules (opencog dummy))
(show)
xxx
(set! xxx 3)
(show)
xxx
```
and it looks like `xxx` is settable just fine.  I just tried it to make sure I'm not crazy.  So: perhaps you are loading `nlp/scm/nlp.scm` after you set `relex-server-host`, so that what you set got clobbered?

AmeBel:
sometimes it works sometimes it doesn't, othertimes you get warning like `WARNING: (opencog nlp chatbot): `r2l-rules' imported from both (opencog rule-engine) and (opencog nlp relex2logic)
` How on earth can `r2l-rules` be imported from `(opencog rule-engine)`? Don't know.

In any case pls send me your .guile file, I will set that as the default in docker containers.

linas:
here is my `~/.guile` but its minimal:
```
;; rlb's method for activiting realine ...
 (use-modules (ice-9 readline))
  (activate-readline)

  (debug-enable 'backtrace)

; Record positions of source code expressions.
  (read-enable 'positions)

(add-to-load-path "/usr/local/share/opencog/scm")
(add-to-load-path ".")
```

Sorry, did not mean to make you go crazy. No clue about the r2l-rules warning. 

Debugging a lot of code is harder than debugging only a small amount of code. Thus, if you can accomplish something in 2 lines of code, you should not write 20.  This commit seems to write 20 lines for something that could be done in 2.  Yes, they are rather "simple", almost trivial lines, but its still more than the minimum.

AmeBel:
thanks and will revert the change.

PR #1992, opened 2016-01-13
Add multi-threaded agent running support to CogServer

hedayat:
This patchset adds the possibility to run agents in multiple threads. One or more agents can be assigned to named threads.

### DIFF
--- a/lib/opencog.conf
+++ b/lib/opencog.conf
@@ -73,22 +73,22 @@ ANSI_ENABLED	       = true
 # Cogserver in OSX will automatically change .so extension to .dylib
 # if .so exists.
 MODULES               = opencog/cogserver/server/libbuiltinreqs.so,
-                        opencog/modules/libPersistModule.so,
-                        opencog/modules/libPersistZmqModule.so,
-                        opencog/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libPersistModule.so,
+                        opencog/cogserver/modules/libPersistZmqModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
                         opencog/cogserver/shell/libscheme-shell.so,
                         opencog/cogserver/shell/libpy-shell.so,
                         opencog/nlp/types/libnlp-types.so,
-                        opencog/modules/libLGDictModule.so,
-                        opencog/modules/libSuRealModule.so,
+                        opencog/cogserver/modules/libLGDictModule.so,
+                        opencog/cogserver/modules/libSuRealModule.so,
                         opencog/learning/pln/libPLNTypes.so,
                         opencog/attention/libattention-types.so,
                         opencog/attention/libattention.so,
                         opencog/embodiment/libembodiment-types.so,
                         opencog/spacetime/libspacetime-types.so,
-                        opencog/modules/python/libPythonModule.so,
-                        opencog/modules/libRuleEngineModule.so,
-                        opencog/dynamics/openpsi/libOpenPsi.so
+                        opencog/cogserver/modules/python/libPythonModule.so,
+                        opencog/cogserver/modules/libRuleEngineModule.so,
+                        opencog/openpsi/libOpenPsi.so
 
 
 # Optional modules, not enabled by default
--- a/opencog/cogserver/server/Agent.cc
+++ b/opencog/cogserver/server/Agent.cc
@@ -121,14 +121,15 @@ stim_t Agent::stimulateAtom(Handle h, stim_t amount)
     logger().fine("Atom %d received stimulus of %d, total now %d",
                   h.value(),
                   amount,
-                  totalStimulus.load());
+                  totalStimulus.load(std::memory_order_relaxed));
 
 #ifdef DEBUG
     std::cout << "Atom " << h.value() << " received stimulus of " << amount <<
-                 ", total now " << totalStimulus << "\n";
+                 ", total now " << totalStimulus.load(std::memory_order_relaxed)
+                 << "\n";
 #endif
 
-    return totalStimulus;
+    return totalStimulus.load(std::memory_order_relaxed);
 }
 
 void Agent::removeAtomStimulus(Handle h)
@@ -170,7 +171,7 @@ stim_t Agent::resetStimulus()
     }
     // reset stimulus counter
     totalStimulus = 0;
-    return totalStimulus;
+    return totalStimulus.load(std::memory_order_relaxed);
 }
 
 stim_t Agent::getTotalStimulus() const
--- a/opencog/cogserver/server/Agent.cc
+++ b/opencog/cogserver/server/Agent.cc
@@ -88,13 +88,17 @@ std::string Agent::to_string() const
 void Agent::atomRemoved(AtomPtr atom)
 {
     Handle h = atom->getHandle();
-    for (size_t i = 0; i < _utilizedHandleSets.size(); i++)
-        _utilizedHandleSets[i].erase(h);
+    {
+        std::lock_guard<std::mutex> lock(_handleSetMutex);
+        for (size_t i = 0; i < _utilizedHandleSets.size(); i++)
+            _utilizedHandleSets[i].erase(h);
+    }
     removeAtomStimulus(h);
 }
 
 void Agent::resetUtilizedHandleSets()
 {
+    std::lock_guard<std::mutex> lock(_handleSetMutex);
     for (size_t i = 0; i < _utilizedHandleSets.size(); i++)
         _utilizedHandleSets[i].clear();
     _utilizedHandleSets.clear();
@@ -103,35 +107,42 @@ void Agent::resetUtilizedHandleSets()
 
 stim_t Agent::stimulateAtom(Handle h, stim_t amount)
 {
-    // Add atom to the map of atoms with stimulus
-    // and add stimulus to it
-    (*stimulatedAtoms)[h] += amount;
+    {
+        std::lock_guard<std::mutex> lock(stimulatedAtomsMutex);
+
+        // Add atom to the map of atoms with stimulus
+        // and add stimulus to it
+        (*stimulatedAtoms)[h] += amount;
+    }
 
     // update record of total stimulus given out
     totalStimulus += amount;
-    
-    logger().fine("Atom %d received stimulus of %d, total now %d", 
-                  h.value(), 
-                  amount, 
-                  totalStimulus);
 
-#ifdef DEBUG    
+    logger().fine("Atom %d received stimulus of %d, total now %d",
+                  h.value(),
+                  amount,
+                  totalStimulus.load());
+
+#ifdef DEBUG
     std::cout << "Atom " << h.value() << " received stimulus of " << amount <<
                  ", total now " << totalStimulus << "\n";
 #endif
-    
+
     return totalStimulus;
 }
 
 void Agent::removeAtomStimulus(Handle h)
 {
     stim_t amount;
-    // if handle not in map then return
-    if (stimulatedAtoms->find(h) == stimulatedAtoms->end())
-        return;
+    {
+        std::lock_guard<std::mutex> lock(stimulatedAtomsMutex);
+        // if handle not in map then return
+        if (stimulatedAtoms->find(h) == stimulatedAtoms->end())
+            return;
 
-    amount = (*stimulatedAtoms)[h];
-    stimulatedAtoms->erase(h);
+        amount = (*stimulatedAtoms)[h];
+        stimulatedAtoms->erase(h);
+    }
 
     // update record of total stimulus given out
     totalStimulus -= amount;
@@ -153,7 +164,10 @@ stim_t Agent::stimulateAtom(HandleSeq hs, stim_t amount)
 
 stim_t Agent::resetStimulus()
 {
-    stimulatedAtoms->clear();
+    {
+        std::lock_guard<std::mutex> lock(stimulatedAtomsMutex);
+        stimulatedAtoms->clear();
+    }
     // reset stimulus counter
     totalStimulus = 0;
     return totalStimulus;
@@ -166,6 +180,7 @@ stim_t Agent::getTotalStimulus() const
 
 stim_t Agent::getAtomStimulus(Handle h) const
 {
+    std::lock_guard<std::mutex> lock(stimulatedAtomsMutex);
     if (stimulatedAtoms->find(h) == stimulatedAtoms->end()) {
         return 0;
     } else {
--- a/opencog/cogserver/server/AgentRunnerBase.cc
+++ b/opencog/cogserver/server/AgentRunnerBase.cc
@@ -20,7 +20,9 @@
 #include <algorithm>
 #include <chrono>
 #include <opencog/cogserver/server/AgentRunnerBase.h>
+#include <opencog/cogserver/server/CogServer.h>
 #include <opencog/util/Logger.h>
+#include <opencog/util/platform.h>
 
 using namespace std;
 using namespace chrono;
@@ -71,6 +73,7 @@ void AgentRunnerBase::removeAgent(AgentPtr a)
     AgentSeq::iterator ai = std::find(agents.begin(), agents.end(), a);
     if (ai != agents.end()) {
         agents.erase(ai);
+        cogserver().systemActivityTable().clearActivity(a);
         a->stop();
         logger().debug("[CogServer::%s] stopped agent \"%s\"", name.c_str(),
             a->to_string().c_str());
@@ -84,7 +87,10 @@ void AgentRunnerBase::removeAllAgents(const std::string& id)
         std::partition(agents.begin(), agents.end(),
                        boost::bind(equal_to_id(), _1, id));
 
-    std::for_each(last, agents.end(), [] (AgentPtr &a) { a->stop(); });
+    std::for_each(last, agents.end(), [] (AgentPtr &a) {
+        cogserver().systemActivityTable().clearActivity(a);
+        a->stop();
+    });
 
     // remove those agents from the main container
     agents.erase(last, agents.end());
@@ -96,25 +102,53 @@ void AgentRunnerBase::removeAllAgents(const std::string& id)
 void AgentRunnerBase::removeAllAgents()
 {
     for (auto &a: agents)
+    {
+        cogserver().systemActivityTable().clearActivity(a);
         a->stop();
+    }
     agents.clear();
     logger().debug("[CogServer::%s] stopped all agents", name.c_str());
 }
 
 void AgentRunnerBase::runAgent(AgentPtr a)
 {
-    auto timer_start = system_clock::now();
+    size_t mem_start = getMemUsage();
+    int atoms_start = atomspace().get_size();
+
     logger().debug("[CogServer::%s] begin to run mind agent: %s, [cycle = %d]",
                    name.c_str(), a->classinfo().id.c_str(), cycleCount);
+
+    auto timer_start = system_clock::now();
 
     a->resetUtilizedHandleSets();
     a->run();
 
     auto timer_end = system_clock::now();
-    auto elapsed = duration_cast<duration<float>>(timer_end - timer_start).count();
-    logger().debug("[CogServer::%s] running mind agent: %s, elapsed time (sec): %f "
-            "[cycle = %d]", name.c_str(), a->classinfo().id.c_str(), elapsed,
+
+    size_t mem_end = getMemUsage();
+    int atoms_end = atomspace().get_size();
+
+    size_t mem_used;
+    if (mem_start > mem_end)
+        mem_used = 0;
+    else
+        mem_used = mem_end - mem_start;
+
+    int atoms_used;
+    if (atoms_start > atoms_end)
+        atoms_used = 0;
+    else
+        atoms_used = atoms_end - atoms_start;
+
+    auto elapsed = timer_end - timer_start;
+    auto elapsed_secs = duration_cast<duration<float>>(elapsed).count();
+    logger().debug("[CogServer::%s] running mind agent: %s, elapsed time (sec): "
+            "%f, memory used: %d, atom used: %d [cycle = %d]", name.c_str(),
+            a->classinfo().id.c_str(), elapsed_secs, mem_used, atoms_used,
             cycleCount);
+
+    cogserver().systemActivityTable().logActivity(a, elapsed, mem_used,
+        atoms_used);
 }
 
 void SimpleRunner::processAgents()
--- a/opencog/cogserver/server/AgentRunnerBase.cc
+++ b/opencog/cogserver/server/AgentRunnerBase.cc
@@ -19,7 +19,7 @@
 
 #include <algorithm>
 #include <chrono>
-#include <opencog/server/AgentRunnerBase.h>
+#include <opencog/cogserver/server/AgentRunnerBase.h>
 #include <opencog/util/Logger.h>
 
 using namespace std;
@@ -119,15 +119,13 @@ void AgentRunnerBase::runAgent(AgentPtr a)
 
 void SimpleRunner::processAgents()
 {
-    if (running) {
-        AgentSeq::const_iterator it;
-        for (it = agents.begin(); it != agents.end(); ++it) {
-            AgentPtr agent = *it;
-            if ((cycleCount % agent->frequency()) == 0)
-                runAgent(agent);
-        }
-        ++cycleCount;
+    AgentSeq::const_iterator it;
+    for (it = agents.begin(); it != agents.end(); ++it) {
+        AgentPtr agent = *it;
+        if ((cycleCount % agent->frequency()) == 0)
+            runAgent(agent);
     }
+    ++cycleCount;
 }
 
 } /* namespace opencog */
--- a/opencog/cogserver/server/AgentRunnerBase.cc
+++ b/opencog/cogserver/server/AgentRunnerBase.cc
@@ -40,7 +40,7 @@ struct equal_to_id: public std::binary_function<AgentPtr, const std::string&, bo
 };
 
 AgentRunnerBase::AgentRunnerBase(std::string runner_name): name(runner_name),
-        cycleCount(0)
+        cycleCount(1)
 {
 }
 
--- a/opencog/cogserver/server/AgentRunnerThread.cc
+++ b/opencog/cogserver/server/AgentRunnerThread.cc
@@ -43,14 +43,14 @@ void AgentRunnerThread::start()
     {
         // locking is required, to prevent dead-lock in condition_variable wait()
         lock_guard<mutex> lock(runAgentsMutex);
-        runAgents = true;
+        runAgents.store(true, memory_order_relaxed);
     }
     runAgentsCond.notify_one();
 }
 
 void AgentRunnerThread::stop()
 {
-    runAgents = false;
+    runAgents.store(false, memory_order_relaxed);
 }
 
 void AgentRunnerThread::addAgent(AgentPtr a)
@@ -71,7 +71,7 @@ void AgentRunnerThread::addAgent(AgentPtr a)
     else {
         lock_guard<mutex> lock(agentsMutex);
         agentsAddQ.push_back(a);
-        agentsModified = true;
+        agentsModified.store(true, memory_order_relaxed);
     }
 }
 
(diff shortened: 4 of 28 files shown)


lib/opencog.conf
```diff
@@ -73,22 +73,22 @@ ANSI_ENABLED	       = true
 # Cogserver in OSX will automatically change .so extension to .dylib
 # if .so exists.
 MODULES               = opencog/cogserver/server/libbuiltinreqs.so,
-                        opencog/modules/libPersistModule.so,
-                        opencog/modules/libPersistZmqModule.so,
-                        opencog/modules/libQueryModule.so,
+                        opencog/cogserver/modules/libPersistModule.so,
+                        opencog/cogserver/modules/libPersistZmqModule.so,
+                        opencog/cogserver/modules/libQueryModule.so,
                         opencog/cogserver/shell/libscheme-shell.so,
```
linas: These changes are not correct, because they only work if you start the cogserver in the build tree. If you don't have a build tree available, then this config file will fail to load these modules.

This is the same problem that Amen is trying to solve by adding symlinks.

opencog/cogserver/server/Agent.cc
```diff
@@ -88,13 +88,17 @@ std::string Agent::to_string() const
...
 {
     Handle h = atom->getHandle();
-    for (size_t i = 0; i < _utilizedHandleSets.size(); i++)
-        _utilizedHandleSets[i].erase(h);
+    {
+        std::lock_guard<std::mutex> lock(_handleSetMutex);
+        for (size_t i = 0; i < _utilizedHandleSets.size(); i++)
+            _utilizedHandleSets[i].erase(h);
+    }
     removeAtomStimulus(h);
```
linas: Although this change is correct, the idea of using notifier chains for this purpose is terrible: it has a large performance impact, and is not fundamentally scalable.   The very concept on which this is built needs to be rethought and redesigned.

linas:
Aftre a quick skim of the code, all of the changes appear to be reasonable.  However, this begs a bigger, more important question:  Wouldn't life be easier if we just remove all of the Agent code completely?   It seems to just be a lot of pointless complexity that does not do anything useful, does not seem to solve any problem that we have, either practical or theoretical.   Why are we maintaining and enhancing code that is useless and pointless?  Why not just remove it entirely?  What am I missing?

pennachin:
There are two components to "old fashioned" Agent processing in the CogServer. One is the round robing scheduler, another is the Agent design. The former definitely should be removed, but at the moment ECAN depends on it. Whether the latter needs to be completely removed is a different question. We need some design to support cognitive processes that run continuously in the background, such as ECAN and background PLN; and we want a design that supports running those processes in parallel, along with "foreground" computing. 

So Ben and I thought it was simpler to proceed in this order:

1. Enable parallel background processes (this PR)
2. Modify ECAN to eliminate the dependency on the cycle and the scheduler.
3. Rethink the Agent design if it seems inadequate for background processes.

linas:
+Cassio OK. Yes, I can see that being able to manage the compute resources devoted to particular tasks in opencog is important.

hedayat:
Thank you for merging, but I was going to remove the wrong commit (module paths), and also work around the GCC 4.8 compatibility. As CircleCI is complaining, the current code doesn't compile with GCC 4.8 as it doesn't have atomic_store/load support for shared_ptrs. I'll try to fix both issues soon though.

linas:
Ah, OK, thank you; I was in a hurry to leave the house, right before
merging, and hadn't looked very carefully.  --linas

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+{
+    removeAllAgents();
+    joinRunThread();
+}
+
+void AgentRunnerThread::start()
+{
+    {
+        // locking is required, to prevent dead-lock in condition_variable wait()
+        lock_guard<mutex> lock(runAgentsMutex);
```
linas: This lock does nothing whatsoever, runAgents is a bool. You even declared it as an atomic bool.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+        runAgents.store(true, memory_order_relaxed);
```
linas: What is the point of relaxing memory order? Especially if you are locking?  The lock forces strict memory order (which isn't needed) and here you have a bool and you are relaxing .. this code does not make sense.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+    }
+    runAgentsCond.notify_one();
+}
+
+void AgentRunnerThread::stop()
+{
+    runAgents.store(false, memory_order_relaxed);
```
linas: relaxing here seems pointless.  Just say `runAgents = false;` -- calling methods on this class doesn't make sense: its a bool!

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+void AgentRunnerThread::addAgent(AgentPtr a)
+{
+    if (!hasAgents()) {
+        joinRunThread();
+        /*
+         * processAgentsThread() will exit immediately if no agents are available,
+         * so we add the first one immediately
+         */
+        {
+            lock_guard<mutex> lock(agentsMutex);
```
linas: This is another mis-use of a lock. You should lock to protect resources.   addAgent() is not a resource, its a method call.  Whatever you think this lock is doing, its not actually doing what you think it is.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+            AgentRunnerBase::addAgent(a);
+        }
+        runThread = thread(&AgentRunnerThread::processAgentsThread, this);
+        // todo: set thread name when using pthreads
+    }
+    else {
+        lock_guard<mutex> lock(agentsMutex);
+        agentsAddQ.push_back(a);
```
linas: This lock is being used correctly, its protecting an actual resource.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+        agentsModified.store(true, memory_order_relaxed);
```
linas: calling this method on the atomic does not make sense.  The relaxation is crazy. Just say `agentsModified = true;`

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+void AgentRunnerThread::removeAgent(AgentPtr a)
+{
+    lock_guard<mutex> lock(agentsMutex);
+    agentsRemoveQ.push_back(a);
+    agentsModified.store(true, memory_order_relaxed);
```
linas: another misuse of atomic

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void AgentRunnerThread::removeAllAgents(const std::string& id)
+{
+    lock_guard<mutex> lock(agentsMutex);
+    idsRemoveQ.push_back(id);
+    agentsModified.store(true, memory_order_relaxed);
```
linas: another misuse of atomic.   It appears you misused atomic everywhere.  You may want to review the concept of what atomic means, and what the point of having atomics is.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void AgentRunnerThread::removeAllAgents()
+{
+    lock_guard<mutex> lock(agentsMutex);
```
linas: I'm fairly sure you don't need this lock.  Based on how use use these two bools, I can't see how there's a race. But I did not look that closely.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+    clearAll = true;
+    agentsModified.store(true, memory_order_relaxed);
+}
+
+const AgentSeq& AgentRunnerThread::getAgents() const
+{
+    lock_guard<mutex> lock(agentsMutex);
```
linas: This lock does absolutely nothing.  Not only that, but the API is fundamentally thread-unsafe.  

The problem here is that by returning a reference to AgentSeq, ... which is a non-atomic vector of non-atomic things, anyone who uses that reference will need to lock it first; however, they don't have access to your locks, and even if they did, they would likely forget to do the lock, because they don't understand what your code does.   Basically, this API allows a protected resource to escape, which is a kind of basic no-no of threaded programming.

The correct way to write this code is to get the lock, make a COPY of the AgentSeq, and then return thee COPY.  Users can then use that copy without being acciedentally clobbered by your code.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+    return agents;
+}
+
+bool AgentRunnerThread::hasAgents() const
+{
+    lock_guard<mutex> lock(agentsMutex);
+    return !agents.empty();
```
linas: this is a correct use of the lock.

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void AgentRunnerThread::processAgentsThread()
+{
+    /*
+     * 'agents' is only modified in this function, so there is no need to
+     * protect read accesses to this container in it.
```
linas: this comment is nonsense. agents is not atomic, and you are modifying it. One cannot safely read something that is not atomic.  You have a bunch of different methods accessing `agents`; you've even locked one of them, and attempted to lock another. So its crazy to not lock the code that actually modifies it!

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+     */
+    logger().debug("[CogServer::%s] Agent thread started", name.c_str());
+    while (!agents.empty()) {
+        if (!runAgents.load(memory_order_relaxed)) {
```
linas: another ongoing misuse of atomics  You have mis-used atomics in every single place where you used them.

opencog/cogserver/server/AgentRunnerThread.cc
linas: Oh, I see, you are using this lock with a condition variable.  This is not obvious from how you named the mutex: you should rename it to something like "runAgentsCondition" or something like that,

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+            unique_lock<mutex> lock(runAgentsMutex);
+            runAgentsCond.wait(lock,
+                [this] {return runAgents.load(memory_order_relaxed);});
```
linas: Ahh, I see what the problem is. You failed to protect agents.empty() with a lock. That is probably because the first time that you tried this, you deadlocked, so you just removed the lock.  But agents is not atomic, so removing that lock is incorrect.

Instead, what you need to do is to lock agentsMutex, then check if the queue is empty. If the ques is non empty, then take the condition lock, then release the agentsMutex, then wait.

Alternately, you could stay racey, and simply call `hasAgents()` instead.   This is no worse than what you do now, and at least has the merit of protecting the vector.

opencog/cogserver/server/AgentRunnerBase.cc
```diff
@@ -0,0 +1,165 @@
...
+    for (auto &a: agents)
+    {
+        cogserver().systemActivityTable().clearActivity(a);
+        a->stop();
+    }
+    agents.clear();
+    logger().debug("[CogServer::%s] stopped all agents", name.c_str());
+}
+
+void AgentRunnerBase::runAgent(AgentPtr a)
```
linas: Please give this method a better name. You use the string runAgent here, and also in runAgentsMutex and also in a boolean called runAgents, and also in a condition called runAgentsCond.   That means a grep for the string "runAgent" has 27 hits, so far, in this patch. That makes it extremely difficult to trace through the code, because of all of the occurrences of the string "runAgent", only 1/4th of them actually refer to this method -- the rest refer to other things.   

alternately you could rename the mutexes the boolean and the contiditon variable -- they don't really have anything to do with this method, and so should not have conflicting names.

opencog/cogserver/server/AgentRunnerBase.cc
```diff
@@ -0,0 +1,165 @@
...
+    size_t mem_start = getMemUsage();
+    int atoms_start = atomspace().get_size();
+
+    logger().debug("[CogServer::%s] begin to run mind agent: %s, [cycle = %d]",
+                   name.c_str(), a->classinfo().id.c_str(), cycleCount);
+
+    auto timer_start = system_clock::now();
+
+    a->resetUtilizedHandleSets();
+    a->run();
```
linas: I'm confused. Are agents threads, now, or are they doing cooperative time-slicing?

opencog/cogserver/server/AgentRunnerBase.cc
```diff
@@ -0,0 +1,165 @@
...
+        atoms_used);
+}
+
+void SimpleRunner::processAgents()
+{
+    AgentSeq::const_iterator it;
+    for (it = agents.begin(); it != agents.end(); ++it) {
+        AgentPtr agent = *it;
+        if ((cycleCount % agent->frequency()) == 0)
+            runAgent(agent);
```
linas: This is weird, too. If agents are now threads, how can you possibly run a thread more than once?  So this seems to imply that agents are performing cooperative time-slicing. 

Or is this some backwards-compatibility interface, where some agents are threads, and others time-slice? 

Perhaps you could change the name of this class, to something like "TimeSlicedAgentRunner" to distinguish it from the "ThreadedAgentRunner" ...?

opencog/cogserver/server/AgentRunnerThread.cc
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+
+        AgentSeq::const_iterator it;
+        for (it = agents.begin(); it != agents.end(); ++it) {
+            AgentPtr agent = *it;
+            // todo: can frequency() be still useful?
+//            if ((cycleCount % agent->frequency()) == 0)
+            runAgent(agent);
```
linas: OK, this is weird, and my head explodes at this point.  

When an agent is added, using `AgentRunnerThread::addAgent()` you create a new thread, and call this function, `processAgentsThread` as it's body.   I am assuming that the correspondence between threads and agents is one-to-one -- for each thread, there is exactly one agent.  Right?  ??? How could it not be?  so why in the world would you loop over *all* agents, here?  Shouldn't you be working only with that one agent that is yours, and ignore all the other agents running in the other threads? They're none of your business ... why are you messing with them?

And its even more confusing, in yet other ways ... Up above, you have a big while-loop, `while (!agents.empty()) ` and here, we are still inside of that while loop.  Now, given that the threads have all started, and given that they are all stuck waiting in the condition variable, then what, exactly is the intent here?    I mean, fist, you have to unstick the condition variable, so that each thread can move beyond the cond-wait up above.  But, once unstuck, each thread will run.

Ohhh ... I am starting to see... Now, the cond-wait means that only ONE thread can run at a time.  Since only the one holding the conditiion variable can possibly run!  ALL the other threads are waiting, helpless!  Yet here, you call the run method on each agent.  

Now, *if* the agent's run() method is written in such a way that it returns, THEN this loop will advance. Otherwise, this loop will deadlock, because typically, the run() method will never return until that agent is done, and that might be never.  So the only way to not deadlock here is to have each timeslice each agent. 

But this completely defeats the whole purpose of having threads to begin with. You create a bunch of threads -- one for each agent. Then you use a condition variable to make sure that no more than one thread can run at a time. Then when that thread runs, it tries to multi-task the agents. 

Why would you want to create a bunch of threads, and then force them to run single-threaded?   Now, maybe I misunderstand something, but I looked at this code over and over and, from what I can tell, it is doing exactly what I say, up above.  This design makes no sense at all. Its deeply, fundamentally flawed...

opencog/cogserver/server/AgentRunnerBase.cc
hedayat: Both is possible. And, by default, agents are run like before: cooperative time-slicing in the main server thread. However, it is possible to run agents in dedicated threads too. These threads might be named, in which case it is possible to assign more than one agent to a single worker thread, and it'll run all of them cooperatively (however, it ignores the 'frequency' parameter: all of them run in each cycle). 

Also note that an agent's run() method is still supposed to run a single cycle of agent process, like before.

opencog/cogserver/server/AgentRunnerBase.cc
hedayat: Yes, this "SimpleRunner" (Yes, the name is not appropriate, I was looking for an alternative name too. And, your suggestion seems a good one) is the cooperative time-sliced runner which behaves like before and runs in the main thread. It is the compatibility interface.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: I use relaxed memory order, because it is already enforced by the lock. so there is no need to re-enforce it here.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: It is an atomic_bool, and I wanted it to be atomic here but I don't want to enforce any memory order here. I think nobody would care if the thread is stopped slightly later. `runAgents = false;` is store() with strict memory order.

opencog/cogserver/server/AgentRunnerThread.cc
linas: You are missing the point. Just say `runAgents = true;` instead of saying `runAgents.store(true, memory_order_relaxed);` -- you should do that here, and everywhere else in the code. The point is, `runAgents` is a boolean, and you don't need to give the compiler hints about memory ordering; it will do the right thing, here.  All that the long form does is make the code hard to read, hard to understand.

opencog/cogserver/server/AgentRunnerThread.cc
linas: You're over-engineering this. This is not a device driver with hard-real-time performance constraints, that require you to save one CPU pipeline stall by relaxing memory ordering, because you've performed a dependency analysis that indicates that out-of-order loading is OK.

Pipeline stalls on a 2GHz machine are exactly 0.5 nanoseconds long.  Don't try to (maybe possibly) gain 0.5 nanoseconds of performance in exchange for making the code hard to read.   That single pipeline stall is wildly overwhelmed by everything else going on in this code.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: AgentRunnerBase is not thread safe. And AgentRunnerBase::addAgent(a) is adding a to 'agents' vector. You say addAgent() is not a resource, how push_back() is a resource then?

Even if the code doesn't 'look good', IMHO the code is actually doing what I think!

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: I wanted to avoid re-enforcing memory order when it is already enforced by the lock().

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Maybe! But I'm pretty sure that when an atomic is used inside a lock, it doesn't need to enforce any memory order itself because it is already enforced (unless it is really needed to re-enforce it inside the protected block of code).

From what I understand about atomic concept, it is OK to want atomic behavior without enforcing any memory order. It depends on the use case.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: I'll check again. The code structure was changed once, so this might be no longer needed (instead, a more strict memory order will be needed for agentsModified).

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Yes, you are right. I'll fix that. Changing the return type to make it non-reference should be enough.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Probably the comment should be written more clearly. But notice that the last 'it' refers to 'this function': I meant that in processAgentsThread() there is no need to protect read accesses to agnts. I don't refer to other methods.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Can you explain what's wrong with it? I think I don't need to strict memory order here.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: I didn't want to protect agents.empty(), as I don't like to lock a mutex each time while condition is being checked. As nobody else is modifying 'agents', I think it is safe to read it without protection in this function although it is not atomic. Can you explain why agents need protection while no other method will change it when this code is being run?

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Yes, I agree that it makes code harder to read & understand. If it really doesn't matter, I'll remove use of explicit memory order parameters. IIRC, I should revert a single commit, as I replaced simple use of atomic_bool variables with store()/load() calls after getting the code to a working state.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: I'm really sorry. You've misunderstood my code, and it is my fault. I've forgot to document my classes, so their purpose is not clear. I'll add some documentation to them, specially AgentRunnerThread to know what you are going to see from the beginning. I'm really sorry.

AgentRunnerThread manages a single worker thread. There is an instance of AgentRunnerThread per worker thread: see CogServer::agentThreads object. And this is why this class only creates a single thread when the first agent is added. If more agents are added, they are supposed to run cooperatively with the first one. An agent's run() method is still supposed to run a single processing cycle, both for backwards compatibility and to make sure that we can do a graceful shutdown. Yes, if it doesn't return, the other agent's would not run. Put such agents in a dedicated thread without any other agents, and this is the default behavior of CogServer::startAgent() if you don't pass any thread names or pass a unique thread name.

The condition variable is used to control if we should run agents in this thread, or we should do nothing. It is used to stop processing when AgentRunnerThread::stop() is called.

opencog/cogserver/server/AgentRunnerThread.cc
linas: Why is AgentRunnerBase not thread safe? Don't you think it would be wise to define it so that its thread safe?

Neither addAgent() nor push_back() are resources. An std::vector is a resource. Please review the concept of a resource, here:   https://en.wikipedia.org/wiki/System_resource

and how resources can be protected with locks: https://en.wikipedia.org/wiki/Lock_(computer_science)

If you are not thinking about resources and locks, you are thinking about the problem incorrectly.

opencog/cogserver/server/AgentRunnerThread.cc
linas: thinking about memory ordering is important only when you are writing to I/O space, such as the PCI bus or the USB bus, and it is important for you to have the bytes appear on the wire in correct order. However, we are not writing to any bus here.   The only other place where memory order is a concern is when you have extremely performance-critical code, and absolutely every CPU cycle counts, and you already know that your code has a pipeline stall in it with the current arrangement of fences, and you want to remove a memory fence in order to save that pipeline stall.  It is complete nonsense to think about how to possibly, maybe, save one single cpu cycle in this block of code.

Please review the basic concept here:  https://en.wikipedia.org/wiki/Memory_barrier

opencog/cogserver/server/AgentRunnerThread.cc
linas: No. The logic you are applying is incorrect, for this situation.

opencog/cogserver/server/AgentRunnerThread.cc
linas: agents is a non-atomic resource. All non-atomic resources must be protected by locks. Its as simple as that.  Just because you don't want to, or don't feel like it, or don't enjoy doing it is not a valid reason to violate a very basic, fundamental rule of multi-threaded programming.

What makes you think that "no other method will change it when this code is being run?" a simple grep of this patch shows multiple uses of agents.push_back and agents.insert, which are not protected by locks. In particular, `AgentRunnerBase::addAgent` modifies `agents` without a lock, and there is no gaurantee whatsoever that some other thread might not be calling that method while you are doing whatever you are doing here.   

The use of the locks here is fundamentally broken.  You are doing voodoo programming. There is no way to randomly insert locks in random places, and hope that magically, things might work.  It won't.  Its broken in a very deep and basic way.

opencog/cogserver/server/AgentRunnerThread.cc
linas: The code is completely broken. You can't argue your way out of this, it has dozens of serious, fundamental design flaws.  This is a complete failure.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: AgentRunnerBase is not thread-safe, because it is also used by SimpleRunner too, which is not multi-threaded. I didn't want to add the overhead of locking there, so that my implementation doesn't affect the current behavior of cogserver for agents who doesn't use dedicated threads.

I know what a resource is, and I'm protecting AgentRunnerBase part of 'this' here. Or more accurately, I'm protecting 'agents'. It it makes you feel better, I can replace calling AgentRunnerBase::addAgent() with its single line definition, to make it explicit what resource I'm protecting. 

I am thinking about resources & locks, but not the way you prefer!

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: Thinking about memory order is important in multi-threaded code when you don't use locks. Here, I use a lock, so using relaxed memory order for agentsModified should not make code harder to understand. I've not tried to write complicated multi-threaded code depending on memory ordering & atomic variables, I just relaxed memory order after writing the code in places that I thought is easy to understand that relaxed memory ordering is fine.

I assumed that agents can be performance critical, so I thought caring about memory order is good when it does not make the code more complicated. If nobody cares, I would remove use of relaxed memory order everywhere.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: As far as I know, reading a non-atomic resource in parallel doesn't need to be protected when there are no writes. Am I wrong?

A simple grep is misleading. Any use of 'agents' outside AgentRunner*.h/cc files is unrelated to this. They are local varaibles.

AgentRunnerBase::addAgent() is a protected member, so no one outside AgentRunnerThread can call it. If you say that in future someone might add some other code in AgentRunnerThread class that can violate this condition, that might make sense. Well, he should not! But if I want to add more protection for such modification in future, I would create a local copy of agents inside processAgentsThread() and copy it to original member when modified. But this code is neither voodoo nor inserting random locks in random places.

opencog/cogserver/server/AgentRunnerThread.cc
hedayat: So you mean that I need consistent memory order here? I don't see it yet.

pennachin:
@linas a clarification on the design for this PR, which seems to have confused you. Agents are not threads. The AgentRunnerThread can run one or more agents. One could have a one-to-one mapping, but ECAN, for instance, is currently designed as set of agents designed to run in sequence. So it's up to people using this code to decide on the mapping of agents to worker threads.

Every agent's run method is supposed to return within a relatively short time. This is the existing "contract" for MindAgents. While we all seem to agree that the MindAgent design should be improved, this PR is agnostic in that respect. It just removes the restriction that all agents are run in a single thread via collaborative timeslicing by enabling the worker threads. As I explained a month ago, we punted on the MindAgent redesign, but didn't see that as a reason to punt on parallel MindAgent execution. 

Given these goals, I don't see how this PR, which you merged, is a complete failure. Sure, there's useful ways to improve it, but it seems to do its job.

PR #1993, opened 2016-01-14
Proposal: Standardizing how new scheme modules are added.

AmeBel:
**Why?** 
1. To avoid having to install modulerized scheme code for every change during developement
2. To avoid having to choose between two ways of defining modules (specifically their installation path)
  * in `/repo-name/openocg/scm/some-module.scm`, or
  * in `/repo-name/opencog/some-module/some-module.scm` .

3 To start using `(use-modules (opencog some-module))` for loading scheme code instead of having [this](https://github.com/AmeBel/opencog/blob/master/lib/development.conf#L114-L142). (Note: this doesn't address https://github.com/opencog/atomspace/issues/508)

**Solution Overivew**
1. Roses standardeized way of defining new packages was used as an example approach.
2. Directory structure for defining new scheme module have two options
  * `/repo-name/opencog/some-module/some-module.scm` imported by `(use-modules (opencog some-module ))`
  * `/repo-name/opencog/some-module/scm/some-module.scm` imported by `(use-modules (opencog some-module))`. 
  * `/repo-name/opencog/some-module/some-module/sub-module.scm` imported by `(use-modules (opencog some-module sub-module))`
  * `/repo-name/opencog/some-module/scm/some-module/sub-module.scm` imported by `(use-modules (opencog some-module sub-module))`. This approach is basically for cases where there are python/c++ modules. If the `sub-module` has a python/c++ code adding`scm` directory under it isn't supported, because we don't many levels deep of submodules.

3 During cmake  configuration stage, the files added as modules are,
  * symlinked in the build directory such that it would have the directory structure to allow import from the cogserver scheme-shell after running `(add-to-load-path "/path/to/build/directory/opencog/scm")` as if the they were installed.
  * configured for installation in the appropriate directory depending on their names

4 The relex2logic module have been used as an example case.

**If accepted then what?**
* Restructure the rest of the scheme modules in this and atomspace repo.
* Do the same for python modules.
* Document it on the wiki and add examples.

### DIFF
--- a/lib/OpenCogFunctions.cmake
+++ b/lib/OpenCogFunctions.cmake
@@ -12,8 +12,10 @@
 # 4. Everything should be explicit
 
 FUNCTION(ADD_GUILE_MODULE MODULE_FILE)
-    # MODULE_FILE = The name of the file that defines the module
-    MESSAGE("^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^")
+# MODULE_FILE: The name of the file that defines the module. It has the same
+#       name as the directory it is in, or the name of the parent directory of
+#       current directory if it is in a folder named 'scm' and used to define
+#       the module.
 
     # Check if the file exists
     IF(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${MODULE_FILE})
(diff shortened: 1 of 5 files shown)


AmeBel:
@linas feedback please.

lib/OpenCogFunctions.cmake
```diff
@@ -0,0 +1,88 @@
...
+        SET(MODULE_FILE_DIR_PATH ${CMAKE_MATCH_2})
+        SET(MODULE_FILES_DIR_PATH ${CMAKE_MATCH_2}/${CMAKE_MATCH_3})
+        SET(GUILE_SYMLINK_DIR "${CMAKE_BINARY_DIR}/opencog/scm")
+        SET(GUILE_INSTALL_DIR "${DATADIR}/scm")
+
+        # Create symlinks in build directory mirroring the install path structure.
+        # Also configure for install.
+        IF ("${MODULE_NAME}.scm" STREQUAL "${MODULE_FILE}")
+            EXECUTE_PROCESS(
+                COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/opencog/scm/${MODULE_FILE_DIR_PATH}
```
linas: you can just say `make_directory ${GUILE_SYMLINK_DIR}/${MODULE_FILE_DIR_PATH}`

lib/OpenCogFunctions.cmake
```diff
@@ -0,0 +1,88 @@
...
+                COMMAND ${CMAKE_COMMAND} -E create_symlink "${CMAKE_CURRENT_SOURCE_DIR}/${MODULE_FILE}" "${GUILE_SYMLINK_DIR}/${MODULE_FILE_DIR_PATH}/${MODULE_FILE}"
```
linas: You probably don't need the `${MODULE_FILE}` at the very very end; this is implicit.

lib/OpenCogFunctions.cmake
```diff
@@ -0,0 +1,88 @@
...
+            )
+            INSTALL(FILES
+                ${MODULE_FILE}
+                DESTINATION "${GUILE_INSTALL_DIR}/${MODULE_FILE_DIR_PATH}"
+            )
+        ELSE()
```
linas: The IF and the ELSE parts seem to be identical...!?

linas:
I like the idea of symlinks, that solves many problems.  

I am confused by the discussion of paths and modules.   A `blah.scm`  is a module if and only if it has `(define-module blah)` inside of it, *and* can be found in the load-path at an appropriate location.  Thus, for example, `utilities.scm` is not a module; it cannot be loaded by saying `(use-modules (opencog utilities))`  and the functions if might define are NOT public, unless explicitly made public or exported.   This is especially the case for the relex2logic rules files... which you can't load by saying `(use-modules (opencog nlp relex2logic rules adverbialpp))` or something like that.

linas:
Other than that, I'm fine with this; the symlink idea alone is worth it.

lib/OpenCogFunctions.cmake
AmeBel: yap, missed that.

lib/OpenCogFunctions.cmake
AmeBel: The IF part is for `foo.scm ` in /path/opencog/foo/foo.scm` where `(define-moudle (opencog foo)) should be present in and thus should be installed/symlinked to `/path/opencog/` directory, for imports to work,  while the ELSE part is for `/path/opencog/foo/bar.scm` that will be installed/symlinked in a similar directory path.

AmeBel:
You are right, my deiscription wasn't complete. What I meant to say was as you described.

lib/OpenCogFunctions.cmake
AmeBel: It is needed

lib/OpenCogFunctions.cmake
AmeBel: Thus there would be no need for https://github.com/opencog/opencog/tree/master/opencog/scm, each module can be self-contained in a directory.

To deal with https://github.com/opencog/opencog/blob/master/opencog/scm/opencog/atom-types.scm though, maybe each module that define a new type should be a part of the module definition. For example nlp types should be part of `(define-module (opencog nlp))`

linas: ah, I see, there was a difference of one S in the text: `MODULE_FILE_DIR_PATH` vs. `MODULE_FILES_DIR_PATH`   and that one S was .. almost invisible.  Could you change the names to something like MODULE_FILE_PATH and MODULE_DIR_PATH  or anything, to make it easier to see that they're different?

AmeBel: sure

PR #1996, opened 2016-01-17
Revert "Fix module paths"

hedayat:
This reverts commit c583d64235dc6907209864174fd359693caf5f06.

These module paths are OK for running from build tree, but not correct for installed version.


AmeBel:
Use https://github.com/opencog/opencog/blob/master/lib/development.conf for running during development

hedayat:
Thanks!

linas:
Amen, can you merge or not as apprriate? I will have irregular
internet/github access for the next week.

--linas

AmeBel:
done

PR #2001, opened 2016-01-20
Remove depth from fuzzy

leungmanhin:
This is required after opencog/atomspace#594 is merged

### DIFF
--- a/opencog/nlp/fuzzy/Fuzzy.cc
+++ b/opencog/nlp/fuzzy/Fuzzy.cc
@@ -107,7 +107,7 @@ bool Fuzzy::try_match(const Handle& soln)
 
     // Check if this is the type of atom that we want
     if (soln->getType() != rtn_type)
-        return false;
+        return true;
 
     HandleSeq soln_nodes;
     get_all_nodes(soln, soln_nodes);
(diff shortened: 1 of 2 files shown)


opencog/nlp/fuzzy/Fuzzy.cc
```diff
@@ -101,20 +101,13 @@ static void get_all_nodes(const Handle& h, HandleSeq& node_list)
...
-    // solutions that are exactly the same size as the target.
-    // Why? I dunno. Might not a similar tree be slightly bigger or
-    // smaller?  XXX Maybe FIXME ?
-    if (0 < depth) return true;
-    if (0 > depth) return false;
-
     // Check if this is the type of atom that we want
     if (soln->getType() != rtn_type)
-        return false;
+        return true;
```
linas: Are you sure that this true-false change is correct??

linas:
Oops, I did not mean to merge this, I thought opencog/atomspace#594 was already merged

opencog/nlp/fuzzy/Fuzzy.cc
leungmanhin: Ya returning true just to let `explore()` to keep exploring the incoming set, meanwhile it will not be added to `solns` so we won't be seeing it in the final results

PR #2004, opened 2016-01-27
fix guile issues when running cogserver from build directory using development.conf

AmeBel:
also fixes anaphora failure


AmeBel:
what

AmeBel:
retest this please

PR #2011, opened 2016-01-30
Remove deprecated atomspace get_outgoing and get_incoming functions

inflector:
Note: this should be merged shortly after https://github.com/opencog/atomspace/pull/610 is merged. The CI tests won't pass until the corresponding opencog/atomspace changes are merged.

### DIFF
--- a/opencog/attention/HebbianUpdatingAgent.cc
+++ b/opencog/attention/HebbianUpdatingAgent.cc
@@ -110,7 +110,7 @@ void HebbianUpdatingAgent::hebbianUpdatingUpdate()
 		bool isDifferent = false;
 
         // get out going set
-        HandleSeq outgoing = a->get_outgoing(h);
+        HandleSeq outgoing = h->getOutgoingSet();
         new_tc = targetConjunction(outgoing);
         // old link strength decays
         old_tc = h->getTruthValue()->getMean();
--- a/opencog/attention/ImportanceDiffusionAgent.cc
+++ b/opencog/attention/ImportanceDiffusionAgent.cc
@@ -152,7 +152,7 @@ int ImportanceDiffusionAgent::makeDiffusionAtomsMap(std::map<Handle,int> &diffus
             continue;
         }
 
-        targets = a->get_outgoing(linkHandle);
+        targets = linkHandle->getOutgoingSet();
 
         for (targetItr = targets.begin(); targetItr != targets.end();
                 ++targetItr) {
@@ -235,7 +235,7 @@ void ImportanceDiffusionAgent::makeConnectionMatrix(bmatrix* &connections_,
         //val *= diffuseTemperature;
         type = a->get_type(*hi); 
 
-        targets = a->get_outgoing(*hi);
+        targets = (*hi)->getOutgoingSet();
         if (classserver().isA(type,ORDERED_LINK)) {
             Handle sourceHandle;
 
--- a/opencog/attention/ImportanceSpreadingAgent.cc
+++ b/opencog/attention/ImportanceSpreadingAgent.cc
@@ -155,7 +155,7 @@ int ImportanceSpreadingAgent::sumDifference(Handle source, Handle link)
     // Get outgoing set and sum difference for all non source atoms
     linkWeight = a->get_TV(link)->toFloat();
     sourceSTI = a->get_STI(source);
-    targets = a->get_outgoing(link);
+    targets = link->getOutgoingSet();
 
     if (a->get_type(link) == INVERSE_HEBBIAN_LINK) {
         for (t = targets.begin(); t != targets.end(); ++t) {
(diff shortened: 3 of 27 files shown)


inflector:
retest this please

inflector:
Jenkins, retest this please

inflector:
@AmeBel The build checks failed.

Compilation fails because it can't find a method I just added to Atom that was previously only defined in Node. 

```
/opencog/opencog/learning/dimensionalembedding/DimEmbedModule.cc: In member function ‘void opencog::DimEmbedModule::addPivot(opencog::Handle, opencog::Type, bool)’:
/opencog/opencog/learning/dimensionalembedding/DimEmbedModule.cc:269:41: error: ‘class opencog::Atom’ has no member named ‘getOutgoingSet’
             HandleSeq newNodes = (*it)->getOutgoingSet();
```

Any idea why the CircleCI build doesn't find `getOutgoingSet` which is definitely in the opencog/atomspace repository for master:

https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Atom.h#L202

It's acting like it didn't do a `make install` of opencog/atomspace. So is there a time delay between when code is merged into master for atomspace and when it is available for use / testing in opencog/opencog?

Does the CircleCI build system rely on an auto-build of the Docker containers that get pushed to the opencog repository on docker.io that already have cogutils and atomspace installed? Or does CircleCI pull from git and  `make && make install` cogutils and then pull from git and `make && make install` atomspace first before building opencog/opencog?

inflector:
@AmeBel The run for the docker container:
```
docker run -v $PWD:/opencog -e PYTHONPATH=/usr/local/share/opencog/python:/opencog/opencog/python/:/opencog/build/opencog/cython:/opencog/opencog/nlp/anaphora
opencog/opencog-dev:cli /bin/sh -c "cd /opencog && /tmp/octool -beti"
```
calls `octool -beti` which does not install cogutils nor atomspace. Shouldn't it really be:
```
-c "cd /opencog && /tmp/octool -cabeti"
```
?

How does the opencog/opencog-dev:cli container image get rebuilt when there are changes to cogutils or atomspace?

inflector:
Jenkins, retest this please

inflector:
@AmeBel The build did finally succeed, so I assume some bot somewhere built new images from the atomspace repository. Does a github trigger run docker-build.sh (or another script) and push to hub.docker.com? 

For CircleCI and Jenkins, there ought to be a way of checking the date of the image and then invoking the additional `'c'` and `'a'` options for `octool` when the last commit for atomspace is later than when the image was last built, or something like that.

AmeBel:
opencog/opencog-dev:cli is autobuilt on docker-hub when there is a merge on
atomspace repo, docker repo, ocpkg repo, cogutils repo and when the
upstream image ubuntu:14.04 is updated, assuming there is no failure in the
chain of image dependencies.

opencog/opencog-dev:cli has atomspace already installed.

AmeBel:
There is already a plan to implement your suggestion of adding the `a` and `c` options once the packaging of binary into deb is done. 

Why wait for packaging? B/c build takes a long time. 

But, it might be possible to configure jenkins to do the install from last atomspace and cogutils build.

inflector:
@AmeBel This means we'll always have breaking CI tests whenever there are cross-dependencies between the projects because of the lag between the time it takes for docker-hub to rebuild the image since the CI tests fire off immediately.

@linas @pennachin Having CI tests that fail for reasons other than an actual problem makes people not trust the tests which means there will be a pressure to merge before the CI tests pass which kind of defeats the purpose of CI. What is the protocol for this? Do we only merge pull requests made by others? Do developers merge their own changes at times but not always?

Why don't we update `octool` to do:
```
wget https://github.com/opencog/cogutils/archive/master.tar.gz
wget https://github.com/opencog/atomspace/archive/master.tar.gz
```
and compare the dates to the docker image date (perhaps indicated by a file touched in docker-build.sh?). If the docker image is out of date, then  `octool` can call `install_cogutil` and / or `install_atomspace` as appropriate before calling `build_source`.

What do you think?

bgoertzel:
BTW Curtis, as a somewhat related comment, we are going to shortly disable CircleCI and use Jenkins for CI for OpenCog.   Please contact wenwei@hansonrobotics.com for info on the Jenkins configuration for OpenCog, and if you have suggestions to make in this regard...

AmeBel:
@inflector good idea on the file touching, that is a a faster way of achieving the same effect of packaging for building images. Thanks for the feedback, will update octool and docker-build.sh to have that flow for use with the ci.

linas:
Heh. Personally, I resigned (was forced to resign) from the opencog project a few months ago, over the CI issue. You are on your own.

inflector:
@bgoertzel Thanks for the heads up. I'll keep an eye on things. I do notice that the Jenkins check seems to be running about 10 times faster which is a really good thing. Something similar to the suggestion I made for `octool` will work for the Jenkins case too.

PR #2017, opened 2016-02-04
Time Space Atom Octomap Array for perception

yantrabuddhi:
Following code is experimental. The purpose is to use the TimeSpaceAtom.h API to create perceptual maps for embodiment and other things like minecraft.


ngeiswei:
@yantrabuddhi could you replace .cpp by .cc? Just to be consistent with the rest of the code...

I see you've added some test file, very good, could turn it into an actually unit test under tests/spacetime? Writing unit tests is an excellent way to test your code and prevent others from eventually breaking it.

opencog/spacetime/TimeSpaceAtomMapsAPI/AtomOcTree.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ *
+ * Copyright (c) 2016, Mandeep Singh Bhatia, OpenCog Foundation
+ * All rights reserved.
+ * License: New BSD
```
ngeiswei: Why do you use a BSD license rather than AGPL?

opencog/spacetime/TimeSpaceAtomMapsAPI/TestTimeSpaceAtom.cpp
```diff
@@ -0,0 +1,167 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+ 
+//TestAtomOcTree.cpp
+#include "assert.h"
```
ngeiswei: `OC_ASSERT` found in `opencog/util/oc_assert.h` is preferred over `assert.h`. That is because, first the failed assertion is logged, second the code calling the code raising the OC_ASSERT may choose to catch the underlying exception and keep running rather than halting.

ngeiswei: Oh, it's a test, so then when you'll turn that into a cxxtest file, use whichever assert is adequate, see

http://cxxtest.com/guide.html#testAssertions

opencog/spacetime/TimeSpaceAtomMapsAPI/TestTimeSpaceAtom.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (ato != UndefinedHandle) {
+    cout << "occupancy probability at " << query << ":\t " << ato<< endl;
+  }
+  else 
+    cout << "occupancy probability at " << query << ":\t is unknown" << endl;    
+}
+
+int main(int argc, char** argv) {
+
+  cout << endl;
```
ngeiswei: Your indentation scheme is a bit inconsistent. Please try to follow

http://wiki.opencog.org/w/Development_standards

yantrabuddhi: Hi Nil,
 Thanks for the review. I will make those changes. I just copy pasted the
license so I made a mistake, I will change the license to gpl.
Thanks,
Mandeep

linas: agpl, not gpl

yantrabuddhi: Hi,
 I just made changes, however there are no CMake File changes, If
everything elese is okay I will make a library in cmake and connect the
tests to it. The known thing to change in code is my use of int to hold
reference to the atom in atomspace and constant of Undefined atom
reference. Could you please let me know what to replace these with?
Regards,
Mandeep

ngeiswei:
@yantrabuddhi sorry for the delay I'm looking into that today. I see you've closed it then reopened. I just want to be sure still want  me to review it. Let me know. You may want to update it as well.

yantrabuddhi:
Hi Nil,
 I would like you to review but I haven't made the CMake files yet because
I am unsure of what correct location of files should be and what should i
do for the atom reference(currently int, and 0 for undefined atom - i need
to have a == and != operation and undefined/invalid atom reference). If the
current setup is okay I will make and upload the cmake files, so please
suggest and i'd make those changes for final changes.
Thanks and Regards,
Mandeep

ngeiswei:
Hi,

regarding the correct location, I don't know, maybe the current one is OK.

Regarding atom's reference, can you use Handle?

yantrabuddhi:
Hi @ngeiswei ,
 I think Handle should be fine, I thought i'd ask as I am not too familiar with atomspace library. Thanks. I will check it out and soon make the changes for final review by you.
Regards,
Mandeep

yantrabuddhi:
hi @ngeiswei ,
 I fixed all the issues i guess. can you please review it now and let me know if i need to do anything else. Regards,
Mandeep

ngeiswei:
Cool, I'll do that today.

opencog/spacetime/TimeSpaceAtomMapsAPI/AtomOcTree.cc
```diff
@@ -0,0 +1,81 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+ 
+#include "AtomOcTree.h"
+namespace octomap {
+  // tree implementation  --------------------------------------
+  AtomOcTree::AtomOcTree(double resolution)
```
ngeiswei: the indentation doesn't follow opencog's standard

http://wiki.opencog.org/wikihome/index.php/Development_standards#Formatting

ngeiswei: BTW I've noticed you have an inconsistent mixture of whitespace and tab. If you like to mix whitespace and tab in a consistent way I highly recommend smart-tab.

BTW, what IDE do you use?

opencog/spacetime/TimeSpaceAtomMapsAPI/README.md
```diff
@@ -0,0 +1,38 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+ 
+Needs -std=c++14 flag set. Currently place holder for Atom in int
```
ngeiswei: You mean "Atom is int" I suppose. But this is wrong anyway you should use Handle (I believe you've corrected the code though).

opencog/spacetime/TimeSpaceAtomMapsAPI/TimeSpaceAtom.cc
```diff
@@ -0,0 +1,242 @@
...
+//TimeSpaceAtom.cpp
+#include "TimeSpaceAtom.h"
+//#include "octomap/OcTreeKey.h"
+#include <assert.h>
+
+TimeSpaceAtom::TimeSpaceAtom(unsigned int num_time_units,vector<double>map_res_meters):time_circle(num_time_units),created_once(false)
+{
+	map_count=map_res_meters.size();//if zero throw exception
+	int i=0;
+	for_each(map_res_meters.begin(),map_res_meters.end(),[&](double resolution){i++;map_res[i]=resolution;});
```
ngeiswei: Wouldn't it be better to use C++11 range loop?

opencog/spacetime/TimeSpaceAtomMapsAPI/TimeSpaceAtom.cc
```diff
@@ -0,0 +1,242 @@
...
+
+bool TimeSpaceAtom::GetAtomCurrentTime(const int map_handle,const point3d location,aHandle& ato)
+{
+	//
+	assert(created_once);
+	int i=time_circle.capacity()-1;
+	if (time_circle.size()<time_circle.capacity()) i=time_circle.size()-1;
+	assert(time_circle[i].has_map(map_handle));
+	OcTreeNode* result = time_circle[i].map_tree[map_handle].search(location);
+	if (result==NULL) {ato=UndefinedHandle;return false;}
```
ngeiswei: C++11 nullptr is better than NULL, type checking wise.

ngeiswei:
@yantrabuddhi I had a rather shallow review. Please do the correction/simplifications suggested (not just in the places with the comments, across the code) and I'll give a deep review them. Thanks.

yantrabuddhi:
Thanks Nil,
 I will fix the tab-space inconsistencies, i was using geany but will shift
to qt-creator. I will also fix the indentation to opencog standard more
carefully. The c++14 requirement has already been taken care of and c++11
is being used now, so i will change my comment. And I will change NULL to
nullptr.
Regards,
Mandeep

yantrabuddhi:
Hi @ngeiswei ,
 I formatted the code with astyle and qt-creator and some manually for fitting 80 column limit. Changed NULL to nullptr. I am just unsure of 1 thing, in the header the class methods are defined as
class xyz
{
void method() {
    ..
}
}

while in .cc file methods are as 
void 
xyz::method()
{
    ...
}

Is that fine or should it be same in both

yantrabuddhi:
Hi @ngeiswei ,
I manually made all method opening brackets to be in next line as well.
Regards,
Mandeep

opencog/spacetime/TimeSpaceAtomMapsAPI/AtomOcTreeNode.cc
```diff
@@ -0,0 +1,112 @@
...
+    }
+    delete[] children;
+    children = nullptr;
+    return true;
+}
+
+void
+AtomOcTreeNode::expandNode()
+{
+    assert(!this->hasChildren());
```
ngeiswei: It's better to use OC_ASSERT (#include <opencog/utils/oc_assert.h>) because

1. It can be disabled if wanted
2. When an assert fails it doesn't halt the program but throws an exception instead

opencog/spacetime/TimeSpaceAtomMapsAPI/TimeSpaceAtom.cc
```diff
@@ -0,0 +1,316 @@
...
+    );
+
+    curr_time = time_p;
+    curr_duration = duration;
+    created_once = true;
+    return true;
+}
+
+bool
+TimeSpaceAtom::PutAtomAtCurrentTime(const int map_handle,
```
ngeiswei: Could you explain to me this choice, why map_handle are int?

How is opencog gonna represent a map?

yantrabuddhi:
Hi @ngeiswei ,
 The constructor of tsa takes a vector of resolutions of all maps in a time unit(A Time unit can contain several maps from 1 to n for 1 to n resolutions, also tsa has a circular buffer of time units). So the int value represents a particular map in a time unit from 1 to n. There is no direct access to map, so the int index (map_handle) is used to access it within a time unit.

ngeiswei:
Hi @yantrabuddhi, the code looks good to me, except perhaps the multiword identifier policy, which according to the dev standard should be camel for types, snake for the rest

http://wiki.opencog.org/w/Development_standards#Multi-word_identifiers

This is barely followed and I don't know if I want to bother too much people about that. I personally like when things are really polished and consistent, but as it seems not everybody agrees I don't want to be too strict. It certainly isn't a big deal.

Anyway, if you feel like following that multi-id policy go ahead. Otherwise I can merge it now.

yantrabuddhi:
Hi @ngeiswei ,
I should have followed conventions from the start, I guess i was just a bit careless. Some of the files override the virtual functions from base octomap except for the TimeSpaceAtom.h,cc API interface for user. I have changed the API to be according to multi-word identifier policy. 
Thanks

ngeiswei:
It makes sense to keep the parent class formatting conventions for the derived class. It's all looking good. Thanks!

PR #2023, opened 2016-02-07
Duplicate printed output to shell interface

linas:
Partial fix for issue opencog/atomspace#629

This sort-of-ish solves the problem, but not quite.


linas:
This is not really a good long-term solution for the general problem, but am merging anyway, as its sufficient to not block progress in other areas.

PR #2024, opened 2016-02-08
Fix test issues

AmeBel:
Fix test issues that have arisen due to https://github.com/opencog/atomspace/blob/master/opencog/atoms/base/Atom.h#L193 

http://61.92.69.43:8080/job/ci-opencog/111/ had been running for `2 days 20 hr`.

### DIFF
--- a/opencog/attention/SimpleImportanceDiffusionAgent.cc
+++ b/opencog/attention/SimpleImportanceDiffusionAgent.cc
@@ -61,7 +61,7 @@ SimpleImportanceDiffusionAgent::SimpleImportanceDiffusionAgent(CogServer& cs) :
 
     // Provide a logger
     log = NULL;
-    setLogger(new opencog::Logger("SimpleImportanceDiffusionAgent.log", 
+    setLogger(new opencog::Logger("SimpleImportanceDiffusionAgent.log",
                                   Logger::FINE, true));
 }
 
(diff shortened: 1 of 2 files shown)


AmeBel:
@linas pls review

AmeBel:
Since there is a lot of whitespace corrections that my editor auto-enforces, the main changes are at 

https://github.com/opencog/opencog/pull/2024/files#diff-13efb1611dce66770e307e0e666e6067L102
and
https://github.com/opencog/opencog/pull/2024/files#diff-25ca1e456937c700c632156b89ab4ed6L310

opencog/attention/SimpleImportanceDiffusionAgent.cc
```diff
@@ -100,10 +100,12 @@ void SimpleImportanceDiffusionAgent::updateMaxSpreadPercentage() {
         Handle h = resultSet.front();
         h->getIncomingSet(back_inserter(resultSet));
         h = resultSet.front();
-        resultSet = h->getOutgoingSet();
-        h = resultSet.back();
-        resultSet = h->getOutgoingSet();
-        h = resultSet.front();
+        if (h->isLink()) {
```
linas: Not sure why these `isLink()` tests are needed.  Presumably, its already known that the handle is a link, and thus, that asking for its outgoing set is a valid operation.  If there are cases where the handle is not a link, then .. well there's a bug in the code, and this test serves only to hide that bag, make it invisible, make it harder to find.

But whatever. A hack to patch things up for now is OK, I guess. I will merge this.

AmeBel: @misgeatgit see Linas's comment above about a possible bug.

PR #2028, opened 2016-02-11
Add Jenkins job config file

wenwei-dev:
This is the job config file that is currently used in Jenkins CI.

### DIFF
--- a/jenkins_job_config.xml
+++ b/jenkins_job_config.xml
@@ -1,0 +1,145 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project>
+  <actions/>
+  <description>opencog/opencog</description>
+  <displayName>opencog/opencog</displayName>
+  <keepDependencies>false</keepDependencies>
+  <properties>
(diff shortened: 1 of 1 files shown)


AmeBel:
best if all the config files are kept @ https://github.com/opencog/docker

AmeBel:
Does jenkins requires the config to be in root of the repo?

wenwei-dev:
No it doesn't. Any place is okay to keep the config files.

AmeBel:
buildbot config are kept in the docker repo, so I think it would be appropriate to have it there. As the jenkins setup is likely to be dockerized when/if we choose to deploy it on a cloud server.

wenwei-dev:
I see. So I'll close the PRs that I've created just now and pull them to https://github.com/opencog/docker/jenkins.

AmeBel:
great thanks

wenwei-dev:
The config files will go to github.com/opencog/docker/jenkins.

PR #2035, opened 2016-02-12
Build release package

wenwei-dev:


CMakeLists.txt
```diff
@@ -599,6 +599,7 @@ SET(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
 SET(CPACK_PACKAGE_VERSION_MAJOR "0")
 SET(CPACK_PACKAGE_VERSION_MINOR "1")
```
linas: we should probably sets version major to 1

bgoertzel:
Hmmm... FWIW, by my intuition, I think Atomspace itself deserves a version 1.* , but OpenCog (as separate from Atomspace) still feels like it's in version 0.* territory...

AmeBel:
atomspace if 5.0.3 https://github.com/opencog/atomspace/blob/master/opencog/atomspace/version.h

linas:
Heh. There have been 5 major redesigns of how the atomspace works, now. So, yeah.

wenwei-dev:
For simplicity, I'll use the version defined in version.h for atomspace, and cogutil.h for cogutils and leave opencog as it is now, which is 0.1.4. Sounds okay?

CMakeLists.txt
```diff
@@ -644,7 +646,14 @@ SET(CPACK_SOURCE_IGNORE_FILES
 	"^${PROJECT_BINARY_DIR}/opencog/cogserver/server/cogserver$"
 	"^${PROJECT_BINARY_DIR}/examples/server/derivedserver$"
 	"^${PROJECT_BINARY_DIR}/examples/hopfield/hopfield$"
+	"^${PROJECT_SOURCE_DIR}/opencog/python/__init__.py$"
 )
+
+SET(CPACK_PACKAGING_INSTALL_PREFIX "/usr/local")
+SET(CPACK_DEBIAN_PACKAGE_DEPENDS
+    "atomspace (>= 0.1.4)"
```
linas: current atomspace version is 5.0.3 and we should request at least that much ...

PR #2037, opened 2016-02-15
Fix warnings and minimize branching

misgeatgit:


linas:
whose job is it to review this code?

misgeatgit:
@linas it used to be maintained by @shujingke but now since she is not active, I guess no one is here to review it except you and others and also, I didn't change anything that affects the logic of the code. At least that's my intention.

ngeiswei:
Ideally this would replaced by an extension of boost statistics, as done in MOSES, like in

https://github.com/opencog/cogutils/tree/master/opencog/util/boost_ext/accumulators/statistics

This is pretty bad ass, but I guess it's not that important, depending on how much code elegance is valued.

linas:
I don't have time to review this. If the code is abandonned, then it's yours to do with as you please.

PR #2038, opened 2016-02-15
Generate lemmas for predicates in R2L

leungmanhin:


leungmanhin:
Just noticed that some updates to sureal will also be needed to support these lemma predicates, I think it would be better to merge this PR after I've finished updating sureal

linas:
yes.

leungmanhin:
Since R2L should generate those instances anyway, I think this PR is not needed

linas:
specifically, it was pull #2041 that makes these changes un-needed/unwanted.

PR #2041, opened 2016-02-18
Update SuReal for supporting lemma as input

leungmanhin:
This is intended to allow these:
```
(nlp-parse "Mike reads")
(nlp-parse "He eats the cakes")
(nlp-parse "She will be eating those chocolates")
```
If we do:
```
(sureal (Set (Evaluation (Predicate "eat") (ListLink (Concept "he")))))
```
Should return: (("he" "eats"))

Or if we include a tense:
```
(sureal (Set (Evaluation (Predicate "eat") (ListLink (Concept "he"))) (InheritanceLink (PredicateNode "eat") (DefinedLinguisticConceptNode "future_progressive"))))
```
Should return: (("he" "will" "be" "eating" "those" "chocolates"))

More tests will be needed, it seems to work ok for simple cases, I'll also update/expand the unit test/examples accordingly.


linas:
Cool!  Thank you!  The examples should be updated to demo the verb-tense thing, and also the documentation too.

PR #2043, opened 2016-02-19
Check the tense before doing disjunct match in SuReal

leungmanhin:
This is also required for the updates in PR #2041

### DIFF
--- a/opencog/nlp/sureal/SuRealPMCB.cc
+++ b/opencog/nlp/sureal/SuRealPMCB.cc
@@ -336,6 +336,44 @@ bool SuRealPMCB::grounding(const std::map<Handle, Handle> &var_soln, const std::
                         continue;
                     else qChkWords.insert(sWord);
 
+                    // make sure the tense matches
+                    // first get the tense of the solution instance node
+                    // from the InheritanceLink in this form, e.g.
+                    // (InheritanceLink
+                    //    (PredicateNode "eats@111")
+                    //    (DefinedLinguisticConceptNode "present"))
+                    std::string sTense;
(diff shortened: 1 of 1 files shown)


linas:
Is there a way to do this check not in the C++ code, but using the search pattern instead?  

Today, you are checking tense. Tomorrow, you will need to check number agreement (singular-plural).  Also the mood (subjunctive, conditional, etc.) https://en.wikipedia.org/wiki/Grammatical_mood   and then aspect ... https://en.wikipedia.org/wiki/Grammatical_aspect ...

And then, some more things, if we ever get this to work for languages other than English -- e.g. declension https://en.wikipedia.org/wiki/Declension

The whole point of having a pattern matcher was to avoid having to hard-code rules in C++ code, yet here we have a rule that is hard-coded in C++.  Maybe its OK for now, but it sets a dangerous precedent.

linas:
Oh, and the reason I am looking at this: I would like support added for synonymous phrases.  For example,  the current chatbot self-awareness module can answer the question "where are you looking?"  with `(Predicate "look") (Concept "leftwards")`  which sureal converts to "I am looking leftwards"

The problem is that this is awkward -- a better reply would be "I am looking to the left" or even "I am looking to the side" or even "I am not looking straight ahead".   The question then becomes: where does phrasing and prosody get handled?  The choice of wording is a deeply personal indicator of personality and intelligence.  Choice of synonymous expressions is part of it, but so is the play of number, tense, aspect and mood.  Hard-coding such transformations in C++ code is maybe OK for now, but not quite right in the long-run.

PR #2072, opened 2016-03-09
Fix R2L gender rule

leungmanhin:
Need to pass the gender to the gender-rule

### DIFF
--- a/opencog/nlp/relex2logic/rules/gender.scm
+++ b/opencog/nlp/relex2logic/rules/gender.scm
@@ -5,14 +5,16 @@
 			(var-decl "$a-parse" "ParseNode")
 			(var-decl "$word" "WordInstanceNode")
 			(var-decl "$lemma" "WordNode")
+			(var-decl "$gtype" "DefinedLinguisticConceptNode")
 		)
 		(AndLink
 			(word-in-parse "$word" "$a-parse")
 			(word-lemma "$word" "$lemma")
 			(word-feat "$word" "person")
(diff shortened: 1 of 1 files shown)


opencog/nlp/relex2logic/rules/gender.scm
```diff
@@ -5,14 +5,16 @@
...
 			(word-lemma "$word" "$lemma")
 			(word-feat "$word" "person")
-			(ChoiceLink
-				(word-feat "$word" "masculine")
-				(word-feat "$word" "feminine")
+			(InheritanceLink (Variable "$word") (Variable "$gtype"))
+			(OrLink
+				(EqualLink (Variable "$gtype") (DefinedLinguisticConceptNode "masculine"))
+				(EqualLink (Variable "$gtype") (DefinedLinguisticConceptNode "feminine"))
 			)
```
linas: this change reverts a previous fix. Why?  It does the same thing as before, but is slower and more awkward.

leungmanhin: because the gender-rule (https://github.com/opencog/opencog/blob/master/opencog/nlp/relex2logic/rule-helpers.scm#L518) wants to know whether it is "masculine" or "feminine" so that it can generate the corresponding gender for a person, so it seems to me that variable $gtype needs to be grounded

linas: ah, I see. The $gtype variable had been clobbered!  OK.

PR #2078, opened 2016-03-14
Remove the SCM_PRELOAD from the config files as well.

linas:
This requires that pull request https://github.com/opencog/atomspace/pull/703 be merged, as a pre-req  (and that pull request is not mergable for unknown reason).


linas:
test this please

PR #2081, opened 2016-03-16
Remove several pre-rule helpers

linas:


linas:
Note: this pull request assumes that the previous pull requests get merged first, otherwise the code will get out of order.

linas:
argh. This is urgently required for the latest HR work, but is blocked by pull req opencog/atomspace#703 which is still reverberating down the line.   This is becoming hackier and hackier.

linas:
DUE TO THE FUCKED-UP-NESS OF JENKINS-CI THIS PULL REQUEST ALSO HAS BEEN MANGLED TO INCLUDE PULL REQUEST #2083 

IF WE DID NOT USE THIS FUCKED UP SYSTEM., THEN THE GT COMMITS WOULD NOT BE FUCKED UP.  DOESN'T ANYONE OUT THERE UNDERSTAND HOW FUCKED UP THIS SYSTEM IS?

WHAT IS WRONG WITH Y'ALL????????????????????????

PR #2082, opened 2016-03-17
minecraft updates

dalecassidy:
These are a couple of changes needed to get minecraft working due to python api changes


opencog/cython/opencog/spatial.pyx
```diff
@@ -105,10 +105,9 @@ cdef class EntityRecorder:
     def __cinit__(self):
         pass
 
-    def __init__(self, long addr, AtomSpace atomspace):
+    def __init__(self, long addr):
         self.c_entity_recorder = <cEntityRecorder*> PyLong_AsVoidPtr(addr)
-        self.atomspace = atomspace
-
+
```
linas: are you sure?  This now differs from how `__init__` works for OctomapOCTree up above. Also, it looks like now, self.atomspace is never initialized; yet it is used by more-or-less every method in this class ...

Atoms, once inserted into some atomspace, always know what atomspace they are in.  Only fresh atoms, that have never been inserted anywhere, don't know where they belong., so they're left floating about.   Maybe the atomspace argument isn't needed for most of these functions? Maybe passing an invalid atomspace is OK ???

linas: p.s. I'm not an expert at his stuff, someone  else should review and commit as needed. This just caught my eye.

dalecassidy: You may be right. I'll take a look at it again.

linas:
rejecting, this appears to be super-ceeded by pull req #2090

linas:
oh, wait, it looks like this branch was also a part of #2090 and got merged when that one got merged.

PR #2084, opened 2016-03-20
Fix broken link for MOSES example experiment in python

eddiemonroe:


linas:
you should always push branches to your own copy of the repo, instead of the main opencog repo.

eddiemonroe:
yeah, i had a feeling you were going to say that... i succumbed to temptation to do such a small change quick right on the github site, but yeah yeah, you're right.

PR #2087, opened 2016-03-22
unfuck the unit tests; fuck junkins instead.

linas:


AmeBel:
@linas protection of this repo has been disabled

AmeBel:
Some more clarification, The test will run but github's branch protection is disabled. So you can merge changes without waiting for the jenkins tests to complete.

linas:
Microplanning passes if one says "make install" first.  However, without the make install step,, Microplanning fails.

PR #2090, opened 2016-03-22
Changes for minecraft api

dalecassidy:
This code now correctly initializes the atomspace


linas:
looks OK, merging.

The other way to handle this would have been to push your commits to the earlier pull request #2082, instead of creating a new pull request.

Issue #2093, opened 2016-03-26
Problem configuring python bindings


Selameab:
I followed the steps in https://github.com/opencog/opencog/blob/master/opencog/python/README.md to configure the python bindings. But "loadpy" doesn't show up in the opencog shell.

The console output from starting the cogserver is attached below:

 ```
Using default config at lib/opencog.conf
Using default config at ../lib/opencog.conf
Using default config at /usr/local/etc/opencog.conf
Traceback (most recent call last):
  File "/opencog/opencog/python/preload_functions/test_functions.py", line 6, in <module>
    from test_helpers.bind_helpers import increment_green, increment_red
ImportError: No module named test_helpers.bind_helpers
Listening on port 17001
 ```

inflector:
It looks like some files were not copied as part of the split of the repositories last year.

linas:
The increment_green and increment_red is some kind of cruft from a demo to incrmeent green lights and read lights, it should not be in the cogserver at all.   I never heard of loadpy before ...

linas:
Hmm  greping through the code .. loadpy does not seem to make sense .. what's wrong with just saying "import blah" the way that modules are normally imported in python? Why would a special import function be needed?

linas:
After looking at `cogserver/modules/python/PythonModule.h` my knee-jerk response is that code should just be removed.  As far as I can tell, it does not offer the ability to do anything that cannot already be done in a simpler fashion, using "ordinary" python, from the usual python prompt.   The file `cython/README.md` also talks about loadpy, but does not explain why it is actually needed, or why its better than just saying `import mymodule`.   I'm thinking that this is all cruft left over from an earlier design philosophy, when the nature of the python bindings to opencog were .. less well understood.

linas:
The increment_green stuff is from `python/preload_functions/test_functions.py` which should not be loaded, except when running the unit tests.  I'll move that to the test directory now.

inflector:
We did rework all the Cython to make it possible to bypass the CogServer for most usages. I don't see any reason for a special import function. 

However, 'loadpy' is doing more than import it is looking for Python MindAgent and Request subclasses that can run in the CogServer. See: https://github.com/opencog/opencog/blob/master/opencog/cython/opencog/agent_finder.pyx#L63

inflector:
[...]

PR #2097, opened 2016-03-29
Use PYTHONPATH to find python code.

linas:
The standard way of finding python code is to use PYTHONPATH.  Just use that.

The needs pull req opencog/atomspace#744 as a pre-req


linas:
This addresses the issues seen in #2093

PR #2103, opened 2016-03-31
Simple verbal robot training

eddiemonroe:
Initial pull of functionality to execute behavior rules based on AIML-style string matching and simple training of stimulus and response behavior through verbal interaction with the bot.


linas:
Thanks Eddie.  I can't review the code right now, I'll try to look at it shortly.  Some brief remarks:

-- I plan to change the AIML converter to use all lower-case, not all upper-case, for word-nodes. 
-- I plan to make minor changes to pattern matching format, so that it fits better into the grand scheme of things.  There may be some turbulence in the coming days/weeks.  Do report bugs if you see them.
-- cog-reduce will go away. replaced by the standard code-execute!
-- I plan to ditch the code in `aiml2oc/aiml2oc_guile/code/` but I don't believe you are using anything important in there. 

My very next commit moves `aiml2oc/aiml2oc_guile/code/` to `aiml/import/scm-variant`

What you are probably using in `aiml2oc/aiml2oc_guile/code/` is just the sentence tokenizer and other scaffolding.  You can continue to use it in the short-term, but the correct long-term solution is to get sentences from relex.  See `aiml/running/run.scm`  but be aware that the name of the file and the contents will change sharply in the coming weeks.

PR #2123, opened 2016-04-14
Change in openpsi rule representation 

AmeBel:


AmeBel:
This moves openpsi-rule from being ure-rules to ImplicationLinks structured as
```scheme
    (ImplicationLink
        (AndLink
            (context)
            (action))
        (demand-goal))
```
Documentation and example are not udated yet.

rTreutlein:
Just curious what is the reasoning behind this?

AmeBel:
So as to make the openpsi-rules content for PLN rules.

ngeiswei:
BindLink and ImplicationLink are semantically different. Could they be unified? Maybe, but for now they aren't, and the TV on the implication allows us to reason and build action plans dealing with uncertainty, using PLN rules (which are represented as BindLinks).

ngeiswei:
So ImplicationLinks are more like facts, and BindLinks are more like rules.

linas:
> This moves openpsi-rule from being ure-rules to ImplicationLinks
> structured as
>
>     (ImplicationLink
>         (AndLink
>             (context)
>             (action))
>         (demand-goal))
>
>
I like this new format.  It aligns (very) well with my most recent thinking
about things.

linas:
> BindLink and ImplicationLink are semantically different. Could they be
> unified? Maybe, but for now they aren't, and the TV on the implication
> allows us to reason and build action plans dealing with uncertainty, using
> PLN rules (which are represented as BindLinks).
>

Well, its more than just this. There is a serious design difficulty that
comes from using BindLinks, that ImplicationLinks avoid.   Let me give an
example: To use a BindLink to process some input text, I have to write it
like so: (BindLink (AndLink (Var $sent is current sentence) (Var $parse is
parse of $sent) (Var $word in $parse) ($word is a verb)....etc))) which
rapidly becomes large and painful to maintain and closely tied to the
current way of representing sentences.

By using ImplicationLinks, some of these details can be fudged and
omitted.  That's because ImplicationLinks don't have to actually be
executable: they can be fragments.  An engine that picks out and executes
the implicationLinks can supply the missing details, like "apply this
processing only to the current sentence".

My vague current daydream is to have some system that  searches and finds
the best ImplicationLinks that apply to the current context+demand-goal,
weeds them down to some manageable sets, creates the needed BindLink on the
fly, runs the BindLink, and then deletes it.

Roughly speaking: BindLinks fail unless they exactly match something in the
atomspace.  This makes it hard/impossible to create "fragments of a
BindLink", or to do fuzzy searches for such fragments.  The ImplicationLink
seems to be a very good way for dealing with fragments.

I plan to move away from the current behavior-tree design for the robot
behaviors, and ditto the current BindLink-based chatbot design, to an
ImplicationLink design.  I don't yet see how this overlaps with OpenPsi,
but it certainly moves closer in that direction.

AmeBel:
In this pr, first the context of the ImplicationLink is wrapped in a SatisfactionLink and evaluated, if `(stv 1 1)` then the requirements for executing the action are assumed to be satisfied and the action is executed and the goal (aka demand-goal) is evaluated. 

One can supply, a DefinedSchemaNode(aka action-selector)  which filters out the `ImplicationLinks` that should be choosen and for their actions to be exectued and their goals evaluated. I will add a simple action-selector that accounts for the stv of the ImplicationLink next.

linas:
I need to spend a week catching up with old email, and a giant list of non-coding stuff I've ignored. After that, I can talk about porting ros-behavior to openpsi.  I think that would be a good trial run. I just wrote an email about this.

PR #2127, opened 2016-04-18
Update psi-step

leungmanhin:
This aims to updates all demands associate with an action when it is executed


leungmanhin:
@AmeBel please review

PR #2142, opened 2016-05-09
Amusing friend fc fixes

ngeiswei:
- Fix support amusing friend demo FC
- Now both the MOSES PLN synergy and the amusing friend demo can co-exist
- More PLN rule tuning mostly due to the lack of distributional TV


ngeiswei:
PLNRulesUTest fails on Jenkins but passes on my system after deleting the cache from guile

```sh
rm -r .cache/guile/ccache/
```

PR #2155, opened 2016-05-27
Workaround scheme load path mystery by using PROJECT_SOURCE_DIR

ngeiswei:


ngeiswei:
I've got that feeling that unit tests were failing less often when we had branch protection...

PR #2157, opened 2016-05-27
Fix more scheme path mystery

ngeiswei:


tests/pln/MOSESPLNSynergyUTest.cxxtest
```diff
@@ -78,7 +78,14 @@ void MOSESPLNSynergyUTest::tearDown()
 
 void MOSESPLNSynergyUTest::setUp()
 {
-	eval.eval("(add-to-load-path \"/usr/local/share/opencog/scm\")");
+	string cur_pp_dir = string(PROJECT_SOURCE_DIR),
+		cur_dir = cur_pp_dir + "/tests/pln";
```
linas: this code is more or less guaranteed to break!  It might work on your machine, this week, but anyone who is running outside of the project source dir will see failures. In particular, any of the ros and robotics setups are gauaranteed to fail.

The work-around for the general pain of the path that I've been using is here:
https://github.com/opencog/atomspace/blob/master/tests/query/CMakeLists.txt#L92
notice that the source files are installed into the build directory.

PR #2158, opened 2016-05-29
Assorted OpenPsi and AIML fixes

linas:


opencog/openpsi/README.md
```diff
@@ -21,7 +21,7 @@ is similar to `Belief + Intention ==> Desire`.
   * A rule is an ImplicationLink structured as
     ```scheme
             (ImplicationLink
-                (AndLink
+                (SequentialAndLink
```
ngeiswei: Hmm, why do you use a SequentialAnd? This is true if action is true after context is true. I would think that you want instead to represent that you should take this action while context is true,

The right way t seems to me is to use a not-yet-implemented-though-defined PredictiveImplicationLink, as follows

```
PredictiveImplication
   And
      Context
      Action
   Goal
```

ngeiswei: See http://wiki.opencog.org/wikihome/index.php/PredictiveImplicationLink

PR #2159, opened 2016-05-30
Return the ImplicationLink in (psi-rule)

leungmanhin:


opencog/openpsi/main.scm
```diff
@@ -17,6 +17,8 @@
 "
   psi-rule-nocheck -- same as psi-rule, but no checking
 "
+    (define implication (Implication a-stv (SequentialAnd context action) goal))
```
ngeiswei: Again I think the `SequentialAnd` is ill-used, better use a `PredictiveImplication`

See http://wiki.opencog.org/wikihome/index.php/PredictiveImplicationLink

Issue #2146, opened 2016-05-17
Minor typo in HebbianUpdatingAgent.cc


moikle:
On line 119, getLTI() should be getSTI()

svadali2:
i made the change and have given a pull request accordingly.

linas:
I merged pull req #2182  please close this bug report if thats OK

PR #2182, opened 2016-06-08
changed typo from LTI to STI, line 119

svadali2:


linas:
Per email [opencog/opencog] Minor typo in HebbianUpdatingAgent.cc (#2146)

fixes bug #2146

Issue #2193, opened 2016-06-14
Lack of homogeneity in R2L outputs


ngeiswei:
See for instance

```
;; r2l output corresponding to (chat "Bob lives in China")

   (SetLink
      (InheritanceLink
         (ConceptNode "in@b8dbff0f-ffe4-4775-9905-f9c87f29f63d")
         (ConceptNode "in" (stv 0.032258064 0.0012484394))
      )
      (ImplicationLink
         (PredicateNode "bob@04c3177d-bb4d-4a44-ae30-f4acbf1ea59f")
         (PredicateNode "bob" (stv 0.16666667 0.0012484394))
      )
      (InheritanceLink
         (SatisfyingSetLink
            (PredicateNode "bob@04c3177d-bb4d-4a44-ae30-f4acbf1ea59f")
         )
         (ConceptNode "in@b8dbff0f-ffe4-4775-9905-f9c87f29f63d")
      )
      (InheritanceLink
         (ConceptNode "China@b5d7ee9c-4cb1-4a70-b7ea-4dec2db35558")
         (ConceptNode "China" (stv 0.032258064 0.0012484394))
      )
      (ImplicationLink
         (PredicateNode "in@b8dbff0f-ffe4-4775-9905-f9c87f29f63d")
         (PredicateNode "in" (stv 0.16666667 0.0012484394))
      )
      (EvaluationLink
         (PredicateNode "in@b8dbff0f-ffe4-4775-9905-f9c87f29f63d")
         (ListLink
            (PredicateNode "bob@04c3177d-bb4d-4a44-ae30-f4acbf1ea59f")
            (ConceptNode "China@b5d7ee9c-4cb1-4a70-b7ea-4dec2db35558")
         )
      )
      (EvaluationLink
         (DefinedLinguisticPredicateNode "definite")
         (ListLink
            (ConceptNode "China@b5d7ee9c-4cb1-4a70-b7ea-4dec2db35558")
         )
      )
      (EvaluationLink
         (PredicateNode "in@b8dbff0f-ffe4-4775-9905-f9c87f29f63d")
         (ListLink
            (ConceptNode "China@b5d7ee9c-4cb1-4a70-b7ea-4dec2db35558")
         )
      )
      (InheritanceLink
         (InterpretationNode "sentence@0c2ab58f-7150-4259-8cd2-52ddf260d7f1_parse_0_interpretation_$X")
         (DefinedLinguisticConceptNode "ImperativeSpeechAct")
      )
[...]

PR #2194, opened 2016-06-14
R2l consistence

ngeiswei:
Here's a (likely very partial) fix for #2193. Look in particular at commit fa14370. I've got say I'm completely stabbing in the dark. I plan on carefully reading http://wiki.opencog.org/wikihome/index.php/Relex2logic in order to make sense of that.

### DIFF
--- a/opencog/nlp/relex2logic/rule-helpers.scm
+++ b/opencog/nlp/relex2logic/rule-helpers.scm
@@ -466,26 +466,26 @@
 	(define adv (cog-name adv-node))
 	(define adv_instance (cog-name adv-instance-node))
 	(ListLink
-		(InheritanceLink  (ConceptNode adv_instance) (ConceptNode adv))
-			(ImplicationLink  (PredicateNode instance) (PredicateNode verb))
-			(r2l-wordinst-Predicate instance)
-			(r2l-wordinst-concept adv_instance)
-			(InheritanceLink  (SatisfyingSetLink (PredicateNode instance)) (ConceptNode adv_instance))
+		(ImplicationLink  (PredicateNode adv_instance) (PredicateNode adv))
+		(InheritanceLink  (ConceptNode instance) (ConceptNode verb))
+		(r2l-wordinst-concept instance)
+		(r2l-wordinst-Predicate adv_instance)
+		(InheritanceLink  (ConceptNode instance) (SatisfyingSet (PredicateNode adv_instance)))
 	)
 )
 
 (define-public (adverbialpp-rule verb instance prep  prep_instance noun noun_instance)
     (ListLink
     (InheritanceLink  (ConceptNode noun_instance) (ConceptNode noun))
-    (ImplicationLink  (PredicateNode instance) (PredicateNode verb))
+    (InheritanceLink  (ConceptNode instance) (ConceptNode verb))
     (ImplicationLink  (PredicateNode prep_instance) (PredicateNode prep))
-    (r2l-wordinst-Predicate instance)
+    (r2l-wordinst-concept instance)
     (r2l-wordinst-concept noun_instance)
     (r2l-wordinst-Predicate prep_instance)
     (EvaluationLink
         (PredicateNode prep_instance)
         (ListLink
-            (PredicateNode instance)
+            (ConceptNode instance)
             (ConceptNode noun_instance))
 	))
 )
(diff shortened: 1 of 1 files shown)


opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -466,26 +466,26 @@
 	(define adv (cog-name adv-node))
 	(define adv_instance (cog-name adv-instance-node))
 	(ListLink
-		(InheritanceLink  (ConceptNode adv_instance) (ConceptNode adv))
-			(ImplicationLink  (PredicateNode instance) (PredicateNode verb))
-			(r2l-wordinst-Predicate instance)
-			(r2l-wordinst-concept adv_instance)
-			(InheritanceLink  (SatisfyingSetLink (PredicateNode instance)) (ConceptNode adv_instance))
+		(ImplicationLink  (PredicateNode adv_instance) (PredicateNode adv))
+		(InheritanceLink  (ConceptNode instance) (ConceptNode verb))
```
linas: Since when are verbs "Concepts"?  I thought only nouns are "concepts", and verbs are "predicates", right? 

Or maybe the R2L subsystem doesn't actually make this kind of distinction?  Or maybe this distinction doesn't matter?

linas: After glancing at the wiki, there does not appear to be any clear, crisp statement as to when some linguistic structure is mapped to "predicate" and when its mapped to "concept".  The wiki should try to explain this.

More generally, I am made somewhat nervous, at the gut-intuition level, that such a distinction could ever be correct.  Lignustics is complicated, for a reason; that complexity is ignored at one's peril.

bgoertzel: Verbs should be predicates, nouns should be concepts...

On Wed, Jun 15, 2016 at 6:23 PM, Linas Vepštas <notifications@github.com>
wrote:

> In opencog/nlp/relex2logic/rule-helpers.scm
> <https://github.com/opencog/opencog/pull/2194#discussion_r67136700>:
>
> > @@ -466,26 +466,26 @@
> >  	(define adv (cog-name adv-node))
> >  	(define adv_instance (cog-name adv-instance-node))
> >  	(ListLink
> > -		(InheritanceLink  (ConceptNode adv_instance) (ConceptNode adv))
> > -			(ImplicationLink  (PredicateNode instance) (PredicateNode verb))
> > -			(r2l-wordinst-Predicate instance)
> > -			(r2l-wordinst-concept adv_instance)
> > -			(InheritanceLink  (SatisfyingSetLink (PredicateNode instance)) (ConceptNode adv_instance))
> > +		(ImplicationLink  (PredicateNode adv_instance) (PredicateNode adv))
> > +		(InheritanceLink  (ConceptNode instance) (ConceptNode verb))
>
> Since when are verbs "Concepts"? I thought only nouns are "concepts", and
> verbs are "predicates", right?
>
> Or maybe the R2L subsystem doesn't actually make this kind of distinction?
> Or maybe this distinction doesn't matter?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/opencog/opencog/pull/2194/files/c273c1c62c0728ab6d575ff7538a130e543ebef8#r67136700>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AFolXNGc2MFihCBssjYmATboet5VEp2Mks5qL9KYgaJpZM4I1WZs>
> .
>



-- 
Ben Goertzel, PhD
http://goertzel.org

Super-benevolent super-intelligence is the thought the Global Brain is
currently struggling to form...

ngeiswei: From a PLN perspective, it doesn't matter whether verbs are predicates or concepts, right? I suppose it does make some sense from an NLP viewpoint (I'm really unfamiliar with NLP in general).

So the problem is that rules like `advmod-rule` should not be used with nouns, I guess something is wrong in either the R2L rules or the LG. I'll look into that, obviously if you know how to fix it, let me know.

ngeiswei:
Oh, it looks like this has been in fixed in LG https://github.com/opencog/opencog/issues/2193#issuecomment-225932347 !

opencog/nlp/relex2logic/rule-helpers.scm
linas: Nil, when you say: "(I'm really unfamiliar with NLP in general)" -- the interface between PLN & NLP is exactly where the hard and interesting problems all are.  That's where the future work lies.

PR #2196, opened 2016-06-15
time-octomap, point memory

yantrabuddhi:
this contains time-octomap library and point-memory: a scheme wrapper module to time-octomap


opencog/timeoctomap/pointmemory/PointMemorySCM.cc
```diff
@@ -0,0 +1,360 @@
...
+    pl.pop_front();
+  }//while
+  
+  return Handle(createLink(LIST_LINK,loc_links));
+}
+
+Handle PointMemorySCM::get_past_locs_ato(const string& map_name,Handle ato,int elapse)
+{
+  time_pt tpt; duration_c dr;
+  tsa[map_name]->get_current_time_range(tpt,dr);
```
ngeiswei: @yantrabuddhi Wait, don't tell you indent yourself! Again, curious what editor you use.

opencog/timeoctomap/TimeOctomap.cc
```diff
@@ -29,67 +29,71 @@
...
 //#include <assert.h>
 #include "opencog/util/oc_assert.h"
-TimeSpaceAtom::TimeSpaceAtom(unsigned int num_time_units,
-                             vector<double>map_res_meters):
-    time_circle(num_time_units), created_once(false)
+
+TimeOctomap::TimeOctomap(unsigned int num_time_units,
+                         double map_res_meters,
+                         duration_c time_resolution):
+    time_circle(num_time_units),map_res(map_res_meters),time_res(time_resolution),created_once(false)
```
ngeiswei: To avoid some gdb warning, time_circle should be initialized after map_res and time_res.

yantrabuddhi: Hi Nil,
 I am confused. I use vim, however some places are supposed to be tabs and
some spaces. I am a very basic user of vim, i don't know advanced stuff. If
it were possible to automatically indent the right way, then can you let me
know?
Regards,
Mandeep

ngeiswei:
Could you fix as well the indentations here and there (2 to 4 whitespaces)?

opencog/timeoctomap/TimeOctomap.cc
ngeiswei: I'm an emacs. I looked into tab vs space autodetection a few months ago for emacs but it didn't work well (though I'm sure it wouldn't be hard to fix). Don't know if there is something like that for vim.

PR #2206, opened 2016-06-20
Initial version of the which-product demo

ngeiswei:


ngeiswei:
Still reviewing once more before merging.

Issue #2192, opened 2016-06-14
Error occurs when nlp-parsing 'Let us do something' kind of sentences


leungmanhin:
For example:
```
(nlp-parse "Let us change the subject.")
```
and I got:
```
ERROR: In procedure opencog-extension:
[...]

PR #2212, opened 2016-06-22
Assorted updates to the chatbot

leungmanhin:
Add a check to the parse status, to be used by new psi-rules, also to be consistent, pass the input/output in the form of a list of WordNodes


opencog/nlp/chatbot-psi/chatbot.scm
```diff
@@ -18,51 +18,62 @@
...
-    (let* ((sent-node (car (nlp-parse utterance)))
-           (list-of-words (get-word-list sent-node)))
-
-        (State input-utterance
-            (Reference
-                list-of-words
-                sent-node
-                (Node utterance)
-            )
+    (catch #t
```
linas: what error is being thrown here? Can you add a few sentences to the code, explaining why this is needed?

leungmanhin: ok I'll add comments to the code, the reason is because nlp-parse might fail to parse a sentence, this rarely happens (and should be fixed instead) but when it happens it would be nice to catch the error so that the robot could react accordingly, e.g. look confused or say something like "I don't understand" etc, soon this should be expanded to also detect inputs that are not really grammatical

linas: well, I understand that, but where is nlp-parse throwing an exception if it fails to parse? Most exceptions are thrown for programmer errors, and indicate a bug in the code, rather than invalid input data.

leungmanhin: Mostly the errors come from R2L... right maybe I should remove this catch, at least in the medium/long run we won't need this, having it would hide any real bugs in the code and may also give a wrong impression that the user has entered something wrong

linas: what exceptions are being thrown? Normally, the file name and the line number are printed as a part of the exception message. do you have some examples?

leungmanhin: Here is one: https://github.com/opencog/opencog/issues/2192

linas: I fixed #2192 just now, but I suppose that there are probably other exceptions too. They should be logged, not hidden ...

leungmanhin: Yes will do, thanks for the fix

PR #2213, opened 2016-06-23
Cache between SuReal and PatternMatcher

andre-senna:
- Implemented a new entry-point in SuReal to be used in the Microplanner. This new function uses a cache to speed up PatternMatcher execution.


opencog/nlp/microplanning/main.scm
```diff
@@ -92,6 +92,13 @@
 	; Initialize the sentence forms as needed
 	(microplanning-init)
 
+    ; Reset (clear) SuReal cache to assure it is empty before start calling sureal.
+    ; Remark that SuReal Cache is not thread-safe. So it is not supposed to be used
```
linas: why is it not thread safe?

opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -66,29 +68,50 @@ SuRealPMCB::~SuRealPMCB()
...
+    if (hPat->getType() != hSoln->getType()) {
+        answer = false;
+    } else {
+        // VariableNode can be matched to any VariableNode, similarly
+        // for InterpretationNode
+        if (hPat->getType() == VARIABLE_NODE or
+            hPat->getType() == INTERPRETATION_NODE) {
+            answer = true;
+        } else {
+            std::string sSoln = NodeCast(hSoln)->getName();
```
linas: You don't need to do NodeCasts any more, and its faster if you don't.

opencog/nlp/microplanning/main.scm
andre-senna: Because in this version there is only one instance of the cache (reached via a singleton wrapper). So if two threads with two different Microplanner queries add stuff to the cache, one may lead to false hits in the other. In addition to this, the cache is reset (clear) just before the Microplanner query starts so if the second query reach the reset point before the first query ended, the cache will be reset during the lifetime of the first query, which may lead to false misses.

If this cache approach proves to be interesting (valuable speedup), I will implement a second version where each Microplanner query have its own cache instance.

linas:
Is there a description anywhere of how sureal is using the pattern matcher? i.e. what patterns it is trying to match?  It's not entirely obvious to me that the pattern matcher is the most effiicient way of doing what needs to be done here, but I'm also unclear about what it is trying to do.

andre-senna:
There is a description of the algorithm in the README.md of SuReal folder.

linas:
That is a general description of the algo. It does not describe how pattern matcher is being used, and why custom callbacks for the pattern matcher are needed, and what they do.

I've seen the pattern matcher abused in the past: in at least one case, just doing the thing directly improved performance by a factor of 50x and cut the total amount of code by 2/3rds. So -- are you sure the pattern matcher is not being abused or not being used in an inefficient manner? how can you tell?

andre-senna:
No, I can't assure it is not being abused. Actually, given that I was able to write a very simple cache to speed up SuReal (when being used by Microplanner), I'd say that you are probably correct that the PatternMatcher should not be used in this case. But I didn't write SuReal and to be honest I didn't fully understand how it works (I don't understand it enough to re-write it from scratch).

linas:
Hmm. well, I know that a few weeks ago, it was not performing well, and was a bottleneck, and was being by-passed.  I'm trying to figure out what the most important things are, that I need to work on, and perfaphs taking a deep dive into sureal might be one of them.   Not sure what my priority list is.

bgoertzel:
Linas -- I believe that the microplanner, not SuReal, was being bypassed
(and still is).  that is why I asked Senna to optimize the microplanner...

-- ben

On Fri, Jun 24, 2016 at 3:29 AM, Linas Vepštas <notifications@github.com>
wrote:

> Hmm. well, I know that a few weeks ago, it was not performing well, and
> was a bottleneck, and was being by-passed. I'm trying to figure out what
> the most important things are, that I need to work on, and perfaphs taking
> a deep dive into sureal might be one of them. Not sure what my priority
> list is.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/opencog/opencog/pull/2213#issuecomment-228158582>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AFolXAta6kYH4joFaqErFlfFsz1-3cxDks5qOt63gaJpZM4I8YOX>
> .
>



-- 
Ben Goertzel, PhD
http://goertzel.org

Super-benevolent super-intelligence is the thought the Global Brain is
currently struggling to form...

PR #2224, opened 2016-06-28
Report whether the pattern miner is compiled

ngeiswei:


CMakeLists.txt
```diff
@@ -694,4 +694,5 @@ SUMMARY_ADD("Server" "The CogServer" HAVE_SERVER)
 SUMMARY_ADD("Spatial" "Vizualisation of the 3D spacemap" HAVE_SPATIAL_TOOLS AND HAVE_OCTOMAP)
 SUMMARY_ADD("Unit tests" "Unit tests" CXXTEST_FOUND)
 SUMMARY_ADD("NLP" "Natural Language Processing" HAVE_NLP)
+SUMMARY_ADD("PatternMiner" "Pattern Miner" HAVE_cpprest)
```
linas: why in the world would the pattern miner depend on the REST interfaces??

linas: Oh, never mind, I see it has a client-server mode.

ngeiswei: Perhaps what I should have done is define a cmake constant HAVE_PATTERN_MINER and use that instead, but I figured we can probably do that later on if the pattern miner requirements evolves.

linas: See #2225 its low priority, though

PR #2245, opened 2016-07-06
Remove '\' from the chatbot output, if any

leungmanhin:


opencog/nlp/chatbot-psi/actions.scm
```diff
@@ -72,8 +72,9 @@
 (define (say . words)
     (define utterance (string-join (map cog-name words)))
 
-    ; Remove those '[' and ']' that may exist in the output
-    (set! utterance (string-filter (lambda (c) (not (or (char=? #\[ c) (char=? #\] c)))) utterance))
+    ; Remove those '[', ']' and '\' that may exist in the output
```
linas: any clue where these are coming from?

leungmanhin: I hit one reply that has a WordNode `(WordNode "\"hello\".")`

I guess '\' may be used to escape quotes exist in some word in some aiml rule, and after converting to opencog psi rules they get the chance to be part of a word

linas: ooofdah. quotes are hard. Can you add a short comment that says `; The backslash is sometimes generated by AIML rules. The square brackets are sometimes generated by ...`  ?  In the short run, stripping stuff out is OK but In the long run, there should not be garbage in the WordNode string ...

leungmanhin: sure, I'll do that

Issue #2246, opened 2016-07-07
Bug in sureal


ngeiswei:
I can't generate speech if the model contains on the same words of the speech to generate.

Steps to reproduce:

1. `cd opencog/opencog/nlp/chatbot-psi`
2. `guile -l chatbot.scm`
3. `(chat "small cats are happy")`
4. `(Word "people") (Word "crazy")`
5. 
```
(sureal
   (SetLink
      (Evaluation
         (Predicate "happy")
         (List
            (Concept "people")))
      (Inheritance
         (Concept "people")
         (Concept "crazy"))))
```

returns `()`.

If you replace `happy` by `cute` in the exemplar sentence it works!

ngeiswei:
So far my findings leads to `SuRealPMCB`, basically `slp->satisfy(pmcb)` doesn't seem to find the exemplar sentence, I don't know why.

BTW, if I enter both exemplar sentences, it doesn't work either (which obviously is the most annoying).

ngeiswei:
The happy exemplar sentence is actually matched by `SuRealPMCB` but for some reason `SuRealPMCB::grounding` discards it. Still hunting...

ngeiswei:
With the happy exemplar sentence the word `happy` has a lemma link, and apparently that is what SuReal doesn't like, I don't why yet. @AmeBel @leungmanhin Any idea? I'll keep looking...

ngeiswei:
I think I found the problem. The when r2l turns `small cats are happy` it forgets to specify the tense on `Predicate "happy"` (which, I recall, actually stands for "be_happy"). I see 2 ways to fix that

1. Fix the r2l rule so it specifies the tense of that weirdo
2. Relax `SuRealPMCB::grounding` so it assumes a present tense by default.

maybe there is a better fix, I'll keep looking...

linas:
Hmm. Well, "happy" is not a verb, so it would not have a tense. "to be" is the verb, "are" is present tense. 

Techically, "are happy" makes "happy" a predicative adjective; Aaron had some plans for this, but note: in general, "to be" is used in zillions of different ways, its hard to cpature all the variations correctly.

linas:
Oh, I see, yes, if PredicateNode "happy" really means "to be happy", then yes, it should have gotten a tense marker on it.

I'm reopening the bug, since I figure the fix to r2l is better, right? cause you want other tenses to work, e.g. "cats were happy before there were dogs"  (?)

PR #2249, opened 2016-07-07
Do not check tense consistence if undefined in the exemplar sentence

ngeiswei:
Fix for issue #2246.

@AmeBel @leungmanhin please check it carefully. It might break stuff elsewhere. I leave it on you to merge it. Thanks.


ngeiswei:
Actually I think there is a better fix.

It's better to check if the logical structure given to SuReal contains tense, if it doesn't contain tense then no tense check needs to be performed.

It it slightly less relaxed and account for the case where a forthcoming exemplar happens to have a tense equal to the one of the logical structure, which would therefore be a better match, whereas this patch would return the first match without tense, regardless of the logical structure tense.

Unless you are sure that this fix it just fine, wait a bit, I'll likely send a new better one soon.

ngeiswei:
Please consider merging #2250 instead.

leungmanhin:
Merged #2250 instead, thanks

PR #2250, opened 2016-07-07
Fix SuReal tense comparison

ngeiswei:
I believe this is a better fix for #2246 than #2249.

It basically ignore tense comparison if the logical structure doesn't have a tense.

@AmeBel @leungmanhin, I believe it is correct but I'll let merge it. Thanks.

### DIFF
--- a/opencog/nlp/sureal/SuRealPMCB.cc
+++ b/opencog/nlp/sureal/SuRealPMCB.cc
@@ -445,21 +445,21 @@ bool SuRealPMCB::grounding(const std::map<Handle, Handle> &var_soln, const std::
                     // then get the tense of the one in this LemmaLink and see if they match
                     Handle hPatPredNode = m_as->get_handle(PREDICATE_NODE, sName);
                     IncomingSet qPatIS = hPatPredNode->getIncomingSetByType(INHERITANCE_LINK);
-                    bool tense = false;
+                    bool has_tense = false;
+                    bool eq_tense = false;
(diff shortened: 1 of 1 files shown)


opencog/nlp/sureal/SuRealPMCB.cc
```diff
@@ -445,21 +445,21 @@ bool SuRealPMCB::grounding(const std::map<Handle, Handle> &var_soln, const std::
...
                         HandleSeq qInhOS = lpInhLk->getOutgoingSet();
                         if (qInhOS[0] == hPatPredNode and
-                                qInhOS[1]->getType() == DEFINED_LINGUISTIC_CONCEPT_NODE and
-                                    sTense.compare(NodeCast(qInhOS[1])->getName()) == 0)
-                        {
-                            tense = true;
-                            break;
+                            qInhOS[1]->getType() == DEFINED_LINGUISTIC_CONCEPT_NODE) {
+                            has_tense = true;
+                            eq_tense = sTense == NodeCast(qInhOS[1])->getName();
```
linas: NodeCast is not needed, in general

leungmanhin:
looks all fine to me, and it works no problem, thanks for the fix!

opencog/nlp/sureal/SuRealPMCB.cc
leungmanhin: sureal is still using the old NodeCast / LinkCast in many different places, maybe I'll do a cleanup later

PR #2254, opened 2016-07-08
Integrate crazy happy demo into chatbot psi

ngeiswei:
@AmeBel @yantrabuddhi @leungmanhin you perhaps need to review that a bit, I don't mean the content but just the way it interact with the other rules. For now the weight 0.95 just so it's obviously easier to test, but given that the context is probably too relaxed (will be true if the question has words involved in inferred implications) it might be better to be set to 0.9 like the others.

I let it on you to merge. Thanks.

### DIFF
--- a/examples/pln/crazy-happy/crazy-happy.scm
+++ b/examples/pln/crazy-happy/crazy-happy.scm
@@ -141,8 +141,41 @@
 ;;    (ListLink
 ;;       (ConceptNode "Ben")))
 ;;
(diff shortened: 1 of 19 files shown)


ngeiswei:
Damn, I accidentally merged it while merging the master into it... Anyway, I think it's OK, but just in case it breaks your use cases you can merely disable loading `pln-reasoner.scm`, and if that's not enough `pln-psi-rules.scm`.

PR #2259, opened 2016-07-09
bug fixes for octomap

yantrabuddhi:

PR #2262, opened 2016-07-10
bug fix and formatting for octomap

yantrabuddhi:


linas:
This pull request appears to be almost identical to #2259 except that you closed that one.  Does that mean that you plan to close this one also?  ???

If you have changes that you want to incorporate into #2259, then please do so there, instead of making another request that clobbers the earlier one!

PR #2265, opened 2016-07-11
Nlp sentiment

ngeiswei:
PR of Ruiting's work on tagging sentences with sentiment.

1. You need to have `PYTHONPATH` properly configured as explained here http://wiki.opencog.org/wikihome/index.php/Python#Setting_up
2. Config files are installed under `/usr/local/etc/opencog`, this is a hardcoded path, would be better if it were configurable. It should be tolerable for the time being though.


leungmanhin:
Hmm.... I just tried to do `nlp-parse` but I got:
```
scheme@(guile-user)> (nlp-parse "hi")
ERROR: In procedure opencog-extension:
ERROR: Throw to key `C++-EXCEPTION' with args `("cog-evaluate!" "Python error in call_sentiment_parse: global name 'basic_sentiment_analysis' is not defined. (/atomspace/opencog/cython/PythonEval.cc:802)")'.
```

I'm running things using docker and already have python configured... Am I missing something?

ruiting:
me too ;(

ngeiswei:
It doesn't crash for me, however I'm not getting the sentiment anymore. Did you load the sentiment word list, as described in the readme2sentiment?

```
(load "../chatbot/loadSentiList.scm") ; from the chatbot-psi location
```
?

Of course this should be automatically done/

ruiting:
The sentiment parse code in the nlp-parse doesn't use "loadSentiList.scm".
They are for different purposes

Ruiting Lian

ngeiswei:
Oh, good to know, I should fix the readme2sentiment then.

Anyway, I'm not having this error, even or not if It load the sentiment word list. What is your `PYTHONPATH`?

ngeiswei:
And what do have under `/usr/local/etc/opencog/`?

ngeiswei:
And what do you have under `/usr/local/share/opencog/python/`?

leungmanhin:
My `PYTHONPATH` is
`/usr/local/share/opencog/python:/opencog/opencog/python/:/opencog/build/opencog/cython:/opencog/opencog/nlp/anaphora
`

I have `dicts` under `/usr/local/etc/opencog`
I have `opencog` and `basic_sentiment_analysis.py` under `/usr/local/share/opencog/python/`

linas:
We have a different repo for large datasets, we should be using that repo, instead of this one, for work such as this.  Please keep in mind that git is a block-chain, so that every time you do this, you bloat git and make it go slower.

Also -- instead of checking the code into the nlp/chatbot directory, it would be better if it was placed into its own directory, e.g. nlp/sentiment, or something like that.   By placing it into the chatbot dir, you end up breaking the chatbot module.

Also, -- mixing both scheme and python in one file is kind of ... interesting to look at, but questionable programming practice.

ngeiswei:
@linas sensible comments, I tried to find this repo for large dataset that you mentioned but I cannot find anything that matches in here https://github.com/opencog.

ngeiswei:
@leungmanhin strange, I'll look into it, but frankly I'm clueless.

ngeiswei:
@leungmanhin I presume you've installed nltk and its data, right?

ngeiswei:
Actually the sentiment is there too on my system.

I don't know what is the problem on your systems guys. What version of Python do you use?

linas:
Datasets get checked into https://github.com/opencog/test-datasets

leungmanhin:
@ngeiswei Yes I have installed nltk and its data.. python version is 2.7.6

leungmanhin:
ok I found the problem, we need to also install pyyaml in order to get `basic_sentiment_analysis` to work
```
sudo pip install pyyaml
```
I'll update the readme

ngeiswei:
Wow, @leungmanhin I'm curious to know how you manage to you find it, well done!

leungmanhin:
@ngeiswei I was suspecting something was wrong with `basic_sentiment_analysis` as the paths and files locations are all correct... so I opened a python interpreter and tried to do `import basic_sentiment_analysis`, and missing ymal was the error it then received

PR #2267, opened 2016-07-11
Helper classes to deal with large .scm files

andre-senna:
(1) SCMLoader 

A helper class that implements a simple strategy to load huge .scm files by parsing
each toplevel atom separately and feeding it (with all the inner atoms, in the case 
of links) in an AtomSpace.

(2) PGStorageDelegate

A helper class that uses SCMLoader to load a .scm file into an AtomSpace and store
this AtomSpace in a Postgress DB.

(3) Helper executables

There are also two executable programs.

    scm2pgsql - Expects a .scm file name and the DB credentials and properly read the 
                file and store the resulting AtomSpace into the DB. This program is
                installed by default by "make install". Call it with no args to see 
                a usage message.
    loadscm   - Just creates an empty AtomSpace and populate it with the contents of
                the passed file before dying. This is useful to test with new .scm files
                and probe the amount of time/RAM required to load them. It is not 
                installed by default. It will be in build/opencog/nlp/scm2pgsql/loadscm


linas:
Since this has nothing at all to do with NLP,  it surely does not belong in the NLP directory.   

I assume that you wrote the scm-loader because calling load took too long.  For large files, the load will try to compile the file; if you turn off compilation, then the loading should be much faster.  To load a file without compiling it, try  `(primitive-load "/tmp/bigfile.scm")`  -- is that still too slow? Did that not work?  Starting guile with the `--no-auto-compile` flag will also disable the compile step, although using `primitive-load` is probably the better idea.

andre-senna:
Actually SCMLoader will have other functionalities I'm currently implementing. I though it was a good idea to share the PR early because of scm2pgsql which may be of interest to whoever is playing with loading large data files.

Anyway, I think it's OK to let them in my fork for while. So I'm just closing this PR.

linas:
well, I figured if `load-primitve` worked with OK performance, then  the saql loader could be written in about 10 lines of scheme code, which (to me) seems like it would be easier to maintain and document -- it could be a small cut-n-paste throw-away happy-go-lucky scriptlet... instead of a formal, we-gotta-unit-test-this-and-document-and-maintian-it thing.

PR #2281, opened 2016-07-13
Integrate sentiment analisys into pln psi

ngeiswei:


ngeiswei:
See https://github.com/opencog/opencog/blob/master/opencog/nlp/chatbot-psi/pln-reasoner.scm#L12 of an example.

ngeiswei:
There is a negative sentence, but overall Ben is still happy which is why the implication is inferred.

PR #2286, opened 2016-07-15
Simplify the opencog network server

linas:
This is part one of a larger set of changes, to come.


pennachin:
@linas We're using the cogserver on a contract, so if you plan to make large changes to it, we should discuss those changes and timing.

leungmanhin:
@pennachin, @linas is very busy right now and he is going to make large changes which is critical for the demos... how about creating a stable branch just in case for now?

pennachin:
We are using the stable branch for the contract work, I just don't want to end up with a Cogserver on master that doesn't meet the needs of that project, which will also need OpenPSI, PLN, etc...

bgoertzel:
Hi Cassio, it seems there are serious low-level design flaws in the networking aspects of cogserver which will cause problems for anything that puts a lot of data into the Atomspace rapidly (e.g. visual perception), and LInas is here in New York w. me, and  Man Hin and Eddie, modifying it to work around those errors.   We're trying to get stuff working pretty rapidly now.... I can see that maybe these fixes should be getting done inside some branch besides master, but in any case, based on discussions w. Linas F2F and observing the behaviors Mandeep was seeing w. networking, I am confident the changes are useful, and I don't think they will hurt anything we need for BuildDirect.  Also, I note that OpenPsi and PLN are part of the demo that this stuff is now being repaired to support

pennachin:
Understood, I was just scared by the "large set of changes to come" without any notice or discussion...

linas:
OK, I think I'm done. Everything seems to work -- the only issue I am
seeing is that sometimes,
when attempting a server shutdown while some other process is still sending
it data, sometimes results in a crash.  Which is not a big deal, since the
point of a shutdown is to halt the thing, anyway.  But it is annoying, as
it means the shutdown is somehow not quite clean.

On Fri, Jul 15, 2016 at 12:59 PM, Cassio Pennachin <notifications@github.com

linas:
The other pull reqs that are a part of this series are: #2289 #2290 #2291 #2293 #2294 #2295 #2296  The last one seems to resolve all the remaining issues, as far as I can tall, at the moment.

PR #2302, opened 2016-07-19
Merged rest of Experimental ECAN

rTreutlein:


opencog/attention/experimental/FocusBoundaryUpdatingAgent.cc
```diff
@@ -0,0 +1,99 @@
...
+{
+    if (log)
+        delete log;
+}
+
+void FocusBoundaryUpdatingAgent::setLogger(Logger* _log)
+{
+    if (log)
+        delete log;
+    log = _log;
```
linas: Isn't there a base class that deals with logging?

Also, the C++ convention is that members of a class have underscores in them, whereas method arguments do not.  So the use of underscores here is exactly opposite of the standard C++ convention.

linas: I mean -- this is a very minor issue, its really just an annoyance, nothing else.

opencog/attention/experimental/FocusBoundaryUpdatingAgent.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+Logger* FocusBoundaryUpdatingAgent::getLogger()
+{
+    return log;
+}
+
+void FocusBoundaryUpdatingAgent::run()
+{
+        a = &_cogserver.getAtomSpace();
```
linas: If a is a member, then just get the atomspace in the constructor, instead of every call.

opencog/attention/experimental/FocusBoundaryUpdatingAgent.h
```diff
@@ -0,0 +1,99 @@
...
+#include <opencog/cogserver/server/CogServer.h>
+#include <opencog/cogserver/server/Agent.h>
+
+namespace opencog
+{
+/** \addtogroup grp_attention
+ *  @{
+ */
+
+class CogServer;
```
linas: If you include CogServer.h, then you do not need this stray class declaration.

opencog/attention/experimental/FocusBoundaryUpdatingAgent.cc
rTreutlein: I just copied that but i can fix that for this and the other agents.

opencog/attention/experimental/FocusBoundaryUpdatingAgent.h
```diff
@@ -0,0 +1,99 @@
...
+
+    AttentionValue::sti_t bottomBoundary;
+
+    /** Set the agent's logger object
+     *
+     * Note, this will be deleted when this agent is.
+     *
+     * @param l The logger to associate with the agent.
+     */
+    void setLogger(Logger* l);
```
linas: again, putting this in every derived class is crazy -- just add a logger to the base Agent class.

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+    forget();
+}
+
+void ForgettingAgent::forget()
+{
+    HandleSeq atomsVector;
+    std::back_insert_iterator<HandleSeq> output2(atomsVector);
+    int count = 0;
+    int removalAmount;
+    bool recursive;
```
linas: The standard C++ convention is to only declare variables when you use them, instead of declaring them at the top of the subroutine.  The need to declare variables at the top of a routine was a compiler limitation dating back to the 1980's, and was fixed in most compilers in the 1990's.  Its not a useful style, and makes it harder to read the code.  (This is also where the underbar convention is useful: so, for example `bool recursive` makes it clear that recursive is a local variable, scoped to the local block, whereas `_recursive` would indicate an object-scope variable.

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    a->get_handles_by_type(output2, ATOM, true);
+
+    int asize = atomsVector.size();
+    if (asize < (maxSize + accDivSize)) {
```
linas:
```
int removalAmount = asize - (maxSize - accDivSize);
if (removalAmount < 0) return;
```

opencog/attention/experimental/ForgettingAgent.cc
rTreutlein: Again not code that i wrote but i can fix. For reading up on C++ conventions could you recommend something in addition to http://wiki.opencog.org/w/Development_standards?

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fprintf(stdout,"Forgetting Stuff, Atomspace Size: %d \n",asize);
+    // Sort atoms by lti, remove the lowest unless vlti is NONDISPOSABLE
+    std::sort(atomsVector.begin(), atomsVector.end(), ForgettingLTIThenTVAscendingSort(a));
+
+    removalAmount = asize - (maxSize - accDivSize);
+    log->info("ForgettingAgent::forget - will attempt to remove %d atoms", removalAmount);
+
+    for (unsigned int i = 0; i < atomsVector.size(); i++)
```
linas: You never use the index i for anything, you don't need it.
```
for (AtomPtr& atm: atomsVector)
```

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+    {
+        if (atomsVector[i]->getAttentionValue()->getLTI() <= forgetThreshold
+                && count < removalAmount)
+        {
+            if (atomsVector[i]->getAttentionValue()->getVLTI() == AttentionValue::DISPOSABLE )
+            {
+                std::string atomName = a->atom_as_string(atomsVector[i]);
+                log->fine("Removing atom %s", atomName.c_str());
+                // TODO: do recursive remove if neighbours are not very important
+                IncomingSet iset = atomsVector[i]->getIncomingSet(a);
```
linas:
```
const IncomingSet&
```
The ampersand avoids a copy of the entire contents of the vector,  without the ampersand, the vector is copied, and you really don't need a copy here.

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+                recursive = true;
+                for (LinkPtr h : iset)
```
linas:
```
const LinkPtr&
```
use ampersand, avoid the copy.

opencog/attention/experimental/ForgettingAgent.cc
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+                if (!recursive)
+                    continue;
+
+                atomsVector[i]->setSTI(0);
+                atomsVector[i]->setLTI(0);
+                if (!a->remove_atom(atomsVector[i],recursive)) {
+                    // Atom must have already been removed through having
+                    // previously removed atoms in it's outgoing set.
```
linas: typo: incoming, not outgoing.

opencog/attention/experimental/ForgettingAgent.h
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * These work in concert to limit how much and what atoms are forgotten.
+ * TODO: Improve Recursive Remove to work with links outher then HebbianLinks
+ */
+class ForgettingAgent : public Agent
+{
+
+private:
+    AtomSpace* a;
+    Logger *log; //!< Logger object for Agent
```
linas: Just ass the atomspace and the logger to the base Agent class.  Use the underbar convention.  Don't call it `a` since that is ambigous -- `a` might be an atom. Many of the unit tests use `_as` as the member, also `_atomspace` would be fine.

opencog/attention/experimental/ForgettingAgent.h
```diff
@@ -0,0 +1,142 @@
...
+    {
+        AttentionValue::lti_t lti1, lti2;
+
+        lti1 = h1->getAttentionValue()->getLTI();
+        lti2 = h2->getAttentionValue()->getLTI();
+        if (lti1 != lti2) return lti1 < lti2;
+        else {
+            float tv1, tv2;
+            tv1 = fabs(h1->getTruthValue()->getMean());
+            tv2 = fabs(h2->getTruthValue()->getMean());
```
linas: fabs is weird .. do you really expect the mean to be negative?

rTreutlein: Yea strange (not my code) but it might have to be changed completely. Cause i don't see a reason why something with a lower Mean should be forgotten earlier. Something with less confidence seems to make more sense. I'll ask in the attention channel on Slack.

opencog/attention/experimental/HebbianCreationAgent.cc
```diff
@@ -0,0 +1,173 @@
...
+            get_target_neighbors(source, ASYMMETRIC_HEBBIAN_LINK);
+    HandleSeq existingAsTarget =
+            get_source_neighbors(source, ASYMMETRIC_HEBBIAN_LINK);
+
+    // Get the set differences between the AttentionalFocus
+    // and the sets of existing sources and targets
+    std::sort(existingAsSource.begin(), existingAsSource.end());
+    std::sort(existingAsTarget.begin(), existingAsTarget.end());
+    std::sort(attentionalFocus.begin(), attentionalFocus.end());
+    HandleSeq needToBeSource = set_difference(attentionalFocus,
```
linas: I'm not sure, but I think you could save a bunch of cpu time by using std::set instead of std::vector, here.  sets are rb-trees, and are sorted automatically, which allows you to skip the explicit sort step above.

opencog/attention/experimental/HebbianCreationAgent.cc
```diff
@@ -0,0 +1,173 @@
...
+                                              existingAsSource);
+    HandleSeq needToBeTarget = set_difference(attentionalFocus,
+                                              existingAsTarget);
+
+
+    int count = 0;
+
+    // Resulting in the sets of nodes that require
+    // a new AsymmetricHebbianLink in either direction
+    for (Handle atom : needToBeSource) {
```
linas: const Handle&

opencog/attention/experimental/HebbianCreationAgent.cc
```diff
@@ -0,0 +1,173 @@
...
+        if (atom == Handle::UNDEFINED)
```
linas: surely this can never happen, can it?

opencog/attention/experimental/HebbianCreationAgent.cc
```diff
@@ -0,0 +1,173 @@
...
+    //How many links outside the AF should be created
+    int farLinks = round(count / localToFarLinks);
+
+    Handle link = Handle::UNDEFINED;
+    Handle target = Handle::UNDEFINED;
+
+    //Pick a random target and create the link if it doesn't exist already
+    for (int i = 0; i < farLinks; i++) {
+        target = notAttentionalFocus[distribution(generator)];
+        link = as->get_handle(ASYMMETRIC_HEBBIAN_LINK, source, target);
```
linas: declare the link inside the loop, not outside of it.

opencog/attention/experimental/HebbianCreationAgent.cc
rTreutlein: The Forgetting Agent might delete it while this is running so yes it is possible.

opencog/attention/experimental/HebbianCreationAgent.cc
```diff
@@ -0,0 +1,173 @@
...
+        if (link == Handle::UNDEFINED)
+            addHebbian(source,target);
+    }
+
+    //Check the ammount of HebbianLinks the Atom has
+    IncomingSet iset = source->getIncomingSet();
+    iset.erase(remove_if(iset.begin(),iset.end(),
+                            [](LinkPtr lp) -> bool {
+                            return !(classserver().isA(lp->getType(), HEBBIAN_LINK));
+                            }),iset.end());
```
linas: It would probably be faster to use `getIncomingSetByType`, instead of getting all of them, and then erasing.

opencog/attention/experimental/HebbianUpdatingAgent.cc
```diff
@@ -0,0 +1,155 @@
...
+void HebbianUpdatingAgent::setLogger(Logger* _log)
+{
+    if (log)
+        delete log;
+    log = _log;
+}
+
+Logger* HebbianUpdatingAgent::getLogger()
+{
+    return log;
```
linas: you declare this log everywhere, but, as far as I can tell, you never actually use it. Do you even need the logger?

linas:
Merging.  This all looks fine to me, except for the large number of cosmetic changes that I suggested.  Those are intended only to improve readability and long-term maintainability; they should be fixed, but the code looks OK otherwise, so I will merge.

opencog/attention/experimental/ForgettingAgent.cc
linas: who wrote the code, and why aren't they creating the pull request?  

The wiki page for the opencog development standards is kind of screwy. It makes various bad recommendations, and needs review and updates.  There's no global C++ convention that everyone agrees on; however, in opencog, we use (I use?) mostly a mixture of the linux kernel, and the fsf gnu coding standards (disregarding some of the gnu indentation standards).

opencog/attention/experimental/HebbianCreationAgent.cc
linas: well, except that it is not possible ... Handles are smart pointers that maintain a use-count for each atom, and an atom is "deleted" only when the use-count drops to zero.- -- i.e. when there are no more handles that reference the atom.  

The only thing you directly do with an atom is the remove it from the atomspace.  However, that will not invalidate any pointers to the atom that other threads/agents/algorithms/whatevers might be holding.

linas: ... and since HandleSeq is a std::vector of Handles, the use-count of any atom in the HandleSeq will always be at least one.  The use count will decrement only when the HandleSeq is deleted (or the handle is popped from the vector, or otherwise over-written).

rTreutlein: I thought I had some issue there but yea that should be impossible. What happens if you create a Link to an Atom that's no longer in the Atomspace?

linas: A Link has a `std::vector<Handle>` in it, so when you create that Link, the use-count for each atom in the link will increment by 1.  Thus, as long as you keep around a Handle (or LinkPtr) to the link, all sub-atoms under that link will be protected from deletion. 

The "only" thing that the atomspace does is to keep a pointer to the atom, so that you can drop your own handles, without acccidentally deleting the atoms they point to. ... so as long as the atomspace is pointing at the atom, it won't be deleted.

Issue #2301, opened 2016-07-19
Creating a thread via AgentThreadRunner class hangs the guile shell thread


misgeatgit:
@hedayat,@linas et al. Here are the steps to reproduce this situation: 

- build and load the [this](https://gist.github.com/misgeatgit/4c9cced9b6ba071917d71a7c73fd8edd) module.
- open guile shell via telnet and start inserting random atoms and queries.

After few seconds both the AgentRunnerThread and guile shell hang. Subsequent opening of other guile shells will also result in an immediate hanging. Furthermore, if the python REST server has been run, starts hanging right after guile shell hanging. Seems a deadlock to me.

linas:
so where is the deadlock?

start gdb, wait till it deadlocks, then say `info threads` and `thread 42` and `bt` to gather the backtraces.

my guess is that the agent code is deadlocking -- last time I looked at the agent code in the cogserver, it was in terrible shape, and very buggy -- it probably needs to be completely thrown away and re-written.

hedayat:
@linas You're welcome! I don't remember you pointing to any bugs however. You complained a lot though!
@misgeatgit I'll try to re-produce. But if you can find the bug, it'd be better.

misgeatgit:
I did some debugging and the lock waiting is happening here https://github.com/opencog/opencog/blob/master/opencog/cogserver/shell/GenericShell.cc#L166 and https://github.com/opencog/opencog/blob/master/opencog/cogserver/shell/GenericShell.cc#L193 .

I put the two sections in one lock_guard and the problem was solved. @linas, Since I am not sure that this a cleaner fix, it would be good if you can take this as a hint and fix it.

linas:
you state that the deadlock is located at those two locks, but do not provide any evidence of this. What is the actual cause of the deadlock? Both of those locks protect a very short segment of code, and are released in microseconds, and make no recursive calls. How can they possibly deadlock? What am I missing?

misgeatgit:
@linas, my usage of deadlock might be incorrect but the problem mimics that situation. Anyways, here is the backtrace.

````
(gdb) info threads
  Id   Target Id         Frame 
  17   Thread 0x7fffe4ff9700 (LWP 30395) "cogserver" __lll_lock_wait () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
[...]

PR #2304, opened 2016-07-20
A temporary work around to #2301

misgeatgit:
I am not sure if this is the right way to fix issue #2301 but it does fix it.


linas:
No, this is not the right fix, it will prevent the display of any python printing output on the shell connection.  The python print statement always prints to stdout, which is the shell on which the cogserver is started.  Thus, if a user is running python code that generates any sort of output, it will show up in the cogserver terminal, instead of the shell prompt, leaving the user very confused about what is happening.  If the cogserver haas no terminal, then the output would be lost.  Since python does not offer any sot of mechanism to control i/o (that I know of), the work-around is to always redirect stdout.

So -- let me pose the question again: for bug #2301, what are the stack traces? where is the deadlock?

PR #2305, opened 2016-07-20
Check the truth value of the psi-rule before evaluating the context

leungmanhin:
It should save some cpu time if we skip evaluating the contexts of the rules in each psi-step that are having zero tv-strength (e.g. turned off by using the sliders)


leungmanhin:
@AmeBel please review

AmeBel:
lgtm

PR #2311, opened 2016-07-24
Partial relief of shell evaluation woes

linas:
This is a work-in-progress for relieving the telnet-shell issues that people are reporting.

More work remains to be done.  This includes:
* interrupts (control-C) of infinite loops does not work; infinite loops will hang the system.
* Printing in the python shell is broken, and does not work -- no prints done in python will ever be visible.


linas:
pull req opencog/atomspace#836 provides infrastructure needed for interrupting threads.

linas:
python printing issues being tracked in #2315

linas:
ctrl-C now works, after pull req #2316

PR #2312, opened 2016-07-25
Fix compile error in PyEvalUTest and PythonModuleUTest

ngeiswei:
OpenCog utests are broken, does even compile.

This fixes compiling errors.

Don't merge just yet, more is coming.


ngeiswei:
Actually merging now so I can merge another branch before pursuing work.

PR #2319, opened 2016-07-27
Openpsi

eddiemonroe:
Documentation and directions are in the in the "Modulators and Internal Dynamics" section of the Readme, as well as more detail in the individual files.

To use and test, 
guile> (load "psi-dynamics.scm") ; in ros-behavior-scripting/src. 
$> tail -f psilog.txt  # for output of select openpsi variables

I'll put more info in the pull request for ros-behavior-scripting/src/psi-dynamics.scm.


linas:
I see no objections after a quick skim of the code, however, I don't work on openpsi actively; maybe Amen should review and merge?

AmeBel:
This doesn't affect any of the openpsi code that is in use by chatbot-psi and pis-btree, so merging

PR #2321, opened 2016-07-27
Attention Experiment

rTreutlein:
This code merges my ECAN Experiment it is located in opencog/experiments/attention.

It also improves/fixes some things in ECAN.


examples/attention/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# The atom_types.h file is written to the build directory
+
+# PLNDynamicExpSetupModule
+INCLUDE_DIRECTORIES(${CMAKE_BINARY_DIR})
+
+IF (TBB_FOUND)
```
linas: really, do we have to use TBB? isn't standard C++ enough?

linas: I ask this, because in my experience, using these non-standard vendor libraries tend to induce snake-bites down the line.  Look at the fiasco that is boost for a shining examples of bad code.

examples/attention/Globals.cc
```diff
@@ -0,0 +1,163 @@
...
+{
+    AtomSpace as;
+    as.add_atom(h);
+    Handle friends_predicate = as.add_node(PREDICATE_NODE, "friends");
+    Handle var_1 = as.add_node(CONCEPT_NODE, "$0343O45FFEWW");
+    Handle var_2 = as.add_node(CONCEPT_NODE, "$0343045FFEWY");
+    Handle friend_list = as.add_link(LIST_LINK, HandleSeq { var_1, var_2 });
+    Handle eval_link = as.add_link(EVALUATION_LINK, { friends_predicate,
+                                                      friend_list });
+    as.add_atom(eval_link);
```
linas: Huh ... this is exactly why we have python and scheme bindings (and haskell!) -- its because adding atoms to the atomspace in C++ is verbose and painful, and any changes require a recompile.

examples/attention/SentenceGenStimulateAgent.cc
```diff
@@ -0,0 +1,110 @@
...
+SentenceGenStimulateAgent::~SentenceGenStimulateAgent()
+{
+    //delete _scm_eval;
+}
+SentenceGenStimulateAgent::SentenceGenStimulateAgent(CogServer& cs) :
+        Agent(cs), _as(cs.getAtomSpace())
+{
+    _scm_eval = new SchemeEval(&_as);
+    current_group = 0;
+    startcount = _cogserver.getCycleCount();
```
linas: Do cogserver agents actually work? Last time I tried them, they didn't work; and fixing them looked to be a major undertaking...  I don't believe that infrastructure is currently usable.

examples/attention/opencog.conf
```diff
@@ -0,0 +1,137 @@
+#
+# This file provides an example configuration for a production OpenCog
+# server. Particularly noteworthy is the list of automatically loaded
+# modules.
+#
+SERVER_PORT           = 17001
+LOG_FILE              = /tmp/cogserver.log
+
```
linas: after skimming the above code, I don't believe that "examples" is really the correct directory for this -- examples are really meant to be tutorials for new users, and this seems to be far beyond a tutorial.

Perhaps this code should go into a directory called `opencog/attention/experiments` ? or something like that?  We certainly have other code, similar to this, that lives in the subdirectory of the subsystem that it is experimenting with.

linas: Or even in the `opencog/experiments/attention.` directory would be fine, (or better) as you wrote in the original pull request.  Or even `experiments/attention`, without the leading `opencog` in front, would be better still, perhaps.

linas:
This is OK to merge, either before or after fixing the above-mentioned issues. I don't know how urgent this is, so I'm not merging just yet.

examples/attention/SentenceGenStimulateAgent.cc
rTreutlein: They work fine for me. The ECAN Systems is basically just 1 Module and a bunch of Agents and it works.
Not sure what kind of issues you had with the Agents.

examples/attention/Globals.cc
rTreutlein: damn, I forget to clean out that file. Most of that stuff I can just delete. It's code from @misgeatgit  Experiment that I no longer need.

rTreutlein:
Ill fix these issues first.

examples/attention/opencog.conf
rTreutlein: I'll move it to experiments/attention I just didn't want it to be in opencog/attention/experiments where it would have to be compiled every time when building opencog without needing it most of the time.

examples/attention/CMakeLists.txt
rTreutlein: I'll try to get rid of it. I think I really need to become more sceptical of code I modify. Instead of leaving stuff like it is when I am not quite sure what purpose it serves.

PR #2336, opened 2016-07-31
Fix for bug #2328

linas:
I believe that this should fix bug #2328, although I have not finished testing.

The core issue was this: shell was being destroyed (because netcat closed the socket) even before evaluation of the expression that was sent even started.  As a result, initialization is skipped (the shell destructor tears down the virtual function vtable; initialization was being done in c++ virtual functions).  The initialization set the atomspace; without that, the atomspace is null, leading to a crash.

### DIFF
--- a/opencog/cogserver/server/ServerSocket.cc
+++ b/opencog/cogserver/server/ServerSocket.cc
@@ -57,8 +57,8 @@ void ServerSocket::Send(const std::string& cmd)
     // I beleive this is a ENOTCON errno, maybe its another one as well.
     // Don't log these harmless errors.
     if (error.value() != boost::system::errc::success and
-        error.value() != boost::system::errc::not_connected and
-        error.value() != boost::system::errc::bad_descriptor)
+        error.value() != boost::asio::error::not_connected and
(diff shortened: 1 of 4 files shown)


ngeiswei:
It works for me (15m so far, all components activated, never happened before)!

Respect, man!!! :-)

ngeiswei:
Shit, I think I spoke too fast. I did experienced a crash, unfortunately the core dump is empty, but based on the message display right before the crash I think it probably corresponds to trace https://github.com/opencog/opencog/issues/2328#issuecomment-236268872

Anyway, it's still a very big improvements, probably enough to let me (and hopefully others) focusing on demoing.

ngeiswei:
Yep, re-stumbled on it, I confirm the bt is https://github.com/opencog/opencog/issues/2328#issuecomment-236268872 however it might triggered by my own demo branch. Still trying to figure that out.

linas:
when I was debugging and fixing, I never saw that particular stack trace
(in the unprotect_object code)  I'll try some fancier test cases, and see
what happens.

--linas

PR #2346, opened 2016-08-03
Provide stimulus instead of changing STI value

misgeatgit:


opencog/nlp/chatbot/chat-utils.scm
```diff
@@ -196,12 +196,12 @@ def call_sentiment_parse(text_node, sent_node):
 (define nlp-stimulation-value 0)
 
 ; -----------------------------------------------------------------------
-(define-public (nlp-start-stimulation STI)
+(define-public (nlp-start-stimulation stimulus)
 "
```
linas: one common (but not widely used) scheme convention is to capitalize the arguments to a function. I think it does improve readability, slightly.

linas:
this looks mergable to me.

Issue #2314, opened 2016-07-25
cogserver agent code has ecan stuff in it.


linas:
CogServerUTest is failing because ecan does not work yet.

The cogserver does not have, should not have, any dependency on agents.

The agent design is broken -- agents should not have any ecan stuff in them.

If some agent is needed that does require ecan, then that code should be in some derived class, not in the base class.  The derived class should not be in the cogserver folder, it should be in the attention folder.

linas:
Also, AgentUTest is failing for the same reason.

linas:
Also, PythonModuleUTest ForgettingAgentUTest and ImportanceUpdatingAgentUTest are all failing -- I assume that they're all failing due to the above, albeit indirectly.

rTreutlein:
When we replace the old ECAN with the new currently experimental code, removing that ECAN stuff will be significantly easier. I already tried and right now it would require a bunch of dynamic_casts which I couldn't get to work easily.

Although I am not sure how that effects the tests as this code has been in the Agent Class for a long time now.

linas:
what `dynamic_casts` are you hitting?

I'm objecting to the agent class having an `_attentionValue, _totalStimulus _stimulatedAtoms _STIAtomWage _stiFundsBuffer` etc. coded into it --- it seems like this should go into a distinct base class that the attention-allocation agents would derive from, rather than being part of the core Agent class.

linas:
I removed some of the cruft here: #2349 -- it looks like the new "experimental" code duplicated what I removed, anyway.  That still leaves `_attentionValue` etc in the base class.

rTreutlein:
Well the in the old ECAN the Agents kept track of which Atoms they stimulated. The ImportanceUpdating Agent then gets all the running agents from the Cogserver to convert the stimulus into actual STI. If i put all the ECAN related code into a AttentionAgent Class I would have to cast the AgentPtrs I get from the cogserver to AttentionAgentPtr.
But since the new experimental code does things differently we can just remove all that once we get rid of the old ecan System.

linas:
OK.

linas:
BTW, when might the new system be ready?  Why can't we discard the old system now?  Does it serve any purpose, at all, to keep around the old system?

linas:
Closing; the Attention/ECAN code was removed from the base Agent class.  This was done in #2349

PR #2349, opened 2016-08-04
Remove dead code from Agent

linas:


linas:
This (partly) fixed issue #2314

PR #2358, opened 2016-08-11
Add a new function for reading properties files for AIML chatbot

leungmanhin:
for reading those properties files for different HR robots directly and convert them into atoms


linas:
OK, for just this once, but please lets not do things like this, in general. Its just ugly and wrong, and goes against the spirit of the whole system.

PR #2363, opened 2016-08-15
Change the default accum weight to zero in aiml chatbot

leungmanhin:
this is causing it to pick the first rule in the list almost all the time

@linas would it be ok to simply set the default value to zero, or change it to something smaller?

### DIFF
--- a/opencog/nlp/aiml/aiml.scm
+++ b/opencog/nlp/aiml/aiml.scm
@@ -316,7 +316,7 @@
 
 	;; Return #t for the first rule in the RULE-LIST for which the
 	;; accumulated weight is above THRESH.
-	(define accum 0.000000001)
+	(define accum 0.0)
 	(define (pick-first ATOM THRESH)
 		(set! accum (+ accum (get-weight ATOM)))
 		(< THRESH accum))


linas:
Yeah, I guess. This was non-zero to make sure that sometimes, the *last* atom would be picked.  Otherwise, the last atom is never picked .... you may want to double-check that list-index at line 331 does not go crazy if pick-first always returns false.  Here:
```
         (list-ref RULE-LIST (list-index
            (lambda (ATOM) (pick-first ATOM thresh)) RULE-LIST))
```

so if `(lambda (ATOM) (pick-first ATOM thresh))` always returns false, will  `(list-index #f RULE-LIST)` always return the last element of the list?

linas:
Hmm, according to this, http://srfi.schemers.org/srfi-1/srfi-1.html#list-index  -- it says that if the `pick-first` always returns false, then `list-index` will return #f ... whereas, we want it to return the last member of the list. So setting the accum weight to 0.0 will cause this code to sometimes throw an exception, and that will need to be fixed ...

linas:
However, you may get lucky for a long time, and not hit the exception. Depends on the random number, and the weights.

leungmanhin:
I see, https://github.com/opencog/opencog/pull/2367 should fix the problem

PR #2372, opened 2016-08-16
Maybe a fix for #2365

### LINKED ISSUE #2365: AIML chat fails sometimes when it receives "fair enough" as input
it failed with this error
```
ERROR: In procedure opencog-extension:
ERROR: Throw to key `C++-EXCEPTION' with args `("cog-execute!" "Not a link! (/home/leungmanhin/hansonrobotics/opencog/atomspace/opencog/atoms/base/Atom.h:206)")'.
```

it seems that it is related to this psi rule
```
(MemberLink
   (ImplicationLink (stv 1 6.1289907e-06)
      (AndLink
         (AndLink
            (EvaluationLink
               (PredicateNode "*-AIML-pattern-*")
               (ListLink
                  (GlobNode "$star-1")
                  (WordNode "enough" (stv 5.5117674e-05 0.0024937657))
               )
            )
         )
         (ListLink
            (ExecutionOutputLink
               (DefinedSchemaNode "AIML-tag srai")
               (ListLink
                  (ListLink
                     (GlobNode "$star-2")
                  )
               )
            )
            (ExecutionOutputLink
               (DefinedSchemaNode "AIML-tag think")
               (ListLink
                  (ListLink
                     (ExecutionOutputLink
                        (DefinedSchemaNode "AIML-tag set")
                        (ListLink
                           (ConceptNode "topic" (stv 0.0018315018 0.0012484394))
                           (ListLink
                           )
                        )
                     )
                  )
               )
            )
         )
      )
      (ConceptNode "AIML chat subsystem goal")
   )
   (ConceptNode "OpenPsi: AIML chat demand" (stv 0.97000003 1))
)
```

linas:
This might fix #2365 -- not sure, just a wild guess. @leungmanhin if this is doing the right thing, then please merge; else, twiddle it so that it does work.  (I"m not merging because I cannot (easily) test)

### DIFF
--- a/opencog/nlp/aiml/aiml.scm
+++ b/opencog/nlp/aiml/aiml.scm
@@ -496,11 +496,12 @@
 
 (define-public (do-aiml-set KEY VALUE)
 	(define flat-val (word-list-flatten VALUE))
+	(define reval (if (equal? '() flat-val) (WordNode "") flat-val))
 	; (display "Perform AIML set key=") (display KEY) (newline)
-	; (display "set value=") (display flat-val) (newline)
+	; (display "set value=") (display reval) (newline)
 	(define rekey (Concept (string-append "AIML state " (cog-name KEY))))
-	(State rekey flat-val)
-	flat-val
+	(State rekey reval)
+	reval
 )
 
 ; AIML-tag get -- Fetch value from a StateLink key-value pair


leungmanhin:
I think this is probably not needed, the crash as stated in #2365 was caused by the `(GlobNode "$star-2")` and it is fixed by #2373, thanks for the fix!

However maybe there is one potential problem, when the system first starts or no topic has been set so far, there is no `(State (ConceptNode "AIML state topic") ...)` in the atomspace, so whenever we do `(do-aiml-get (Concept "topic"))` in `is-topical-rule?` (https://github.com/opencog/opencog/blob/master/opencog/nlp/aiml/aiml.scm#L280), it returns an empty scheme list '(). But after executing things like
```
(ExecutionOutputLink
    (DefinedSchemaNode "AIML-tag set")
        (ListLink
            (ConceptNode "topic" (stv 0.0018315018 0.0012484394))
            (ListLink
            )
        )
    )
)
```
`(do-aiml-get (Concept "topic"))` will return an empty `ListLink` instead, this may affect the result returns by `is-topical-rule?`. So to be consistent maybe we should either create `(State (ConceptNode "AIML state topic") (List))` by default, or remove it whenever we want to unset the topic?

linas:
YYes, we should either create it by default, or remove it whenever we want to unset the topic.  Not sure which.  Maybe create it by default is the best?

leungmanhin:
agree, https://github.com/opencog/opencog/pull/2378 will create it by default

linas:
merging, seems like the right thing to do, in any situation...

PR #2373, opened 2016-08-16
Fix for #2365

linas:
This fixes the star-counting problem in bug #2365.  However, it might
maybe introduce bugs with multi-star expressions. I have not thoroughly
tested this.  This will probably result in fairly large change in
chatbot behavior! As I suspect that this may be a wide-spread bug.


linas:
I suspect that this is a big, important fix, that ill dramatically alter how AIML is handled ...

However, I am not merging right now cause I don't really have time to test and double-check. @leungmanhin please test and merge or tell me what else got broken by this.

leungmanhin:
merging this, it seems that there is no observable negative impact to the system (so far), and it does fix #2365, thanks!

PR #2385, opened 2016-08-23
Handle wildcard that may exist in <that> tag in AIML

leungmanhin:


opencog/nlp/aiml/aiml.scm
```diff
@@ -298,7 +298,9 @@
 (define (satisfy-that? RULE)
 	(define pred (get-pred RULE "*-AIML-that-*"))
 	(if (null? pred) #t
-		(equal? (do-aiml-get (Concept "that")) (gdr pred)))
+		(or (equal? (do-aiml-get (Concept "that")) (gdr pred))
+			(not (null? (gar (cog-execute! (MapLink (gdr pred)
+				(Set (do-aiml-get (Concept "that"))))))))))
 )
```
linas: I think I understand what this does, but it would be nice to add a comment that explains this, so that future maintainers will also be able to understand.

leungmanhin: ok I'll add comments to the code

PR #2388, opened 2016-08-24
Use cog-fc instead of cog-bind

misgeatgit:
cog-bind works on whole atomspace  whereas cog-fc can take a set of atoms as a focus set or work on whole atomspace. This feature is important for integrating ECAN with the chat system.


ngeiswei:
Cool! So did you manage to reproduce the demo after this change?

misgeatgit:
Yes, it works.

ngeiswei:
Cool, it shows the forward chainer has reached a good level of maturity.

PR #2389, opened 2016-08-24
Support "input index" and "that index" in AIML

leungmanhin:


opencog/nlp/aiml/aiml.scm
```diff
@@ -468,8 +467,19 @@
 	(let ((response (word-list-flatten (do-while-same SENT 5))))
 
 		; The robots response is the current "that".
+		; Store up to two previous inputs and outputs
+		; XXX TODO: Would be better to log and retrieve the chat
+		;           history using AtTimeLink and the time server
 		(if (valid-response? response)
-			(do-aiml-set (Concept "that") response))
+			(let ((that (do-aiml-get (Concept "that"))))
```
linas: I would very much prefer it if you used `(Concept "AIML state that")` and `(Concept "AIML state input")` for these, as it makes debugging easier.  The problem is that `(Concept "that")` is very generic, 

It is also hard to do a string-search for "that" in the source code, yeilding a lot of false positives.  Ability to grep source code is important.

linas: never mind!  I see that `do-aiml-get` already does this, automatically.   Kind of weird, though :-(

linas: OK, I see -- it needs to be done this way, so that `(DefinedSchema "AIML-tag get")` works correctly. So again, never mind

PR #2393, opened 2016-08-25
Update the list for doing 'person' and 'person2' substitution 

leungmanhin:
be consistent with the one in [HEAD] (https://github.com/hansonrobotics/HEAD/blob/master/src/chatbot/src/chatbot/aiml/DefaultSubs.py)


linas:
thanks

PR #2394, opened 2016-08-25
Do simple word substitution before doing AIML matching

leungmanhin:
also to be consistent with the one in [HEAD] (https://github.com/hansonrobotics/HEAD/blob/master/src/chatbot/src/chatbot/aiml/DefaultSubs.py)


linas:
very interesting!

PR #2396, opened 2016-08-26
Filter the usable AIML pattern rules

leungmanhin:
Pattern matching in the standalone AIML is like a tree search, it starts with the 1st word, and then 2nd, 3rd... and so on until there is a mismatch. So for example if there are two rules
```
<pattern>THERE IS A *</pattern>
<pattern>THERE IS *</pattern>
```
and the input is "There is a cat", it will always choose the first one and never choose the second. In this case I think OC AIML should also do the same in order to behave similarly


linas:
right, yes, thanks

PR #2399, opened 2016-08-29
Fix process_set

leungmanhin:
I notice that when it converts a rule with a set tag arranges in a way like this:
```
<template>
FIRST<set name="topic">SECOND</set>THIRD
</template>
```
The text after `</set>` is discarded, this would fix the problem

### DIFF
--- a/opencog/nlp/aiml/import/aiml2psi.pl
+++ b/opencog/nlp/aiml/import/aiml2psi.pl
@@ -569,7 +569,7 @@ sub process_set
 	my $text = $_[1];
 	my $tout = "";
 
-	$text =~ /(.*?)<set name='(.*?)'>(.*)<\/set>(.*?)/;
+	$text =~ /(.*?)<set name='(.*?)'>(.*)<\/set>(.*)/;
 
 	$tout .= &split_string($indent, $1);
 	$tout .= $indent . "(ExecutionOutput\n";


linas:
Well, I just merged this, but it's not obviously right.  So: here's the idea:  The question mark is the non-greedy matcher -- see http://perldoc.perl.org/perlre.html section "Quantifiers" about 1/4th down.  For your example, the regex with and without the question mark should both behave *exactly* the same way. So I am very surprised by your claim that it fixes something.

The point of the greedy/non-greedy thing was to deal with nesting, and with multiple matches. So, I guess `<set>` will never nest, but what about this:
```
FIRST<set name="topic">SECOND</set>THIRD<set name="user">FOURTH</set>FIFTH
```
Please make sure that the topic is not accidentally set to `SECOND</set>THIRD<set name="user">FOURTH`

I'm thinking the right regex might be 
```
$text =~ /(.*?)<set name='(.*?)'>(.*?)<\/set>(.*)/;
```
Of course, its the opposite, if `<set>` can be nested (which it can't but other tags can .. e.g. random)  

This is one reason the code is hacky and ugly ....

leungmanhin:
Yes you are right, the topic is incorrect if there are more than one `<set>`, and your regex fixes it, I'll update the regex

leungmanhin:
I am not sure if `<set>` can be nested, but I do found at least one rule that has nested `<set>`, which is in `bot.aiml`
```
<category><pattern>DO YOU LIKE *</pattern>
<template>
<think>      <set name="it">       <set name="topic">        <person/>       </set>      </set>     </think>     
<random>  
<li>I like <bot name="forfun"/>.</li>     
<li>I like <bot name="friend"/>.</li>      
<li>I like <bot name="kindmusic"/>.</li>      
<li>I like <bot name="favoriteband"/>.</li>      
<li>I like <bot name="favoritesong"/>.</li>      <li>I like <bot name="favoritebook"/>.</li>      
<li>I like <bot name="favoritefood"/>.</li>      
<li>I like <bot name="favoritecolor"/>.</li>      
<li>I like <bot name="favoritemovie"/>.</li>      
<li>I like <bot name="favoriteartist"/>.</li>      
<li>I like <bot name="favoriteauthor"/>.</li>      
<li>I like <bot name="favoriteactor"/>.</li>      
<li>I like <bot name="favoriteactress"/>.</li>      
<li>I like <bot name="talkabout"/>.</li>      
<li>I like <bot name="celebrity"/>.</li>      
<li>I really like <bot name="favoriteshow"/>.</li>      
<li>I really like the <bot name="religion"/> faith.</li>     
</random></template>
</category>
```
and the script cannot convert it correctly...

linas:
Yep. right.  I think the original regex, before you changed it, does do this nested set correctly. That is why the final, trailing bit was non-greedy, so that the middle bit would pull together the nested part. 

I think that this problem with perl regex and nesting and matching eventually lead me to somehow change the design of the thing, but I no longer remember just how I did that.  (its really really late here)

leungmanhin:
Hmm.. even the original regex could not handle this rule correctly, I mean it could handle the nested set  correctly, even with my changes (?)... but it skips the rest of the text within the <random> tag, I'm looking into it

linas:
OK, so I am guessing that pull #2404 fixes this last issue.

PR #2402, opened 2016-08-30
Additional flag argument to fuzzy matching function

misgeatgit:
Used for matching atoms only with in the AF.  Depends on PR [#886] (https://github.com/opencog/atomspace/pull/867).


misgeatgit:
@leungmanhin please review this.

leungmanhin:
thanks

PR #2404, opened 2016-08-30
Fix process_tag in aiml2psi

leungmanhin:
There is a problem when converting a rule that will call `process_tag`more than one time, e.g.
```
<template>
  <think>
    <set name="topic">
      <person><star/></person>
    </set>
  </think>
  <random>
    <li>I like <bot name="forfun"/>.</li>
    <li>I like <bot name="friend"/>.</li>
  </random>
</template>
```
`<think>` will be handled in `process_tag`, the text within `<random>` is assigned to `$3` and is supposed to be sent back to `process_aiml_tags` for further processing. However `<person>` is also handled in `process_tag` and handling `<person>` happens before sending `$3` to `process_aiml_tags`, and calling `process_tag` again seems to reset the value in `$3`, and as a result the text within `<random>` will be lost. This PR should fix this problem

### DIFF
--- a/opencog/nlp/aiml/import/aiml2psi.pl
+++ b/opencog/nlp/aiml/import/aiml2psi.pl
@@ -531,6 +531,10 @@ sub process_star
 	$tout;
 }
 
+my @pt1 = ();
+my @pt2 = ();
+my @pt3 = ();
+
 # process_tag -- process a generic, un-named tag
 #
 # First argument: the tag name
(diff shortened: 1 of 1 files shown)


opencog/nlp/aiml/import/aiml2psi.pl
```diff
@@ -545,17 +549,24 @@ sub process_tag
 
 	$text =~ /(.*?)<$tag>(.*?)<\/$tag>(.*)/;
 
-	$tout .= &process_aiml_tags($indent, $1);
+	push(@pt1, $1);
+	push(@pt2, $2);
+	push(@pt3, $3);
+
```
linas: Huh. I guess maybe this works, but a much simpler solution is possible: just make a copy -- so
```
my $first = $1;
my $middle = $2;
my $rest_of_it = $3;
$tout .= &process_aiml_tags($indent, $first);
...
$tout .= &process_aiml_tags($indent, $middle);
if ($rest_of_it ne "")
{
    $tout .= &process_aiml_tags($indent, $rest_of_it);
}
```

or just call it `my $t1, $t2, $t3,` etc.

linas: I mean, once you make a copy, the the values should not get garbled up.

There are more places in the code which make this error....

leungmanhin: Oh right that would be better, I'll update it and hunt down the rest in the code before the demo

leungmanhin: Updated in https://github.com/opencog/opencog/pull/2410

linas: thanks

PR #2407, opened 2016-09-01
Handle multiple and nested random tags in aiml2psi

leungmanhin:
this extends the code for handling random tags so that it is able to handle those multiple and nested random tags in the set of rules we are using


linas:
Huh. Looks like code that I would write.  Good job.

leungmanhin:
Thanks Linas

PR #2409, opened 2016-09-01
Utilities and cleanup for openpsi-pln demo

AmeBel:


linas:
I don't know how good the PLN answers are, but ... if it does sometimes give a good answer, but it takes too long to get it, then it would be useful to launch multiple PLN threads, each with some different random seed.

Do you have a collection of PLN results, both good and bad, for "typical" chat scripts?  Could you (or someone) collect up a bunch of these PLN results?  I'd like to know exactly how good/bad they look, and also, I'd like to know how often they come up with answers, as opposed to not having one at all ...

AmeBel:
The example is too crude. It uses a single-rule and restricted on sentences that have "people" and some other words like "happy" in them to be triggered. And is not making the many-world exploration just merely picking one of the worlds that are found at the first level of inference. Thus, speed has not been an issue.

It also lays out the pipeline to from face-recognition +  who-spoke-what  to pln and back to the nlg. Haven't tested this part yet.

__TODO__ 
1. test with face_recognition + who-spoke-what 
2. making it more robust so as to use more rules, and make inferences in broader range of sentences. 
URE's forward-chaniner has the features required for that but haven't been used in this case yet.
3. Collect PLN inference results.

ngeiswei:
Obviously the pre-processing rules `sentiment-sentence-to-person-l2s-rule` and `unary-predicate-speech-act-l2s-rule` are complete hacks and should be replaced by PLN rules or macros, using extra assumptions, for instance

1. Most concepts and predicates involved in NLP are cohesive (as Ben coined), that is if P(X) holds and Y is similar to X to some degree then P(Y) holds to some degree.
2. Or, that instances of the same concepts tend to be similar (yesterday's Ben is similar to today's Ben).
3. etc

If we do that, + fix all the R2L inconsistencies, then I think it should work.

PR #2412, opened 2016-09-06
Declare the type of atoms to be grounded by using GetLink in chatbot-psi and aiml

leungmanhin:
To avoid similar problems that might exist as described in https://github.com/opencog/atomspace/issues/863


linas:
As mentioned on opencog/ros-behavior-scripting#97 -- these changes are probably mostly harmless, but really are not needed, and if they are changing behavior, then there is a bug somewhere.  The StateLink has only one unique value -- so if its the wrong type, then some setter somewhere is setting it to the wrong value.

PR #2418, opened 2016-09-07
ECAN

misgeatgit:
- Optimize source vector building method -  get_all_atoms will only be called once to build the diffusion source vector and from that onwards, AtomAdded and AtomRemoved signals are used to maintain the diffusion source vector.

- Remove MinMax STI updating agent in favor of doing it inside the AttentionBank class

- Modify target selection algorithm for WholeAtomSpace Rent Agent so that instead of relying on the get_all_atom function for choosing atoms, it uses importanceIndex  to pick atoms from certain random range of STI values.


bgoertzel:
Misgana -- you say " get_all_atoms will only be called once" but in my opinion get_all_atoms should be called exactly ZERO times...

misgeatgit:
@bgoertzel Are you suggesting to randomly select atoms and build their incident vector separately?

bgoertzel:
@misgeatgit  maybe I am confused....   It seems OK to "get all Atoms" that are in the AttentionalFocus at a certain point in time.   But there should never be any need to "get all Atoms" from  the whole Atomspace....  So if you are talking about the "whole Atomspace" ECAN, then yeah, I think you should not "get all Atoms" from the whole atomspace, but should instead randomly select atoms and build their incident vectors separately...

misgeatgit:
Yes, I just remembered we discussed this long ago. I will change  it.

misgeatgit:
I want to close this PR and make another one in accordance with the above discussion.

PR #2437, opened 2016-09-17
Fix NLP learn configs

jswiergo:


linas:
OK I merged, .. except that I don't expect that anyone uses these files, except for me... and the IP addresses were correct for my network.

PR #2450, opened 2016-09-22
chat-ecan

misgeatgit:
This PR is the following of closed PR #2418. Changes are made as per the discussion made in that particular PR. Basically, now target selection is done via random selection of atoms from Importance Index.


linas:
I merged cause it was sitting for a while. I'm thinking it would be best if you and @rTreutlein  reviewed each others code, because I am not really looking carefully at what is being done here.  I'm trusting that its the "right thing".

Issue #2315, opened 2016-07-25
python printing is broken


linas:
Printing to the python shell is broken.  This got broken by pull request #2311, the refactoring of the network code is not yet complete. 

Tor reproduce:  telnetto the cogserver, enter the python shell, and execute any python code -- nothing is printed.

linas:
So, there is now a partial fix in pull req opencog/atomspace#837 -- but more work in the cogserver shell is needed.

linas:
The original issue is described here: opencog/atomspace#629  and its probably an issue again...

linas:
I'm pretty sure that issue #1596 is an issue again.  Basically, at this time, all python prints are going to the cogserver shell.

linas:
This commit: https://github.com/opencog/atomspace/commit/32e8bf01beeb415748151f4ef4200dc56d94a48d

trashed the previous python stdout-capture code.  This was done because python agents or any python inf loop would hang the cogserver, because that code was waiting forever for the output to arrive, slowly growing over time, per comment https://github.com/opencog/opencog/issues/2301#issuecomment-235012670

The right fix might be to capture stdout in a different python thread, and forward that to the shell prompt. This would require restoring the above commit, and changing it to multi-task and send the print results back in dribbles.

Alternately, PERFORM_STDOUT_DUPLICATION could be re-enabled in the GenericShell.cc

Dagiopia:
how about redirecting stdout to a pipe and then reading to a variable like _result
I got i working <a href="https://github.com/Dagiopia/atomspace/blob/master/opencog/cython/PythonEval.cc">here</a> 

```

/*  REDIRECTING STDOUT TO A PIPE
    because no string representation can be found for the PyObject returned by 
    PyRun_StringFlags after execution of expression, the only option I can see
    is a temporary hack to redirect the stdout to a variable for the brief
    moment PyRun_StringFlags is running and then reconnecting it back 
    when which we can read from that variable to _result so that the execution 
    result is displayed on the users shell and not on the stdout of cogserver
*/
    int _output_pipe[2];
    int _stdout_original = dup(STDOUT_FILENO);

    if(pipe(_output_pipe) == 0)
        #define _PIPE_OPENED_OKAY_
    #ifdef _PIPE_OPENED_OKAY_
        dup2(_output_pipe[1], STDOUT_FILENO); //redirecting stdout to pipe
        close(_output_pipe[1]);
    #endif

    //execute python script
[...]

PR #2458, opened 2016-10-11
Fixing issue #2315

Dagiopia:
There was a method that stopped working to redirect stdout to a pipe and drain with another thread in GenericShell as a workaround for the Python direct printing to the stdout.

I relocated that part to atomspace/opencog/cython/PythonEval.cc to get it working

### DIFF
--- a/opencog/cogserver/shell/GenericShell.cc
+++ b/opencog/cogserver/shell/GenericShell.cc
@@ -135,8 +135,7 @@ void GenericShell::set_socket(ConsoleSocket *s)
 
 /* ============================================================== */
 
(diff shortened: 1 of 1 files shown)


linas:
`PythonEval.cc` is the wrong place to fix the issue, as explained there.

Let me repeat: there is no network socket in `PythonEval.cc` and so it makes no sense at all to move the `dup2` code there -- you cannot possibly copy stdout to the network socket from that location, because the network socket does not even exist!  The code in atomspace does not do any networking; the network shell only exists  inside the cogerver ...  There is no issue, no bug, if one does not use the cogserver.

PR #2469, opened 2016-11-07
Split implications into scope and non scope for NLP

ngeiswei:
Implications splitting companion for the NLP/OpenPsi code.

I'm not entirely sure about that one so please review and merge. I pushed it on upstream in case you need to work on it before merging.

### DIFF
--- a/opencog/nlp/Tutorial
+++ b/opencog/nlp/Tutorial
@@ -301,7 +301,7 @@ Semantic normalization aka Semantic Triples
          THEN ^3_$phrase($var2, $var1)
 
    Above is actual rule: it is converted to an OpenCog 
-   ImplicationLink (quite verbose!) and run through pattern matcher.
+   ImplicationScopeLink (quite verbose!) and run through pattern matcher.
 
 -- Noteworthy: Makes use of "processing anchors" i.e.
    the sentence $sent MUST be connected, via ListLink to 
--- a/opencog/nlp/aiml/README.md
+++ b/opencog/nlp/aiml/README.md
@@ -425,7 +425,7 @@ psi-get-dual-rules calls psi-get-member-links
 ))
 
 (cog-execute! (Map
-	(Implication
+	(ImplicationScope
 		(ListLink (Word "who") (Word "supports") (Glob "$star-1"))
 		(ExecutionOutput (DefinedSchema "AIML-tag srai")
 			(ListLink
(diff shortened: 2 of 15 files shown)


opencog/nlp/aiml/README.md
```diff
@@ -425,7 +425,7 @@ psi-get-dual-rules calls psi-get-member-links
 ))
 
 (cog-execute! (Map
-	(Implication
+	(ImplicationScope
```
linas: this use is not meant to be scoped.

opencog/nlp/aiml/aiml.scm
```diff
@@ -265,7 +265,7 @@
 ; XXX Need to handle that, topic rules as appropriate.
 (define (run-pattern-rule RULE SENT)
 	(define maplk (MapLink
-		(ImplicationLink
+		(ImplicationScopeLink
 			(gdr (get-pred RULE "*-AIML-pattern-*"))
```
linas: I'm pretty sure this use is not meant to be scoped.  I think that this is supposed to be the same as what openpsi uses, which is not scoped.

opencog/nlp/relex2logic/post-processing.scm
```diff
@@ -252,7 +252,7 @@
 		(list
 			(ForAllLink
 				(VariableNode "$X")
-				(ImplicationLink
+				(ImplicationScopeLink
```
linas: this does not look right -- I think the variable $X is supposed to be that in the ForAllLink -- this implication link is assumed to be not scoped.

opencog/nlp/relex2logic/rule-helpers.scm
```diff
@@ -631,7 +631,7 @@
 	(cond
 		((or (string=? determiner "those") (string=? determiner "these"))
 			(ListLink
-				(ImplicationLink
+				(ImplicationScopeLink
```
linas: this looks correct -- it looks like a classic example of where scoping is needed and is important.

opencog/nlp/Tutorial
```diff
@@ -301,7 +301,7 @@ Semantic normalization aka Semantic Triples
          THEN ^3_$phrase($var2, $var1)
 
    Above is actual rule: it is converted to an OpenCog 
-   ImplicationLink (quite verbose!) and run through pattern matcher.
+   ImplicationScopeLink (quite verbose!) and run through pattern matcher.
```
linas: then intended use here is the unscoped variant.

linas:
Closing -- pul req #2470 makes the only change that seems to actually be needed.

PR #2470, opened 2016-11-07
Alternate to pull req #2469

linas:
I beleive that only one change is needed for ImplicationScopeLink,
and even that change is somewhat dubious...


ngeiswei:
Great, thanks.

PR #2482, opened 2016-11-16
Hack pattern miner

ngeiswei:
Add a utest for the pattern miner under `tests/learning/PatternMiner`. It tries to run `TestPatternMinerAgent` which is entangled in the pattern miner code. For now the utest crashes in the middle. The goal is to fix it and then disentagle TestPatternMinerAgent from the pattern miner and move it into the utest. And of course add more utests as desired.


linas:
Hey, if pattern miner is hacked to remove the dependency on the cogserver, it should probably get moved to the atomspace repo.

PR #2485, opened 2016-11-17
Disable pattern miner utest

ngeiswei:
This is the only unit test that doesn't pass. I'm disabling it since it
might only start passing after a while.


linas:
Only?  PLNRulesUTest has not passed in months. Not passing right now.

ngeiswei:
They all pass for me, although I should say not all components are built on my system
```
The following components WILL NOT be built:
-----------------------------------------------
   REST             - REST Atomspace Publisher module.
   SpaceTime        - 3D Space-Time object tracker.
   Spatial          - Vizualisation of the 3D spacemap.
   NLP              - Natural Language Processing.
```

linas:
Well, aren't you actively maintaining PLNRulesUTest ? I'll open a different bug, then.

linas:
Bug #2487

PR #2489, opened 2016-11-20
Lojban

rTreutlein:
This merges the Lojban <-> Atomese Translator I have been working on into OpenCog Repo.

The easiest way to use it is to load the LojbanModule for the Cogserver, then you can easily translate Lojban to Atomese with the command:

parse-lojban mi jimpe ti


rTreutlein:
retest this please

rTreutlein:
retest this please

linas:
can you please unmerge this?  you should not be checking in large files into this repository, its not designed for that.  Large data files should go into the data repo, or maybe a brand-new lojban repo.

rTreutlein:
@linas I will be moving the files into their own lojban-data Repo. Unfortunately, they will remain in the history. They could be removed from history locally and then force pushed onto the repo of course that's not really a good option. So I am asking you if you think that would be necessary.

PR #2496, opened 2016-11-24
stub out broken code

linas:
Please test your code before merging it into master. Right now, it does not even compile.


rTreutlein:
The code is tested and builds fine for me locally and in Jenkins. Also if you have an issue could you please say what kind of error you are getting so I can fix it.

PR #2500, opened 2016-11-25
Bugfix in SuReal

leungmanhin:
Sometimes it skips the disjunct-match and generates incorrect sentences as a result, this should fix the problem


leungmanhin:
microplanning is failing, I'll open another PR once I fix it

PR #2507, opened 2016-11-29
Haskell build cleanup

AmeBel:
@rTreutlein please reveiw


rTreutlein:
@AmeBel why did you close this? The idea behind these changes seems to good. I would leave in the
#Force Rebuild
 rm -fr ".stack-work/"
because for some reason Jenkins doesn't properly rebuild if something changed.

AmeBel:
If you have `rm -fr ".stack-work/"` then during install stack build is run resulting in dependencies being downloaded as root, which is the issue I was trying to fix by this pr. But, I missed updating the test.

PR #2510, opened 2016-12-01
Assorted openpsi refactoring

linas:
Assorted refactoring of the openpsi code -- almost no functional changes, except for one: the action selector now uses a weighted rule selection, so that more heavily-weighted rules are more likely to be chosen.  Most of the other changes are just organizational.

### DIFF
--- a/opencog/openpsi/utilities.scm
+++ b/opencog/openpsi/utilities.scm
@@ -31,6 +31,21 @@
 )
 
 ; --------------------------------------------------------------
+(define (sort-by-weight RULE-LIST WEIGHT-FN)
+"
+  sort-by-weight RULE-LIST WEIGHT-FN
+
+  Take RULE-LIST and sort it, according to the weights assigned by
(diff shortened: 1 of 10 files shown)


opencog/openpsi/utilities.scm
```diff
@@ -10,9 +10,11 @@
 (use-modules (opencog) (opencog exec))
 
 ; --------------------------------------------------------------
+; XXX TODO: does this really need to be public?
 (define-public psi-prefix-str "OpenPsi: ")
```
AmeBel: No, it doesn't.

opencog/openpsi/utilities.scm
```diff
@@ -10,9 +10,11 @@
...
 
 ; --------------------------------------------------------------
+; XXX TODO: does this really need to be public?
 (define-public (psi-suffix-str a-string)
```
AmeBel: Yes, it is used by chatbot-psi.

PR #2511, opened 2016-12-01
remove old file

eddiemonroe:


linas:
Hmm. Earlier today, I moved code from `updater.scm` to this file, because it looked like it did not belong in `updater.scm` (its PAU code, and all PAU code should be in ros-behavior-scripting, I think)

PR #2521, opened 2016-12-12
add chatscript.scm to CMakeLists

eddiemonroe:
@leungmanhin, does chatscript.scm need to be added to CMakeLists here?

Without it, I get an error in the cogserver scheme shell:
opencog> scm
Entering scheme shell; use ^D or a single . on a line by itself to exit.
guile> (use-modules (opencog nlp chatbot-psi))
. . .
ERROR: In procedure primitive-load-path: Unable to find file "opencog/nlp/chatbot-psi/chatscript.scm" in load path
ABORT: misc-error


leungmanhin:
yes, thanks!

PR #2558, opened 2016-12-28
Port the Attention agents to the new AttentionBank API

linas:
This requires an atomspace with pull req https://github.com/opencog/atomspace/pull/1051 applied to it.


linas:
jenkins can't pass until atomspace is updated.  I guess this happens nightly?

PR #2578, opened 2017-01-06
ECAN refactoring

misgeatgit:
-  Some more refactorings
-  Plugged in the Stochastic Diffusion calculator for WAImportanceDiffusion


opencog/attention/WAImportanceDiffusionAgent.cc
```diff
@@ -25,18 +25,21 @@
...
+#include <opencog/attentionbank/StochasticImportanceDiffusion.h>
 
 #include "WAImportanceDiffusionAgent.h"
 
 using namespace opencog;
 
 
 WAImportanceDiffusionAgent::WAImportanceDiffusionAgent(CogServer& cs) :
-    ImportanceDiffusionBase(cs)
+    ImportanceDiffusionBase(cs), _as(opencog::server_atomspace())
```
linas: well, at this point, it doesn't much matter, but this really should be cs.getAtomSpace() so that the cogserver and the agent both use the same atomspace.

linas: (defined in `opencog/cogseerver/server/BaseServer.h`)

opencog/attention/WAImportanceDiffusionAgent.h
```diff
@@ -55,14 +55,18 @@ namespace opencog
...
  * work is explained.
  */
+class AtomSpace;
 class WAImportanceDiffusionAgent : public ImportanceDiffusionBase
 {
 private:
     unsigned int SAMPLE_SIZE = 1;
     unsigned int _tournamentSize;
+    float  _decayRate; // percentage decay
+    AtomSpace*  _as;
```
linas: Anything that inherits from `class Agent` already has an `AtomSpace* _as;` in it see line 126 of `Agent.h`  -- since you've inherited from `ImportanceDiffusionBase` which nherits from `Agent`, you don't need this at all. Its already there.

misgeatgit: Ahh, I'll modify it in my next PR.

PR #2582, opened 2017-01-10
Answer questions like “How many people do you see?”

leungmanhin:


opencog/nlp/chatbot-psi/contexts.scm
```diff
@@ -3,6 +3,9 @@
 (load "states.scm")
 (load "utils.scm")
 
+; For var-decl etc
+(load "../relex2logic/rule-utils.scm")
+
```
AmeBel: I don't think loading from another module like this is a good idea. This kind of addition is what causes the warning `WARNING: (opencog nlp chatbot-psi): `nlp-parse' imported from both (opencog nlp chatbot-eva) and (opencog nlp chatbot)`, in other modules.

leungmanhin: Attempt to deal with that in https://github.com/opencog/opencog/pull/2583, would be nice if @linas can have a look at that PR as well

PR #2583, opened 2017-01-10
Avoid "nlp-parse imported from both (opencog nlp chatbot-eva) and (opencog nlp chatbot)" warning

leungmanhin:


opencog/nlp/relex2logic/rule-utils.scm
```diff
@@ -6,54 +6,54 @@
 ;--------------------------------------------------------------------
 ;
 ; Short-hand for declaring a variable.
-(define (var-decl var type)
+(define-public (var-decl var type)
```
linas: In the long-run, these should be moved out of relex2logic/rule-utils, to some generic nlp/rule-utils file, somewhere.

linas:
I merged, because it looked reasonable. I did not actually test this.

PR #2587, opened 2017-01-12
Fix rest-api test failure

AmeBel:


linas:
Huh. I thought all the code in that directory was completely deead, and we were going to remove basically almost everything there ... is this no longer the plan? Is anything in there worth saving, besides the rest-api?

(you can merge.. I've no clue what this code does, or why you;d want it)

AmeBel:
It is optional, but the restapi uses it. And is tested as part of the restapi unit test suit.

AmeBel:
It generates [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) representation of atoms. Why? I don't know.  Should it be removed? If it doesn't have any use case, sure.

PR #2588, opened 2017-01-12
ECAN

misgeatgit:


linas:
I looked at the code, the code looks OK. I looked at the architecture, I don't understand why you want to do it this way -- its clunky and hard to use.

Storing parameters in a config file, (and then mirroring them in a C++ class) has lots of negatives, and no positives, that I know of:
-- the config file is hard to use and is awkward -- I'd like to get rid of it (and replace it with config values stored in scheme, or better yet, in atoms)
-- storing parameters in C++ classes makes them inaccessible and hard to change, without writing lots of new C++ code.  For example: you have to invent new requests .. 
-- Requests are klunky and awkward to use -- they are not scriptable.  The cogserver shell is NOT bash!  You can't write scripts like `if ($my_parameter>42) then $other_parameter=69`  The only way to use your new `list-ecan-param` is by hand, by typing it in -- you can't put it into any shell script, you can't put it into any perl script, or into python or scheme. it can only be used by hand.

-- which means that there is no practical way to change parameters at runtime. 
-- Also: there's no practical way of editing and reloading the config file. if instead of using the config file, you used python or scheme, then you *could* re-run the python or scheme to update the paramters. But the config file does not allow this.

-- Maybe its OK to store the parameters in a C++ class ... maybe. . But access to data stored in C++ classes is very awkward -- you have to write new Request classes, and/or create scheme bindings with SchemePrimite (which is ugly and painful) and/or create new cython bindings (which is also icky).  But if you store them as atoms, then ZERO new code needs to be written to access them in scheme, python, haskell, ... or C++  Because if you know the name of the atom, you can just ask for it by name, and bingo -- you've got all the parameters.

So, if you really really want to do it this way, you can, -- I won't stop you. But I see only minuses, and no pluses,  ... are there any?

misgeatgit:
I agree with most of what you said. But I did exactly what you said. The code creates StateLinks for storing parameters and their values as StateLinks. The thing which might have confused you are the two ECAN related commands that I created as an extra feature and a shortcut instead of going to guile shell and set the values of the StateLinks in Scheme. Also, The loading from config file is done only once for initialization purpose. So, getting rid of config file means no need to call this initializer function. Which is a matter of commenting out the one line function call.

misgeatgit:
I have removed the hardcoded default parameter values to a scheme file and load the default value from there. The names of the parameters should stay in the code bc that's the only way to get a reference to a particular parameter atom.

misgeatgit:
@linas I have extracted out hard coded initialization to a scheme configuration file. The hard coded names are essential for uniquely identifying each ECAN parameter; so, I have left as it is.

Issue #2584, opened 2017-01-11
Segfault caused while running the pattern miner test 


ruiting:
The pattern miner test is from:  http://wiki.opencog.org/w/Pattern_miner, i.e.

opencog> loadmodule opencog/learning/PatternMiner/libTestPatternMinerAgent.so

============================
Debug: PatternMining start! Max gram = 4, mode = Depth_First
Start thread 0 from 0 to 4294967295
100% completed in Thread 0.
Finished mining 1~4 gram patterns.

processedLinkNum = 0
Debug: PatternMiner:  done frequent pattern mining for 1 to 4gram patterns!
gram = 1: 0 patterns found! 
Debug: PatternMiner: writing  (gram = 1) frequent patterns to file FrequentPatterns_1gram.scm

gram = 2: 0 patterns found! 
Debug: PatternMiner: writing  (gram = 2) frequent patterns to file FrequentPatterns_2gram.scm

gram = 3: 0 patterns found! 
Debug: PatternMiner: writing  (gram = 3) frequent patterns to file FrequentPatterns_3gram.scm

gram = 4: 0 patterns found! 
Debug: PatternMiner: writing  (gram = 4) frequent patterns to file FrequentPatterns_4gram.scm

Calculating interestingness for 2 gram patterns by evaluating surprisingness
100% completed.Debug: PatternMiner:  done (gram = 2) interestingness evaluation!0 patterns found! Outputting to file ... 
Debug: PatternMiner: writing (gram = 2) interesting patterns to file SurprisingnessI_2gram.scm

Debug: PatternMiner: writing (gram = 2) interesting patterns to file SurprisingnessII_2gram.scm

Debug: PatternMiner: writing  (gram = 2) pattern statics to csv file PatternStatics_2gram.csv
[2017-01-11 08:54:56:460] [ERROR] Caught signal 11 (Segmentation fault) on thread 139790070507584
	Stack Trace:
	2: ./opencog/cogserver/server/cogserver (opencog::Logger::Base::~Base() +0x5e) [0x407fbe]
	3: ./opencog/cogserver/server/cogserver (sighand(int) +0x266) [0x4074b6]
	4: /lib/x86_64-linux-gnu/libc.so.6 (sighand(int) +0x36d40) [0x7f23684a0d40]
[...]

PR #2593, opened 2017-01-13
Re-enable pattern miner utest

ngeiswei:
Note that it only tests that the pattern miner didn't crash, not if the result is correct.


ngeiswei:
Related to fixing issue #2584

Issue #2592, opened 2017-01-13
Cogserver skips evaluating some of the messages being sent at a high frequency?


leungmanhin:
For example, at a point the cogserver was receiving:
```scheme
msg1: (StateLink (AnchorNode "Sudden sound change value") (NumberNode 0.0))
msg2: (StateLink (AnchorNode "Decibel value")  (NumberNode 24.7139434814))
msg3: (StateLink (AnchorNode "Sudden sound change value") (NumberNode 0.0))
msg4: (StateLink (AnchorNode "Decibel value")  (NumberNode 30.5852890015))
```

For msg 1 to 3, ConsoleSocket got them, sent them to the shell for evaluation, they were added to the `evalque`, but didn't get the chance to be evaluated (before the shell reaching the dtor?), only msg4 got evaluated in this case, and msg 1 to 3 were really gone. Similar behavior repeated when more messages were received.

It's even worse for "lost face" messages e.g.
```scheme
(cog-execute! (PutLink (DeleteLink (EvaluationLink (Predicate "name") (ListLink (ConceptNode "93943") (VariableNode "reco-id")))) (GetLink (EvaluationLink (Predicate "name") (ListLink (ConceptNode "93943") (VariableNode "reco-id"))))))

(cog-delete   (EvaluationLink (PredicateNode "visible face")     (ListLink (NumberNode "93943"))))

(cog-delete   (ListLink (NumberNode "93943")))

(cog-delete (NumberNode "93943"))
```
as these messages are sent and received almost at the same time, and the shell skips all of them almost all the time (or only evaluates one of them). Because of this, the robot is not really "perceiving" anything at the moment

leungmanhin:
There are still codes that I have not studied or fully understood... @linas any thought on this is appreciated

mjsduncan:
not sure if this related or shuld be it's own issue but cat-ing scheme files through the cogserver port using example code `cat filename | telnet localhost 17001` from [here](http://wiki.opencog.org/w/Scheme#Load_scheme_data_from_a_file) doesn't work with current build 571f05b.

leungmanhin:
`cat filename | telnet localhost 17001` is working for me, maybe there is a missing ')' in the file or something like that? just a wild guess

linas:
[...]

PR #2600, opened 2017-01-18
Not to destroy the shell too early

leungmanhin:
When a message arrives and has been pushed to the `evalque`, the shell may be destroyed immediately even before the expr in the `evalque` has the chance to be evaluated. This causes the robot to "ignore" a lot of the messages it receives

This should fix this particular problem, as described in https://github.com/opencog/opencog/issues/2592 as well


linas:
I doubt that this will work, because the eval thread will wait forever, waiting for more commands to execute.  The cancel is the only thing that breaks it out of that loop. Anyway, pull req #2603 should now solve this same issue.

PR #2601, opened 2017-01-18
Lojban

rTreutlein:
The main addition in this is to allow parsing of placeholder i.e. x1 as an argument to a predicate.

Other than that it contains various improvements and fixed to the parser.


rTreutlein:
While currently the Lojban part isn't built during the test I will never the less merge this now. Shouldn't cause an issue for most people as it isn't build by default. It works on my system so that has to be good enough for now.

PR #2604, opened 2017-01-26
Fix segfault.

misgeatgit:
Checking for ipos value should be done before calling read_synset to prevent a segfault.


linas:
good lord, how'd you even find this old code?

misgeatgit:
@linas I was just trying to generate a wordnet atomspace for our ECAN experiment.

PR #2610, opened 2017-02-07
Fix load path for loading the ECAN default params

leungmanhin:


opencog/attention/AttentionParamQuery.cc
```diff
@@ -104,7 +104,7 @@ HandleSeq AttentionParamQuery::get_params(void)
 void AttentionParamQuery::load_default_values(void)
 {
      SchemeEval scm(_as);
-     scm.eval("(load \"opencog/attention/default-param-values.scm\")");
+     scm.eval("(load \"" PROJECT_BINARY_DIR "/opencog/attention/default-param-values.scm\")");
```
AmeBel: What happens when people just install opencog  using octool?

leungmanhin: using octool should also work (?) Without this it will only search the file in the directory that the cogserver is started

AmeBel: Have you tested it after installing and removing the build directory ?

leungmanhin: Got it, should be fixed in https://github.com/opencog/opencog/pull/2612, thanks!

PR #2612, opened 2017-02-08
Fix load path for loading the ECAN default params again

leungmanhin:


misgeatgit:
IMO this fix appropriate only for release mode. However, for development mode, there should not be a requirement to install before running the system. I'm not completely sure though.

leungmanhin:
I guess it's better to add that to the load path instead, but I'm not sure why `add-to-load-path` doesn't work here?

linas:
Be careful hard-coding this stuff, it makes bugs hard to debug. Relative paths should work, I think, so that a single `add-to-load-path`, set in the main opencog module, should be enough to find all the relative paths.

PR #2607, opened 2017-02-06
Openpsi - variable decay to baseline values 

eddiemonroe:
Internal psi variables will slowly move toward their baseline values absent of influence from external factors.

PR #2616, opened 2017-02-17
Modulator variable decay to baseline values in OpenPsi

eddiemonroe:
Internal psi variables will slowly move toward their baseline values absent of influence from external factors.

This PR replaces PR #2607


ngeiswei:
OpenPsiUTest fails but it also fails on master. I'm merging.

PR #2618, opened 2017-02-22
Fix build break

AmeBel:
The failure to build occur because of https://github.com/opencog/atomspace/pull/1121


AmeBel:
Just pushed fixes due to change in https://github.com/opencog/atomspace/pull/1132

linas:
Thanks for making these changes!

PR #2633, opened 2017-03-15
FindPython only version 2.7

rTreutlein:
same as https://github.com/opencog/atomspace/pull/1169


rTreutlein:
@AmeBel could you take a quick lock at this?

AmeBel:
Thanks

PR #2639, opened 2017-03-21
Chatlang

pennachin:
This has work in progress code skeletons for the chat authoring DSL to Atomese translation. It's not complete nor fully working yet, so don't merge. I'm creating this PR to make it easer for others to collaborate on this code. See the README for a list of tasks (not all of those need to be finished before a first merge, but the trivial ones definitely do).

### DIFF
--- a/opencog/nlp/chatlang/README.md
+++ b/opencog/nlp/chatlang/README.md
@@ -1,0 +1,132 @@
+# ChatScript-like DSL for Chatbot Authoring
+
+This is a work in progress, not yet functional implementation of a DSL designed
+for easy authoring of chatbot rules and supporting behavior and knowledge. The
+DSL is inspired by ChatScript in its syntax, but aiming to use OpenPSI as the
+engine for topic selection and topic management.
+
+## Design Overview
+
+Let's look at an example. The ChatScript rule
+
+```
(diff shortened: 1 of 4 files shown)


linas:
FYI, I think its just fine to merge incomplete, partially working code, as long as it doesn't break existing subsystems.  Its often better to merge early, merge often, as that can help avoid code bases from drifting apart.

In this case, its not urgent, nor a big deal, so we can all wait until later.

linas:
Note also: I plan to move significant parts of the chatbot to the opencog/eva directory, after pull req #2641 is merged.  The point is that significant portions of the chatbot need to be able to talk about what it sees and what it knows, and the current knowledge base is the eva-model directory.

Other chat infrastructure should stay in the nlp directory; I'm hoping that generic nlp stuff stays in the nlp diretory, but that all "grounded knowledge" dealing with perceived physical objects, sounds, events all go in the opencog/eva directory.

I've called it "eva" mostly because it implicitly works with the  Hanson Robotics robots; some of it could be more generic, but its not there, yet.

pennachin:
I guess it makes sense to merge this before #2641 and the reorganization of the chatbot code, as suggested by @linas, so here we go. This folder isn't in the build yet, but should be added as soon as the test script is beefed up a bit and passing.

PR #2641, opened 2017-03-23
Merge the non-ROS robot behavior code into main opencog tree.

linas:
This pull request merges about half of the opencog/ros-behavior-scripting into this repo.

This fixes a circular dependency between opencog and ros-behavior-scripting.  It leves behind all of the ROS nodes, including the sensory inputs and the blender API, in ros-behavior-scripting, thus avoiding a dependency on hansonrobotics/HEAD, where all of these ROS nodes and ROS messages are defined.

This move allows (will allow) a cleaner integration of the NLP & chatbot subsystems with behaviors and self/world-models.

This also is a pre-requisite for chatscript-style authoring of dialog+behavior.


linas:
Not yet tested.  I will test tomorrow, and merge; I don't expect any bugs.

linas:
After this is tested and merged, so should also https://github.com/opencog/ros-behavior-scripting/pull/170

PR #2643, opened 2017-03-24
Remove unused StringManipulator and StringTokenizer includes

ngeiswei:


linas:
this no longer compiles, I get an undefined opencog::toString at LocalSpaceMap2DUtil line 133 and also at AStarTest line 94

PR #2647, opened 2017-03-24
Reivse the SpaceTime server API

linas:
Change the spacetime server API so that it uses Handles instead of ints and floats.


linas:
Merging.  I **think** this works, but haven't tested (can't test, the primary test platform got broke).

PR #2649, opened 2017-03-25
Move the Eva-Chatbot to the embodied-Eva directory 

linas:
This moves the eva-specific embodied chatbot to the eva directory.

It also breaks out the sentiment analysis code into it's own module, instead of having it tangled up in the chatbot.


linas:
Chatbot tested, works after this: the following works, as before:
```
$ cd opencog/eva/chatbot-eva
$ guile -l run-chatbot.scm
scheme@(guile-user)> (grounded-talk "Billy Bob" "Turn right")
Robot looks at point (1.000000 -0.500000 0.000000)
Robot turns head towards (1.000000 -0.500000 0.000000)
```

PR #2653, opened 2017-03-27
BC demos part 4

ngeiswei:
Not finished yet.


linas:
Not sure if it is this pull req or an earlier one, but I now get a build break; cmake croaks: 
```
CMake Error at lib/OpenCogFunctions.cmake:51 (MESSAGE):
  lambda-predicate-construction-rule.scm file does not exist in opencog/pln/rules
```

ngeiswei:
Oh, sure, I'll have that fixed soon.

PR #2665, opened 2017-04-02
Added the first guile pattern miner binding.

shujingke:


opencog/cogserver/scm/config.scm
```diff
@@ -44,6 +44,7 @@
 (add-to-ltdl-path (string-append (getcwd) "/opencog/nlp/sureal"))
 (add-to-ltdl-path (string-append (getcwd) "/opencog/nlp/lg-dict"))
 (add-to-ltdl-path (string-append (getcwd) "/opencog/planning"))
+(add-to-ltdl-path (string-append (getcwd) "/opencog/scm/opencog/learning"))
```
linas: this cannot possibly be right, since there is no ltdl in there.  In fact, this entire section appears to be incorrect, as there are no ltdl's located at any of these paths.  This entire portion of this file needs to be removed.

AmeBel: This adds to LTDL_LIBRARY_PATH when developing without installing. I guess since most of the hr development requires installing this might not be necessary.

linas: yes, except that there is no ltdl at that location!  Do this:
```
ls -la build/opencog/scm/opencog
```
and you will see that there is no such file.  LTDL stands for "libtool dynamic library" and it is a convenience utility for working with dynamically loadable libraries.

All of the pattern miner shard objects are located in `build/opencog/learning/PatternMiner` ant the only one that seems to be an LTDL is the strangely named `liblibguile-patternminer.so` (perhaps one too many libs in there, eh?)

shujingke: Will removed this line... I added it when I was trying to fix the "lib not found" problem, and forgot to removed it when I pushed.

shujingke: Fixed. See #2668.

AmeBel: FYI, https://stackoverflow.com/questions/33061198/what-is-the-difference-betweend-ltdl-library-path-and-ld-library-path

AmeBel: In any case it doesn't matter

PR #2672, opened 2017-04-05
ECAN insect-poison experiment data and scripts

misgeatgit:


linas:
Although this check-in is relatively small, don't we have a special github repo for experimental data?

misgeatgit:
Yes, It is better to keep it outside since it might grow at later stages. I will close this PR then.

PR #2673, opened 2017-04-05
Create the unit test for chatlang

leungmanhin:
Not complete yet, will keep expanding it


pennachin:
@leungmanhin looks like this only works after `sudo make install`, otherwise the chatlang module isn't found. I guess the UTest needs the same ugly load path hackery found on other tests?

leungmanhin:
@pennachin I'll have a look at this, thanks!

leungmanhin:
@pennachin I've added extra load paths to the unit test in https://github.com/opencog/opencog/pull/2679, please let me know if that fixes the issue, thanks

PR #2674, opened 2017-04-05
Hack to get OpenPsiUTest to pass.

pennachin:
@AmeBel does this make sense to you?


linas:
wow. string compares. unexpected.  But, `eval()` returns strings, and only `eval_)tv()` returns TV's.

pennachin:
Got it, #2677 replaces the string compares with TVs.

PR #2679, opened 2017-04-07
Assorted updates to chatlang unit test

leungmanhin:
- Fix paths, error messages etc
- Expand the coverage


pennachin:
@leungmanhin I still have the same issue when running before make install; you should be able to reproduce it using Docker containers so the contents of `/usr/local/share/opencog` and `opencog/build` are pristine. The `(opencog nlp)` and `(opencog eva-behavior)` modules aren't found.

pennachin:
#2681 has a partial fix that handles the nlp module, but still don't understand how to find `(opencog eva-behavior)` without installing.

pennachin:
I think I found the issue, module naming doesn't always follow guile convention, see #2682

Issue #2682, opened 2017-04-07
Guile modules naming convention


pennachin:
While debugging the Chatlang unit test (which didn't work without `sudo make install`, see #2673, #2679, #2681) I ran into this issue. According to this:

https://www.gnu.org/software/guile/manual/html_node/Modules-and-the-File-System.html

a module in path `foo/bar/baz.scm` should be named `(foo bar baz)` so it will be found by guile when searching the load path. E.g., `(opencog nlp relex2logic)` and most of our modules. But a few  don't follow that convention, such as  `(opencog ato pointmem)`, `(opencog eva-behavior)`, and `(opencog eva-model)`. There may be others, I didn't run an exhaustive search; these are the ones I ran into.

The CMake files specify install instructions and paths, so the installed scheme files end up with the proper structure, but that explains why the modules can't be found before installing. Does it make sense to rename them?

AmeBel:
- For non scheme-bingding modules there is a cmake function [ADD_GUILE_MODULE](https://github.com/opencog/opencog/blob/master/lib/OpenCogFunctions.cmake) that creates a symlink in the `/opencog/build/opencog/scm` directory similar to the install path. 

- For scheme-binding modules, as per https://github.com/opencog/opencog/pull/2679#issuecomment-292592067, there are some utilities @ https://github.com/opencog/opencog/blob/master/opencog/cogserver/scm/config.scm that can be used with https://github.com/opencog/opencog/blob/master/lib/development.conf

linas:
The ato pointmem stuff if just broken. all the space-time server junk is broken, and someone needs to go in there and fix it.

The eva-behavior and eva-model install into `/usr/local/share/opencog/scm/opencog` on my machine, which is exactly where they should be.  So there should be no problem there.

linas:
For chatlang,  I would recommend installing it as `(opencog chatlang)`  seems like an OK spot, to me...

linas:
There is no expectation anywhere that anything at all can even remotely work until it is installed. The user MUST install to get a usable system.  In particular, the unit tests are guaranteed to fail if you try to compile or run them before performing an install. You MUST install before compiling the unit tests.

leungmanhin:
I'm not sure what is the best way to fix this kind of issues... I personally prefer to install before testing, and Amen helped to updated the ocpkg https://github.com/opencog/ocpkg/pull/84 accordingly so it should now be passing in docker container as well. But if this is not the best practice maybe we could try come up with some other solutions

[...]

PR #2694, opened 2017-04-19
Update ADD_GUILE_MODULE

AmeBel:
As per https://github.com/opencog/opencog/issues/2682

### DIFF
--- a/opencog/nlp/chatbot/CMakeLists.txt
+++ b/opencog/nlp/chatbot/CMakeLists.txt
@@ -1,4 +1,6 @@
-ADD_GUILE_MODULE (chatbot.scm
+ADD_GUILE_MODULE (FILES
+	chatbot.scm
 	chat-utils.scm
 	bot-api.scm
+	MODULE_DESTINATION "${DATADIR}/scm/opencog/nlp/chatbot"
 )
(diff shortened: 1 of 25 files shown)


opencog/nlp/chatbot/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
-ADD_GUILE_MODULE (chatbot.scm
+ADD_GUILE_MODULE (FILES
+	chatbot.scm
```
linas: Is this correct?  The `chatbot.scm` file normally goes one directory lower, than the rest of the files in the module.  In this example,  `chatbot.scm` would normally go into `${DATADIR}/scm/opencog/nlp/`  where guile finds it, and all the other files go into `${DATADIR}/scm/opencog/nlp//chatbot/`  ... or does guile automagically find the "main" entry point somehow?

AmeBel: You are right. Since `chatbot.scm`  has file name equal to the basename of `MODULE_DESTINATION`, ADD_GUILE_MODULE automagically configures the install directory to be one directory lower. It doesn't prevent developers from using the function similar to the `INSTALL` command.

PR #2729, opened 2017-05-12
Fix

misgeatgit:
Handling AF overflow issue.

### DIFF
--- a/opencog/attention/FocusBoundaryUpdatingAgent.cc
+++ b/opencog/attention/FocusBoundaryUpdatingAgent.cc
@@ -123,6 +123,12 @@ AttentionValue::sti_t FocusBoundaryUpdatingAgent::get_cutoff(HandleSeq& afset)
     auto afSTIValues = std::vector<AttentionValue::sti_t>(af_sti.begin() + minAFSize, 
             ((int)af_sti.size() > maxAFSize ? af_sti.begin() + maxAFSize : af_sti.end()));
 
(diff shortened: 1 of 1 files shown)


opencog/attention/FocusBoundaryUpdatingAgent.cc
```diff
@@ -123,16 +123,21 @@ AttentionValue::sti_t FocusBoundaryUpdatingAgent::get_cutoff(HandleSeq& afset)
...
+    if(afSTIValues[0] == afSTIValues[afSTIValues.size()]){
+        afSTIValues = std::vector<AttentionValue::sti_t>(af_sti.begin(), af_sti.begin() + minAFSize);
+    }
+
+
     int cut_off_index = 0;
-    int biggest_diff = -1 ; // diffs are always +ve. so, its okay to initialize it with -ve number.
-    constexpr int DIFF_MAGNITUDE = 0.5; // make this a parameter
+    AttentionValue::sti_t biggest_diff = -1.0f ; // diffs are always +ve. so, its okay to initialize it with -ve number.
+    constexpr AttentionValue::sti_t DIFF_MAGNITUDE = 0.5f; // make this a parameter
```
linas: why use floats? you'll have fewer bugs, long run, if you just make sti_t be double-precision.  (the 1.0f creates a 32-bit float, so if you store it in a double, it will put garbage in the lower bits of the mantissa)

misgeatgit: I see. I will change it.

PR #2743, opened 2017-05-26
Finish transition to generic pair statistics API

linas:
Just to make clear what this is:

This code allows you to designate any kind of set of atoms in the atomspace to be a correlation matrix, and then to perform some fairly general statistical analysis on this matrix.  The matrix is some high-dimensional, MxN sparse matrix.  The row atoms can be different than the column atoms.  One writes some smallish wrapper class that shows how to get the j'th row or the k'th column, and how to get a numeric value for the (j,k)'th entry.  Because every entry in the matrix is represented by a pair of atoms (j,k), these are called "pairs".

Statistics include: rms length of rows, columns, lp-norm of rows, columns, entropy and mutual information of rows, columns, cosine distance and jaccard distance between any two rows (or any two columns), and a generic 'fold' that can apply any function to any p rows or columns.  Not implemented: PCA or factor analysis of the matrix.

Due to historical reasons, the code is in scheme. Perhaps someone might want to re-write it in C++.  Perhaps someone might want to figure out how to attach it to some generic stats library, e.g. scipy.  Sparse, high-dimensional PCA or Lanczos is tedious to implement, so doing it in scheme does not seem wise.  Perhaps there is a way of gluing this to tensorflow. ?


linas:
The correlation-matrix code was copied to the atomspace, here: https://github.com/opencog/atomspace/pull/1241  It can be thought of as providing a generic matrix-shaped "window" into the atomspace.

PR #2758, opened 2017-06-07
Sumo test3

ngeiswei:
To run that well the backward chainer needs meta-rule full support, I need bite the bullet and implement it.


ngeiswei:
@AmeBel beware that PLN rule paths have been changed.

PR #2768, opened 2017-06-12
Remove unused(?) un-needed(!) cogserver modules.

linas:
The strategic goal here is to eventually retire and remove the cogserver, and all the code that drags along with it: the shells, the module system.  But before this can happen, all the various accumulated modules have to be removed first. I believe that the following are unused by anyone in any work-flow:
```
benchmark
persist
persistzmq
viterbi
wsd
dimensional-embedding
```
This pull request removes the first four.  I'll let this pull req hang out here, for a while, in case anyone has any objections.  But mostly I'm looking at this as an opportunity to clean up and remove old dead code. 

The code for the remaining two is maybe-ish OK, and should not be deleted; only the cogserver-module setup should be deleted.


ngeiswei:
I have no idea what should be done about `dimensional-embedding`, maybe @bgoertzel knows.

bgoertzel:
Well the dimensional-embedding code is bit-rotted, but the core algorithm in there makes sense, and it used to work interestingly for dimensional embedding of Atomspaces containing NLP knowledge....   It is probably not usable in its current form but that's just because other stuff was changed and it was not updated....   We could remove it from the codebase but tag it so that someone can find it later if they want to resurrect and update it.   

Some form of dimensional embedding is obviously useful for Opencog ... this algorithm is fundamentally more suited for OpenCog in general than PCA for example (because it works naturally on graphs) ...

There's no reason dimensional embedding needs to be in a cogserver module of course, that's just a somewhat arbitrary way to run the algorithm...

linas:
The goal here s not to remove the code; the goal here is to remove the opencog module subsystem.  The proposal is not to remove the dimensional embedding code; the proposal is to remove the  fact that it uses the module subsystem.

PR #2779, opened 2017-06-16
OpenPsi updates

eddiemonroe:


linas:
Eddie, please (as discussed before) separate the code for human-psychological state modelling from the rest of the OpenPsi rulebase.    About 90% of the code in the openpsi directory is about generic rule selection from batches of generic rules.  I'd like to get that up to 100%.

The parts dealing with human psychology, e.g. emotions, really need to go into a distinct directory, devoted to ... human psychology. I suggest something in the `eva` subdirectory.  Things like `emotion` could then use the generic open-psi rule-selection subsystem to pick whatever human emotions you might want to model.   The goal here is to avoid spaghetti code, stove-pipe design.

Another remark: from the point of view of the robot, modelling affects is gonna give a much bigger bang for the buck than modelling emotions.  The affects have a much closer correspondence with facial expressions and behaviors than emotions do.  For example, you may feel sad, but will still smile at your friends.  That's because your affective state is to be pleasant, despite your unhappy emotional state.  We are going to have trouble getting this to work right, look good, without these changes.  See for example https://github.com/opencog/opencog/blob/master/opencog/eva/README-affects.md

eddiemonroe:
Linas, OpenPsi has two high level parts, a) motivation & action select, and b) cognitive modulation & emotion. Perhaps it makes sense to have two high level directories separating these versus all together? Maybe we should just be naming these by their functionality rather than using the label OpenPsi for all of them? I don't know. But for now at least, OpenPsi is being used by the group as a label for all of this.
 
Emotion and affect are concepts from psychology/mind that could be applicable to a wide range of agents (not just for modeling human psych). The generic emotion code in OpenPsi is just for creating an emotion and listing all the emotions created. The particulars of how the emotion is named and defined and used is to be specified for the individual agent and is not part of the openpsi module.

I don't think it makes sense to have this generic openpsi emotion code in the eva subdirectory, which seems to be for HR in particular.

Regarding what you wrote about affects versus emotions, there isn't general consensus about the particular definition of these, with different researchers and models defining them in different ways. It doesn't matter so much to me which labels are used for what, as long is there is agreement within the group over the definitions we are using. 

From your example with the smile, it seems to me you are pointing to the distinction between affect and affect display, which I agree is a useful distinction!

eddiemonroe:
Linas, you wrote: " About 90% of the code in the openpsi directory is about generic rule selection from batches of generic rules. I'd like to get that up to 100%."

I don't think of the Psi rules as being generic, but rather of a specific form, context + actions implies goal. And the action selector does not seem to me generic rule selection either. But maybe you are meaning generic at the level that the particular rules are not specified.

bgoertzel:
On Wed, Jun 21, 2017 at 5:56 AM, Eddie Monroe <notifications@github.com>
wrote:

> Linas, OpenPsi has two high level parts, a) motivation & action select,
> and b) cognitive modulation & emotion. Perhaps it makes sense to have two
> high level directories separating these versus all together? Maybe we
> should just be naming these by their functionality rather than using the
> label OpenPsi for all of them? I don't know. But for now at least, OpenPsi
> is being used by the group as a label for all of this.



I am not attached to the name "OpenPsi" at all, btw ... but nor am I
especially eager to get rid of it... whatever...

I agree that splitting up

-- motivated action selection

-- cognitive modulation based on indicator predicates ("emotions" being a
convenient label for these indicator predicates)

into separate subfolders might be cleaner conceptually (though there would
not be a lot of code in either folder at this stage)...

We could do

openpsi/action-selection

openpsi/modulation

...

I also agree with the current design choice to keep definitions of specific
emotions (happy, angry, whatever) separate from the above two subfolders...

And I agree w/ Eddie that putting the openpsi/modulation stuff in the
character-specific "Eva" folder would be more confusing than clarifying...

ben








-- 
Ben Goertzel, PhD
http://goertzel.org

"I am God! I am nothing, I'm play, I am freedom, I am life. I am the
boundary, I am the peak." -- Alexander Scriabin

linas:
Hi Eddie: I see two different things at work in that code base.  These are:

1) Achieving strategic goals by taking certain tactical actions, modulated by time-varying demands. The tactical actions are carried out by different sets of rules run by the rule engine, and OpenPsi picks out the best ruleset for the current situation.  

For example, a fork-lift robot might have the strategic goals of keeping the oldest boxes on the north end of the warehouse, while also minimizing distance traveled. Time-varying "demands" might be the arrival of shipments at the loading dock, and new dispatch orders. Tactical actions might include temporary re-stacking of boxes to get them out of the way.  All of this involves modulation and demands and actions and goals, but has nothing to do with human emotion.  Its a generic rule-selection system that folds together strategy, planning, demands, which modulate action.

2) OpenPsi provides a model of human emotion. 

I'm very interested in 1) because we need that kind of a generic system. I'm not interested in 2) because I think that, as a model of emotions, its simplistic and inaccurate, and I think that, for the robot a model of affects and behaviors will be far superior to a model of emotional states. 

I tried to write up a sketch for an affective model that is psychologically more correct than what OpenPsi provides.  Have you read it? It's in the README-affects file.

yantrabuddhi:
Hi Linas,
 can you point me to the README-affects file, i'd like to read it.
Regards,
Mandeep

linas:
https://github.com/opencog/opencog/blob/master/opencog/eva/README-affects.md

linas:
Hi Mandeep, and also Eddie: I think a key point comes very early in the Ostrovsky PDF: "Affect is biology, emotion is biography."  For me, this says that we should be modelling something that is biologically correct.  The other bullets are:

* "A feeling is one or more affects that attain consciousness."
* "An emotion is a bundle of feelings, given a (cultural) label."
* "Affects are transient, feelings may linger."
* "A mood is a set of chronically triggered feelings." (mood is to feeling what climate is to weather)
* "A drive is body-specific need (e.g.) oxygen, water, food, sex."
* "Personality is a collection of behavior-scripts in response to affects."

So a couple bullets down, I hope its clear why modelling "emotion" is fundamentally incorrect: it's a bundle of affects, and different cultures don't even agree on the bundles. This pre-occupation with happy/sad  seems (to me) to be an anglo-saxon thing,  although now that I think about it I know someone who's half welsh and half british, and its never happy/sad for them its always passive-aggressive+low-self-esteem.  That is, "emotions" are not primal, and what the body/mind experiences does not always align well with the elementary-school labels of happy/sad.  There are better models, better theories out there.

yantrabuddhi:
Hi Linas,
 I think I do agree to some degree. I have never given it in depth thought
before. I think of emotion as a complex cognitive reaction to state of
interaction with environment and health based on past experiences and
learning. Of-course by that definition a child would have to learn
emotions. I want to think more about it to have an opinion, so I want to
read about it.
Thanks,
Mandeep

mjsduncan:
i've been meaning to respond to this thread for awhile since i spent a month researching emotion for hanson robotics and concluded that the component process model (cpm) is the most biologically plausible theory that is compatible with and builds on the openpsi framework.  i'm not up on linus' and eddie's code so i don't know how relevant openpsi/cpm theory is to the current direction of opencog emotion modeling but that said, i think the nathanson/ostrfsky framework is generally out-dated.  [*this link*](http://www.iep.utm.edu/emotion/) gives a good run down of the breadth of current theories on emotion.  nathonson isn't mentioned but i would categorize his work under "somatic feedback theories".  scherer's component process model is cited as a "cognitive appraisal theory" though i think it has elements described under "non-cognitive" and "somatic feedback" theories in the review.  [*this interview*](http://emotionresearcher.com/the-component-process-model-of-emotion-and-the-power-of-coincidences/) is a good introduction to cpm in scherer's own words (skip to the 5th bolded question).   [*this zip file*](https://drive.google.com/file/d/0By7BDnSNB6uSa0UzYmx4NGU3UDQ/view?usp=sharing) is a minimally organized collection of pdfs of current research in affective neurology and most of the literature on the cpm.

i won't go into detail now because i'm supposed to be doing other stuff but this is an important discussion to have.  when i worked on this 6 months ago the infrastructure wasn't developed enough for it really to be applicable and now the work is being done to integrate the sensory input with goal formulation and behavior planning and implementation.  maybe this conversation should be moved to the google group?

PR #2789, opened 2017-06-20
bring over changes for psi-dynamics.scm from ros-behav repo

eddiemonroe:
per discussion in https://github.com/opencog/ros-behavior-scripting/pull/173#event-1127499130


linas:
sorry for long delay swamped in email

PR #2802, opened 2017-06-27
Use setTruthValue not merge

linas:
I think the intent here is to set, and not to merge.
Am trying to remove the troublesome merge() method.


ngeiswei:
I'm almost sure it is to set, I'm merging.

ngeiswei:
@linas why do you want to remove merge? Some PLN rules use merge, a certain conservative form of it, it simple enough that it could be moved to scheme but then other language bindings won't be able to re-use it.

linas:
There are merge functions on the TV itself; I am NOT removing those. If those are used from within C++, there is no loss of performance by using those, directly.  It just moves code around.

I want to simplify the Atom a bit; this is part of solving a bottleneck elsewhere.

linas:
For backwards-compat, I am adding this:
```
(define-public (cog-merge-tv! ATOM TV)
" cog-merge-tv! -- merge truth values on atom"
       (cog-set-tv! ATOM (cog-tv-merge (cog-tv ATOM) TV)))
```

linas:
See opencog/atomspace#1291 for the actual change

ngeiswei:
Ah, fine.

PR #2834, opened 2017-07-24
Separate out (opencog openpsi dynamics) from (opencog openpsi)

AmeBel:


linas:
Thank you. This seems like the right thing to do.

PR #2847, opened 2017-08-09
cmake fix 

AmeBel:


Dagiopia:
This change should be to the `/opencog/pln/rules/propositional/CMakeLists.txt` file.

PR #2857, opened 2017-08-18
Remove GenericTruthValue usage

AmeBel:
See https://github.com/opencog/atomspace/pull/1340


linas:
thanks, I missed that somehow

PR #2863, opened 2017-08-25
Btree-psi Updates

AmeBel:


opencog/openpsi/main.scm
```diff
@@ -141,9 +150,17 @@
...
 
-    (cog-logger-debug opl "Ending psi-step, loop-count = ~a"
-        (psi-get-loop-count))
+    ; Do garbage collection. This is a replacement to (run-behavior-tree-gc)
+    (when (equal? 0 (modulo psi-loop-count 1000))
+        (cog-map-type (lambda (a) (cog-delete a) #f) 'SetLink)
+        (cog-map-type (lambda (a) (cog-delete a) #f) 'ListLink)
+        (cog-map-type (lambda (a) (cog-delete a) #f) 'NumberNode)
+        (cog-map-type (lambda (a) (cog-delete a) #f) 'ConceptNode)
+        (cog-logger-debug opl
```
linas: Could you change `cog-delete` to `cog-extract` ? Right now, this make no difference, but it could avoid tragic data loss, if someone had a database open.  Probably should rename `cog-delete` to something else...

PR #2871, opened 2017-09-05
Ure rule definition updates

AmeBel:
Continuation of https://github.com/opencog/opencog/pull/2868


ngeiswei:
@AmeBel please note that the TV is optional, since they seem all equal, that is you do not prioritize a rule over another, you might as well not specify it.

PR #2890, opened 2017-09-19
Feature pattern index

ngeiswei:
## Overview

Atomspace index builder prototype by Andre Senna + example of pattern
matcher and miner algorithms taking advantage of that index.

Also offers a scm load function with smaller RAM footprint.

I have improved the code in shallow ways, mostly by using more modern
C++.

## Description

The main algorithm

1. Loads an scm file by splitting into chucks of root atoms (atoms
   with no incoming), thus using a smaller RAM footprint than using
   scheme load (roughly divided by the number of chunks).

2. As it loads the scm file, it indexes the root atoms.

3. After loading, based on the index, it associates patterns and their
   occurances.

Once built, the code offers rudimentary pattern matcher and miner
functions with possibly better performances than existing algorithms.

Also, during this processing, hypergraphs are flattened into vectors

Similarity
  And
    Concept "A"
    Concept "B"
  Concept "C"

is turned into

{(SimilarityLink 2),
 (AndLink 2),
 (ConceptNode 0) "A",
 (ConceptNode 0) "B",
 (ConceptNode 0) "C"}

where the outgoing arity allows not to loose information about their
structure.

The Pattern matching algorithm seems to have a different strategy than
Linas pattern matcher. It processes recursively the query by dividing
it into sub-queries. I don't know how pattern mining works.

# Conclusion and further work

I think the idea of pre-building indexes + occurrences to subsequently
speeding up matching and mining has merit. I'm not sure about turning
hypergraphs into vectors. I vaguely recall that the pattern miner
operates such a similar transformation but it feels convoluted.
Wouldn't it better if the algorithms worked directly on hypergraphs?

Anyway, it seems we want either to

1. improve this code and use it as it is

2. generalize it so that Linas and Shujing algorithms can use it.

I think it would be worth having @linas and @shujingke take a look at it. But
before that, when/if possible, it would be great if @andre-senna  can add
more comments, especially to important methods such as
`TypeFrameIndex::computeSubPatterns`, like a brief description, their
inputs and outputs, etc.


linas:
I skimmed this, but am confused. Can a simpler explanation be provided? 

* The commit message mentions scheme, but what does this have to do with scheme? 
* It also mentions "file" but why should it matter if its a file or not? What if the data is not scheme, but is being generated on the fly? 
* What do you mean by a "smaller RAM footprint"? Smaller than what?
* It says that it "indexes the root atoms", but that is also what the atomspace does.  Is this a replacement for the atomspace?  
* The README says: "Build an index with only the relevant atoms" -- what is a "relevant" atom?
* The README says: "Use this index to gather information from the AtomSpace as required by the algorithm" - what is the "information"? how does one "gather" it? what is the "algorithm"? What is it "requiring"?  How does an algorithm require something? What is being indexed?
* The README seems to suggest that this is a replacement for the pattern miner, but the example looks like a pattern-matcher example, not a pattern-miner example.  

Overall: What does this system do, and when would one use it? What problem does it solve?

linas:
I looked at this more carefully, and from what I can tell, it is a prototype proposal for a redesign of the atomspace, from the very foundations. It fundamentally alters:

* how atoms are read in (using a scheme-like format, but this could also be changed to be json-like) 
* how atoms are stored, (using a flattened vector format)
* how the pattern matcher & miner work (by using indexed mini-patterns that can be assembled)

In principle, this kind of exploration is not a bad idea -- its worthwhile to explore, to try to discover something different, more scalable than the current code base.  The current code-base is frustratingly bloated and slow -- each atom takes about 1.5KB RAM, and can be manipulated at the pathetic rate of about 3-50K atoms/sec. Surely it must be possible to improve on this.

On the other hand, the atomspace is the way it is partly due to history and partly due to user-demand.  Its bloated and slow because people want to do certain things this prototype does not handle; the biggest one being fast access to the incoming set of an atom.  If you  can find some aspect of the current design that no one uses, I would happily remove that.

linas:
On the other hand, if you insist that this is only about user-defined indexes, and nothing else ... well, there are certainly other, simpler, less radical designs that fit better into the existing architecture.  These are discussed here: http://wiki.opencog.org/w/User-defined_index  However, so far, no one has asked for these.

If you want to say that hey, these indexes can be used as mini sub-patterns out of which bigger search patterns can be constrcuted, then yes, this is true.  However, please note that the PatternLink was designed in such a way that patterns can be compiled, before execution.  Some compilation is already done, and more could also be done.  The PatternLink is a natural place to store that kind of "assembly code" for patterns.

pennachin:
This is not a prototype for a redesign of the Atomspace. This is also not a replacement for the pattern miner. This is meant as a faster way to query largely static Atomspaces by creating application-specific indexes.

bgoertzel:
I would add that Senna did not create this "just for fun", he created it out of his experience using Atomspace and Pattern Matcher in the BuildDirect customer-support-chatbot project, which involved real-time user queries against a large Atomspace that was mostly fixed for long periods of time....

ngeiswei:
* The commit message mentions scheme, but what does this have to do with scheme?

Because currently only fetching data from scheme files is supported.

* It also mentions "file" but why should it matter if its a file or not? What if the data is not scheme, but is being generated on the fly?

The idea is not tied to scheme, you just need to provide a TypeFrameIndex builder, the only existing one is

https://github.com/opencog/opencog/blob/master/opencog/learning/pattern-index/TypeFrameIndexBuilder.h#L33

* What do you mean by a "smaller RAM footprint"? Smaller than what?

Smaller w.r.t. to guile, because it doesn't need to load the whole scheme only chunks (root atoms). 

* It says that it "indexes the root atoms", but that is also what the atomspace does.

I didn't know the AtomSpace was indexing the root atoms, can you indicate the code that is doing that?

andre-senna:
Hi Linas.

This is unfinished prototype code. As Cassio said it is not meant to replace any Opencog subsystem. It have been conceived to allow pattern matching (with reasonable time cost) in a huge hypergraph of opencog atoms without necessarily loading than into an Atomspace (the internal structure I use to represent atoms can be extended to keep atoms in disc rather than RAM).

In the Opencog-based application we have been building for a client, we had two sets of atoms.

(a) the result of NLP-parsing lots of sentences from their blog/faq/chat pages
(b) the "atomized" representation of their products database 

I will not enter in details but (b) is a semi-static set of atoms (changed regularly each N weeks). This applications were supposed to make product suggestions and to answer technical questions related to specific products. To do this, we used to do queries in (a) or (b) (usually disjoints queries, ie either  we query (a) XOR (b)). The problem is: (b) is HUGE and loading it into RAM is impossible.

We've tried to filter the products database and load only the most important data. It kind of worked but all the queries become slow. So I though we could use some sort of indexing mechanism to prevent us from reading (b) into RAM but allowing us to perform (at a reasonable time cost) pattern matching queries.

In this first implementation (RAM index) stage my goal is to equal performance of Opencog's Pattern Matcher. This is a hard goal and I could not achieve it properly.

I've decided to implement a pattern miner as well to have a concrete example of something that uses the index-based pattern matcher. Actually I regret this decision badly because pattern mining is (very) hard to implement and I missed the opportunity to finished the "in-disc" version of the index. But results are also promising  so I think we might be able to run a pattern miner over a huge atom set without loading it into an Atomspace.

andre-senna:
Nil,

Actually the original api is C++. So you can use it without scheme. BTW I've just added the scheme API because I thought it might be useful to ease testing.

I'll improve documentation and code comments as you requested.

linas:
>
>    - It says that it "indexes the root atoms", but that is also what the
>    atomspace does.
>
> I didn't know the AtomSpace was indexing the root atoms, can you indicate
> the code that is doing that?
>

!!! ??? This is the one and only thing that the atomspace does. The
atomspace serves no other purpose in life, than to do this.

Maybe the problem is one of terminology. Maybe I should say "the atomtable
does this; the atomtable serves no other purpose than to provide an
index."   The code in AtomSpace.cc is just a thin thin wrapper around the
atomtable, and the only thing that thin wrapper does is access the sql
backend.  The atomtable itself is quite small too -- its about 3KLOC total,
which makes it half the total size of this commit.

linas:
Hi @andre-senna OK.  I too now have databases that are too big to fit into RAM. Currently, I just load slices (from postgres), do some computation, and delete that slice. Its really very slow. 

It would be great if you made your pattern-matcher be more compatible with the kind of output that BindLink does. -- that is, BindLink can be used to discover your variable bindings; you do not need to invent a brand-new format for this.   Note: I am NOT saying you should use BindLink; just that you should make the API and the output be compatible with it.

I don't know if you followed any of the "sheaf" conversations; I'd like to repeat them here, they might be useful for you. So -- the textbook-canonical way to describe a graph is to list all of its edges and vertexes. In opencog, an edge is often written as `(EvaluationLink (PredicateNode "edge-label") (ListLink (ConceptNode "one vertex") (ConceptNode "another vertex")))`  You are probably using something like this in your stuff.

There's another way to do this (and for me, this other way is turning out to be much much better):  I am doing this:
```
(Section (Concept "a vertex") 
    (SetLink 
        (ConnectorLink (Concept "vertex two") (Predicate "some-edge-label")
        (ConnectorLink (Concept "vertex three") (Predicate "another-edge-label")
        ... and so on ...
     ))
```
The above specifies a vertex, and all the edges attached to it. In my case, I use a slightly different format: I have only two edge labels, + and -. I uses an ordered link instead of a set link... etc.  

abstractly, the above represents the same knowledge as the EvaluationLink+PredicateNode idea, but for my algos its just easier to think about and work with, since I get all the local information about the graph in one place.

Basically, I get all the "nearest neighbors" of a vertex in one place -- all the other vertexes that are exactly one hop away.

linas:
p.s. @andre-senna  it looks like you have not set your email adsress in git, and so when Nil commits your work, it looks like he wrote the code, not you.  You do not show up on the contributors page.   I'm not sure, but I think if you just set your email, that will fix it.

ngeiswei:
On 09/21/2017 04:43 AM, Linas Vepštas wrote:
>  >
>  > - It says that it "indexes the root atoms", but that is also what the
>  > atomspace does.
>  >
>  > I didn't know the AtomSpace was indexing the root atoms, can you indicate
>  > the code that is doing that?
>  >
> 
> !!! ??? This is the one and only thing that the atomspace does. The
> atomspace serves no other purpose in life, than to do this.

Well, I know the classes TypeIndex and FixedIntegerIndex but they don't 
index just root atoms, they index all atoms.

ngeiswei:
On 09/21/2017 05:09 AM, Linas Vepštas wrote:
> p.s. @andre-senna <https://github.com/andre-senna> it looks like you 
> have not set your email adsress in git, and so when Nil commits your 
> work, it looks like he wrote the code, not you. You do not show up on 
> the contributors page. I'm not sure, but I think if you just set your 
> email, that will fix it.

This might be cause I rebased.

PR #2924, opened 2017-10-23
More pattern miner fix

ngeiswei:
+ Add functions to print `HTree`
+ Move parameters to a `Parameter` class
+ Simplify functions like `PatternMiner::generateALinkByChosenVariables` (renamed into `PatternMiner::substitute`), etc.
+ More consitification
+ More tests

PR #2926, opened 2017-10-23
Fix pattern miner

ngeiswei:
Should address https://github.com/opencog/opencog/pull/2924#issuecomment-338593618 and  https://github.com/opencog/cogutil/pull/93#issuecomment-338587410


AmeBel:
Thanks

PR #2934, opened 2017-10-27
Fixing OSX build error

issacnitin:


opencog/nlp/irc/IRC.cc
```diff
@@ -149,7 +149,7 @@ int IRC::start(const char* server, int port, const char* nick,
 
 	/* Ping every .. I dunno -- 5 minutes? */
 	optval = 300;
-	rc=setsockopt(irc_socket, IPPROTO_TCP, TCP_KEEPIDLE, &optval, optlen);
+	rc=setsockopt(irc_socket, IPPROTO_TCP, TCP_KEEPCNT, &optval, optlen);
 	if (0 > rc)
```
linas: Hmm. Actually, both the original, and your change are wrong; it should have been `TCP_KEEPINTVL` -- does MacOS support that?

linas:
Anyway, I just now changed it to `TCP_KEEPINTVL`

opencog/nlp/irc/IRC.cc
issacnitin: Yes it does, thanks.

ngeiswei:
Can we close this PR?

ngeiswei:
I assume we can.

PR #2949, opened 2017-11-27
modify python rest code apimain.py and add apighost.py for mozi ghost interface

mjsduncan:
code by xabush_semrie and aman

### DIFF
--- a/opencog/python/web/api/apighost.py
+++ b/opencog/python/web/api/apighost.py
@@ -1,0 +1,38 @@
+__author__ = 'Cosmo Harrigan'
+
+from flask.ext.restful import Resource, reqparse
+from opencog.scheme_wrapper import scheme_eval
+from flask import abort, jsonify
+
+
+
+class GhostApi(Resource):
+
+    @classmethod
+    def new(cls, atomspace):
+        cls.atomspace = atomspace
+        return cls
+
+    def __init__(self):
+        self.reqparse = reqparse.RequestParser()
+        self.reqparse.add_argument('query', type=str, location='args')
+        super(GhostApi, self).__init__()
+
+
+    def post(self):
+        data = reqparse.request.get_json()
+        print(data)
+        if 'query' in data:
+            query = data['query']
+            print(query)
+            action = "(test-ghost \"{}\")".format(query)
+            scheme_eval(self.atomspace, action)
+            response = scheme_eval(self.atomspace, '(get-result)')
+            print(response)
+        else:
+            abort(400,
+                  'Invalid request: required parameter command is missing')
+
+        return jsonify({'response': response})
+
+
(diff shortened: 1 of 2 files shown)


opencog/python/web/api/apighost.py
```diff
@@ -0,0 +1,38 @@
+__author__ = 'Cosmo Harrigan'
+
```
linas: `__author__ = 'Cosmo Harrigan'`  ???  I don't think so.

opencog/python/web/api/apimain.py
```diff
@@ -61,6 +65,16 @@ def __init__(self, atomspace):
...
                               '/api/v1.1/scheme',
                               endpoint='scheme')
+        self.api.add_resource(ghost_api, 
+                              '/api/v1.1/ghost', 
+                              endpoint='ghost')
+
+
+    def allow_cross_domain(self, response):
+        """Hook to set up response headers."""
+        response.headers['Access-Control-Allow-Origin'] = '*'
```
linas: I'm unclear as to what this is about ... cross-domain scripting is considered to be one of the great evils of the last decade, its that infamous technique that allows systems to be compromised, hijacked. 

Now, I don't really understand quite what this is going to do, so perhaps its totally harmless. Still, turning off all access control is traditionally one of those dangerous things to do... can this be double-checked?  What happens if these lines of code are removed?

opencog/python/web/api/apighost.py
linas: just change this to   `__author__ = 'xabush_semrie and aman'`

linas:
except for the alarming removal of all access control, this seems fine.  Please double check if the access control stuff really is needed...

mjsduncan:
this code was added for the mozi product ghost chat interface, which incorporates the relex server into the REST interface.  i'll notify the icog authors so they can answer your questions...

linas:
relex!?? surely ghost is not using relex, is it? Maybe some other subsystem is using R2L ?

Anyway, if this is blocking progress, then merge; I kind-of-ish don't care; I'm just trying to make sure that we're following healthy coding styles, and in particular, giving security at least some afterthought.

opencog/python/web/api/apimain.py
amanfocuslabs: This is a hack that I created because the pre-flight request was pending when I tried to connect to the scheme endpoint. It might show up on your local machine if you remove this and try to send queries to the scheme end-point. But we don't need this for the ghost end-point so you can go head and remove it. And mike took the code locally from us when he was in addis that is why such hacks where in the pull-request. Remove the whole funciton definition and also it usuage at line 90.

mjsduncan:
thanks, linus!

leungmanhin:
ah yes, inputs to ghost will be parsed by `nlp-parse`, so relex is needed

PR #2951, opened 2017-12-06
fix failing MOSESPLNSynergyUTest

Yidnekachew:


linas:
I'm merging, but its clear that loading from paths like that is a really bad idea.

PR #2962, opened 2017-12-28
Remove AtomSpace pointer from LGParseLink::execute()

leungmanhin:
as per the changes made in https://github.com/opencog/atomspace/pull/1486

### DIFF
--- a/opencog/nlp/lg-parse/LGParseLink.cc
+++ b/opencog/nlp/lg-parse/LGParseLink.cc
@@ -136,17 +136,14 @@ LGParseMinimal::LGParseMinimal(const Link& l)
 
 // =================================================================
 
-Handle LGParseLink::execute(AtomSpace* as) const
+Handle LGParseLink::execute() const
 {
 	if (PHRASE_NODE != _outgoing[0]->get_type()) return Handle();
 	if (LG_DICT_NODE != _outgoing[1]->get_type()) return Handle();
 	if (3 == _outgoing.size() and
 	   NUMBER_NODE != _outgoing[2]->get_type()) return Handle();
 
-	if (nullptr == as) as = getAtomSpace();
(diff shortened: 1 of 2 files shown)


opencog/nlp/lg-parse/LGParseLink.cc
```diff
@@ -136,17 +136,14 @@ LGParseMinimal::LGParseMinimal(const Link& l)
...
 	if (PHRASE_NODE != _outgoing[0]->get_type()) return Handle();
 	if (LG_DICT_NODE != _outgoing[1]->get_type()) return Handle();
 	if (3 == _outgoing.size() and
 	   NUMBER_NODE != _outgoing[2]->get_type()) return Handle();
 
-	if (nullptr == as) as = getAtomSpace();
-	if (nullptr == as)
-		throw InvalidParamException(TRACE_INFO,
-			"LgParseLink requires an atomspace to parse");
+	AtomSpace* as = getAtomSpace();
```
leungmanhin: Actually it doesn't seem to get the atomspace here, I guess it's fine to get the atomspace from one of the  `_outgoing` atoms though that requires them to be in the atomspace in the first place, or should it be handled in the `Instantiator` instead?

linas: Yeah. Getting rid of the atomspace in the execute() method seemed like a good idea, based on other parts of the code. Not so much, here, where its central.

I'm looking now.

linas:
I just merged #2963 which is nearly identical to what you propose. Maybe it will work; I haven't tested.

leungmanhin:
It works, thanks!

leungmanhin:
I got another maybe-related problem: https://github.com/opencog/opencog/issues/2964

PR #2963, opened 2018-01-03
Modify signature of the execute method to match atomspace.

linas:
This was changed in https://github.com/opencog/atomspace/pull/1486


linas:
Alternative to pull req #2962

PR #2977, opened 2018-01-17
Use 'LEFT-WALL' consistently

leungmanhin:
`LEFT-WALL` is used when doing random-parse so all the MI calculation is based on `LEFT-WALL` (with another word), but `###LEFT-WALL###` is used during mst-parse, causing no `LEFT-WALL`-* link get generated.

@linas should I change it to `LEFT-WALL` just to make it consistent, or is there any reason it should be kept this way?

### DIFF
--- a/opencog/nlp/learn/scm/mst-parser.scm
+++ b/opencog/nlp/learn/scm/mst-parser.scm
@@ -170,7 +170,7 @@
 
 	; The left-wall indicates the start of the sentence, and
 	; is used to link to the head-verb, head-noun of the sentence.
-	(define left-wall "###LEFT-WALL###")
+	(define left-wall "LEFT-WALL")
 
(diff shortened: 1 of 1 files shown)


linas:
Whoops, right bug, wrong fix.  `###LEFT-WALL###` is correct. It used to be generated by relex.  It should still be getting used in the newer code.   It needs to be with the hash-marks, so that its clearly "not a word". An alternative would be to use `(AnchorNode "LEFT-WALL")` to prove that its not a word, but that introduces other issues.

linas:
Pull req #2978 does "the right thing", I think.

linas:
FWIW, the hash-mark version shows up in relex2logic, also, and in other places as well. Since LgParse is brand-spanking new, I figure its better to make it compatible with the old usage.

leungmanhin:
Yes thanks! #2978 works, I'm closing this one

PR #2978, opened 2018-01-18
RelEx compatibility fix, see bug #2977

linas:
Relex compatibility fixes.

### DIFF
--- a/opencog/nlp/lg-parse/LGParseLink.cc
+++ b/opencog/nlp/lg-parse/LGParseLink.cc
@@ -275,6 +275,12 @@ Handle LGParseLink::cvt_linkage(Linkage lkg, int i, const char* idstr,
 			wrd = strndupa(phrstr + sb, eb-sb);
 		}
 
+		// LEFT-WALL is not an ordinary word. Its special. Make it
+		// extra-special by adding "illegal" punctuation to it.
+		// FYI, this is compatible with Relex, relex2logic.
+		if (0 == w and 0 == strcmp(wrd, "LEFT-WALL")) wrd = "###LEFT-WALL###";
+		if (nwords-1 == w and 0 == strcmp(wrd, "RIGHT-WALL")) wrd = "###RIGHT-WALL###";
+
 		char buff[801] = "";
 		strncat(buff, wrd, 800);
 		strncat(buff, "@", 800);


akolonin:
Linas, in en_pairs_cfive_mst, there is no any LEFT-WALL in parses. Is it a bug or there is another reason for this? 
en_pairs_cfive_mst=# select * from atoms where name like '%LEFT-WALL%';
 uuid | space | type | height | name | outgoing 
------+-------+------+--------+------+----------
(0 rows)

linas:
> in en_pairs_cfive_mst, there is no any LEFT-WALL in parses. Is it a bug or there is another reason

This is a bug. I'm actually rather surprised/confused by this bug, since I thought it had been fixed by then.  It first showed up in the Chinese parses, where there was an off-by-one issue having to do with how java passed these through. I fixed that ... according to my diary, I have claims like: `"The top-ten connector-sets are "LEFT-WALL: He+;" "LEFT-WALL: The+;" "LEFT-WALL: “+;" "LEFT-WALL: "+;" ".: "+;" "LEFT-WALL: I+;" "and: ,-;" "LEFT-WALL: It+;"  "LEFT-WALL: She+;" ".: ”+;" he first disjunct with more than one connector in is is "It: LEFT-WALL- "` -- so clearly, I analyzed some dataset that had this in it. Are you sure you did the SQL query correctly?

linas:
... it is also possible that dataset is broken. I've recomputed it many many times, each time after another bugfix ... I thought I found found and fixed them all .. but .. anything is possible.

leungmanhin:
@linas, just curious, are we currently using RIGHT-WALL for the parse?

akolonin:
@linas, in en_pairs_cfive_mst there are just no atoms carrying anything looking like LEFT-WALL in their name. If that may be critical for production of correct dataset and you manage to re-create it, let me know, so I re-export. 
BTW, please check my other yesterday email with on sheaves with questions.

linas:
In the English dictionary, RIGHT-WALL is used to link punctuation at the end of sentences, and smoe oddball cases where that punctuation is absent.   For the language-learning, I don't believe that RIGHT-WALL is ever used.  Its there, its linked, but the code explicitly ignores it.  Or at least, it used to explicitly ignore it... It did not seem important for language-learning ... but .. maybe it is, if you want to know when sentences actually stop.

linas:
OK, so merging then.

PR #2995, opened 2018-01-30
Fixed AtomSpacePublisherModuleUTest failure

DeborggraeveR:
Both close and destructor of zmq::context_t call the same code, which causes the destructor to fail.
Not sure if this is a bug in zmq, but at least this commit should prevent the unit test from failing

from zmq.hpp:
```
        inline ~context_t () ZMQ_NOTHROW
        {
            int rc = zmq_ctx_destroy (ptr);
            ZMQ_ASSERT (rc == 0);
        }

        inline void close() ZMQ_NOTHROW
        {
            int rc = zmq_ctx_destroy (ptr);
            ZMQ_ASSERT (rc == 0);
        }
```


linas:
thanks!

PR #2997, opened 2018-02-02
Advance URE pattern miner

ngeiswei:
This is the current a prototype of the URE-based pattern matcher. By URE-based I mean that it runs **entirely** on the URE.

There are basically 4 rules:
1. create shallow abstractions from text
2. create top abstraction (pattern that matches any text)
3. specialize abstraction by composing abstractions
4. reduce variables in abstractions (not implemented yet)

Combining these 4 rules on a query like
```
Evaluation
  Predicate "minsup"
  List
    Variable "$Pattern"
    <text>
    <ms>
```
allows the URE to discover patterns with a certain minimum support `ms`.

More needs to be work, in particular to implement the variable reduction rule and refine the existing rules to take advantage of some optimizations. Then it will have to be extended to support surprisingness, conditional patterns (like ImplicationScope, etc).


linas:
what does "create shallow abstractions from text" mean?

ngeiswei:
A shallow abstraction of a text (according to my made-up terminology) consist for a lambda with a body starting with the same link as the text, and only variables as outgoings, for instance, if the text is
```
Inheritance
  Concept "A"
  Concept "B"
```
then its corresponding shallow abstraction is
```
Lambda
  VariableList
    Variable "$X"
    Variable "$Y"
  Inheritance
    Variable "$X"
    Variable "$Y"
```
Types in the variable declaration are ignored for now.

linas:
Ah. OK. At some point, we should have a deeper discussion about general principles. Certainly, the "language learning code" is trying to do something like this, but based on various measures of statistical likelihood.   So, for example,  inferring

```
 Lambda
    Variable "$Y"
  Inheritance
    Concept A
    Variable "$Y"
```
because it saw lots of different things in position Y matched with `Concept A`, and never saw Inheritance other than from A.    Also what we call clustering or "grammatical categories --

```
 Lambda
    Variable "$Y"
  Inheritance
    (Concept A  or Concept Q or Concept Z)
    Variable "$Y"
```
because there were many different kinds of $Y's, but only those three things A, Q, Z (usually) were observed in the $X position.   I say "usually", because if there are a handful of exceptions, there's a way for dealing with those.   At any rate, I've got a big splotch of code for extracting complex patterns like this, based on statistical observations.

linas:
"complex patterns" the above is actually a super-duper simple pattern. the "complex patterns" are patterns-of-patterns, that's where the action is.

PR #3000, opened 2018-02-07
Build Python3 by default, if Python3 is found.

linas:
This requires opencog/atomspace#1567 as a pre-requisite.

It is strongly suggested that, when pulling and building, that a `make clean` be done first, as well as removing the `build/CMakeCache.txt` file, as otherwise there is a strong risk of mixing python2 and python3 together in the libraries, which will cause some unit tests to fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -385,18 +385,13 @@ IF (HAVE_PY_INTERP)
 		# differ on each operating system.
 		ADD_DEFINITIONS(-DHAVE_CYTHON)
 		SET(HAVE_CYTHON 1)
-		EXECUTE_PROCESS(
-			COMMAND ${PYTHON_EXECUTABLE} -c "from distutils.sysconfig import *;print(get_python_lib())"
-			OUTPUT_VARIABLE PYTHON_DEST
-		)
 
-		# Replace new line at end.
-		STRING(REPLACE "\n" "" PYTHON_DEST "${PYTHON_DEST}")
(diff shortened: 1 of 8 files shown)


edyirdaw:
@linas I don't know what to say, but its a great way to simplify dev with opencog and the timing seems perfect. I will try it out. Thanks very much. Lunch's on me :D

edyirdaw:
I tested it out and it works perfectly in working with my application with ghost with 
from opencog.atomspace import AtomSpace
from opencog.scheme_wrapper import load_scm, scheme_eval, scheme_eval_h

but pycharm is redding 'scheme_wrapper' saying it cannot find the reference. But that's beside the point I guess.

linas:
Be careful. pycharm might be finding the old python2 code, which is still installed, unless you removed it.

You should also do this:
```
sudo rm -r /usr/local/lib/python2.7/dist-packages/opencog
```

The new packages are in the same kind of path but python3.5 not python2.7

edyirdaw:
This is what pycharm outputs when hovering above scheme_wrapper

> Cannot find reference 'scheme_wrapper' in '__init__.py' 
> This inspection detects names that should resolve but don't. Due to dynamic dispatch and duck typing, this is possible in a limited but useful number of cases. Top-level and class-level items are supported better than instance items.

linas:
You have two options:
1) stop using pycharm
2) debug it.

I have no clue why pycharm doesn't work correctly.

edyirdaw:
Actually the code runs on pycharm. And yes i will try to see why it complains.

PR #3001, opened 2018-02-08
ODR rule, opencog/atomspace#1568 is needed.

linas:
Another attempt to avoid python difficulties.


linas:
The requires an update of the atomspace. Also, almost surely requires a `make clean` to avoid build errors.

Issue #2991, opened 2018-01-27
Invalid use of globals in openpsi


linas:
The file `OpenPsiRules.h` decalres multiple arrays storing handles as globals. This will often lead to crashes in `cxa_finalize` when the shared library is unloaded. This is the current reason for why  `OpenPsiRulesUTest` and the other OpenPsi tests crash.

Basically, when those tests finish, the C libarary unloads all shared libraries that were used, which causes all finalizers in those shared libraries to run.  Whenever a shared library contains global C++ objects, the finalizer calls the destructors on those objects.  a global `Handle _psi_category` would be an example of a global C++ class.

The problem here is that the finalizers are run in arbitrary order, and there is no way in which to control that order.  In this case, the finalizer that unloaded `libatombase.so` which contains the definition for `Handle` has already run, so the code that defines the `Handle` destructor is now gone.  At some later time, the finalizers for `libopenpsi` are run, which try to call the destructor for `Handle` which does not exist.   Then boom!! an awful bunch of corruption messages spill out.

PR #3003, opened 2018-02-08
Partial fixes for failing openpsi unit tests

AmeBel:
New state
```
>> ctest -I 12,14
Test project /home/amen/src/opencog/opencog/build/tests
    Start 12: OpenPsiRulesUTest
1/3 Test #12: OpenPsiRulesUTest ................   Passed    0.38 sec
    Start 13: OpenPsiImplicatorUTest
2/3 Test #13: OpenPsiImplicatorUTest ...........   Passed    0.50 sec
    Start 14: OpenPsiSCMUTest
*** Error in `/home/amen/src/opencog/opencog/build/tests/openpsi/OpenPsiSCMUTest': free(): corrupted unsorted chunks: 0x0000000002840020 ***
3/3 Test #14: OpenPsiSCMUTest ..................***Exception: Other  8.75 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =   9.63 sec

The following tests FAILED:
	 14 - OpenPsiSCMUTest (OTHER_FAULT)
Errors while running CTest
```
* If cmake is configured in debug mode, by running` cmake -DCMAKE_BUILD_TYPE=DEBUG ..`  for both atomspace and opencog all tests pass.

Issue #2991 still not fixed.


opencog/openpsi/OpenPsiImplicator.cc
```diff
@@ -127,3 +126,9 @@ Handle OpenPsiImplicator::imply(const Handle& rule)
     return Handle::UNDEFINED;
   }
 }
+
+OpenPsiImplicator& opencog::openpsi_implicator(AtomSpace* as)
+{
+  static OpenPsiImplicator implicator(as);
+  return implicator;
```
linas: I believe the issue will persist as long as you have statics like this.   The fix would be to code this as
```
OpenPsiImplicator& opencog::openpsi_implicator(AtomSpace* as)
{
   static OpenPsiImplicator implicator * = nullptr;
  if (nullptr ==  implicator) implicator = new OpenPsiImplicator(as);
  return *implicator; 
}
```
This will leak memory upon process exit, but it won't ever get a double free. 
You might want to double-check that this doesn't get called with multiple 
different atomspaces, as that might lead to trouble.  Some day in the future,
we will need to find a generic solution for this kind of thing, but that's for .. later.
The new to a static pointer will fix it.

PR #3013, opened 2018-02-20
Use microseconds for rent collection in ECAN

leungmanhin:


opencog/attention/AFRentCollectionAgent.cc
```diff
@@ -59,23 +59,23 @@ void AFRentCollectionAgent::selectTargets(HandleSeq &targetSetOut)
...
     // calculate elapsed time Et
-    seconds elapsed_time = duration_cast<seconds>
-                           (high_resolution_clock::now() - last_update);
+    microseconds elapsed_time = duration_cast<microseconds>
+                                (high_resolution_clock::now() - last_update);
 
-   // Without this w might always be zero. Which will make rent to be zero all
-   // the time.
-    if (elapsed_time.count() <  1/update_freq )
+    // Without this w might always be zero. Which will make rent to be zero all
```
misgeatgit: Thanks! This comment might not make sense anymore though. The if statement would only be useful in preventing the agent from running more frequently and waste CPU time for calculating very small amount of rents.

leungmanhin: Yes just removed that too

PR #3016, opened 2018-02-26
add function in xpatternminer unit test

Bitseat:
@ngeiswei Here is the function and the files I added.

### DIFF
--- a/opencog/learning/xpattern-miner/ure-based/rules/pattern-miner-utils.scm
+++ b/opencog/learning/xpattern-miner/ure-based/rules/pattern-miner-utils.scm
@@ -4,9 +4,9 @@
 (use-modules (opencog rule-engine))
 (use-modules (srfi srfi-1))
 
-(cog-logger-set-level! "debug")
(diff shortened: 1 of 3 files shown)


linas:
Unit tests should not depend on immense datasets.  Large datasets can be tested with system tests; unfortunately, right now, we do not have any policy for system tests, or any repository in which they could be placed.   A discussion and resolution is needed on the mailing list.

Please do not merge this, until the datasets are cut down to several hundred lines of code at most.

linas:
Also, given that this is testing lojban, it would be better to create a brand-new, stand-alone `LojbanUTest` instead of shoe-horning this into an existing unit test.

ngeiswei:
Agreed. Though I don't mind too much having this test inserted in `XPatternMinerUTest.cxxtest` since it's only purpose is to test the pattern miner, independently of how the lojban parser works. Moving it to a standalone `LojbanUTest` would certainly be justified if the unit test relied on calling the lojban parser though.

linas:
Replaced by #3019

PR #3021, opened 2018-03-05
Nlp learn readme

glicerico:
Modified opencog/nlp/learn README for better understanding of some steps and removed references to unused relex server.


linas:
Looks reasonable. The only reason some of the steps got numbered 9a) and 9b) and so on is that I was too lazy to renumber all the other steps, when I had to insert a new step in the middle.

PR #3022, opened 2018-03-06
Simplified nlp running

glicerico:
- Included scripts to simplify running of MST-pipeline; particularly, giving database name and credentials as command line arguments, instead of modifying different files. These scripts were originally written by Zed Redstone.

- Also added info on nlp/learn/README on how to activate @leungmanhin 's changes to output sentence parses to file


glicerico:
Hi @linas Any comments about accepting this PR?

linas:
looking now

opencog/nlp/learn/README
```diff
@@ -104,18 +104,16 @@ statistics.  Most of this revolves around setting up postgres.
    https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26616
    (in guile-2.2, it doesn't hang, but still behaves very badly)
 
-1) Set up and configure postgres, as described in
-   `atomspace/opencog/persist/sql/README.md`
+1) Install `link-grammar-5.3.15` or newer, either from here:
```
linas: The current version of LG is 5.4.3  and contains fixes needed by other parts of the system 5.3.15 is too old.   Version 5.5 will be available in a few weeks or a month

linas: anyway, this is kind-of-obsoltee; the opencg repo should be build with link-grammar, so this is now a step-zero kind of thing: it has to be done before compiling opencog.

opencog/nlp/learn/README
```diff
@@ -141,27 +143,25 @@ So, set up the text-ingestion pipeline:
...
+   credentials. See next step.
 
-6) Select a language, and copy one of the `pair-count-??.scm` files
-   from the `run` directory to your working directory. Review and edit
-   the configuration as necessary. It contains the REPL port number,
-   and a default prompt: These are set up by default to avoid conflicts
-   and confusion, and to allow multiple languages to be processed at
-   the same time.  It also contains the database credentials -- that's
-   probably the only thing you need to change.
-
```
linas: what happened to step 6? The configurations are somewhat different depending on the language you choose to work with.

opencog/nlp/learn/README
```diff
@@ -180,7 +180,7 @@ So, set up the text-ingestion pipeline:
    for parsing, counts the returned word-pairs, and stores them in
    the database.
 
-8a) Verify that the above resulted in data sent to the SQL database.
+7a) Verify that the above resulted in data sent to the SQL database.
```
linas: Once again -- if you are renumbering everything anyway, you may as well get rid of the a,b,c numbering too.  The only reason that there is an a,b,c is because I did not want to waste time renumbering everything. its pointless to preserve this hack.

opencog/nlp/learn/README
```diff
@@ -286,35 +286,31 @@ alpha-pages after all that work)
...
+  scripts into your working dir, if you didn't do it before.
 * Modify as needed for chosen language. Most of them don't need modification,
   by default, unless you want to do something unusual.
 
-  -- `run-all-servers.sh` runs the cogserver for
-     one language, in a multi-paneled `byobu` window.
-     It doesn't start the final batch script `./wiki-ss-<lang>.sh`,
-     which needs to be started by hand (for the <lang> that you want to
-     process). Use F4 to move to the 'parse' tab and start it there.
-
```
linas: what happened to run-all-servers.sh?  I can see you trying to o it with just once script below, but you still need to have a byobu window so that you can control processes all at once.   I've been running with 3 or 4 or 5 virtual machines, each running different stages of the pipeline, ad you can just do this all in just one window -- you need multiple windows, one or two for each VM, and then have byobu in each window, just so the whole thing doesn't spiral out of control.   Also, some of these may be working on different languages. ... the ability to monitor progress is very important.

opencog/nlp/learn/README
```diff
@@ -286,35 +286,31 @@ alpha-pages after all that work)
...
-          (sql-open "postgres:///en_pairs?user=ubuntu&password=asdf")
+        guile -l observe-launch.scm  -- --db ??-pairs --user your_user --password asdf
       ```
-      to contain your database name, username and password (if needed)
+      (--user and --password are optional, as explained above).
 
-* Run `run-all-servers.sh` in your working directory.
-* In the byobu terminal, navigate to the 'parse' tab (with F4),
-  go back to the input text dir (the dir that contains `beta-pages`), and
+* In a separate terminal (same working directory),
```
linas: No not in a separate terminal. Run it all in byobu. If you have five different sessions going at once, and each secsion takes 3 or 4 or 5 windows -- you get 25 windows .. that would be a horrible disaster.  Its all got to fit into one window, else its just not scalable.

opencog/nlp/learn/README
```diff
@@ -344,24 +340,22 @@ where type 7 is `CountTruthValue`.
...
 ```
-	(use-modules (opencog) (opencog persist) (opencog persist-sql))
-	(use-modules (opencog matrix))
-	(use-modules (opencog nlp) (opencog nlp learn))
-	(sql-open "postgres:///en_pairs?user=ubuntu&password=asdf")
-	(define ala (make-any-link-api))
-	(define asa (add-pair-stars ala))
-	(batch-pairs asa)
-	(print-matrix-summary-report asa)
+guile -l compute-mi-launch.scm  -- --db ??-pairs --user your_user --password asdf
```
linas: Careful adding premature automation.  Some things are better done by hand. Scripting too much too soon is dangerous.  This part might be OK, but at this stage, I'm not sure its robust enough to be fully automated.

opencog/nlp/learn/README
```diff
@@ -418,38 +409,43 @@ Minimum spanning tree code is in `mst-parser.scm`. The current version
...
 
-* Start the REPL server by running ./run-all-servers.sh or
+* If you are interested in the actual sentence parses, open
+  `mst-one.sh` and change the observe variable from `observe-mst` to
+  `observe-mst-extra`. Parses will be written in `mst-parses.txt`.
+
+* Start the REPL server by running 
 ```
-  guile -l mst-count-en.scm
+  guile -l parse-launch.scm -- --db new_dbname --user your_user --password asdf
```
linas: Parsing was done in the earlier step. This is the MST step, not the parse step.   Call it "mst-launch" or something like that.

opencog/nlp/learn/run/mst-one.sh
```diff
@@ -33,7 +34,7 @@ observe="observe-mst"
 # 	exit 1
 # fi
 # Alternate cogserver test: use netcat to ping it.
-haveping=`echo foo | nc -N $coghost $cogport`
+haveping=`echo foo | nc $coghost $cogport`
```
linas: Some versions of nc have to have the -N flag, and without it, nc will hang forever and do nothing. Other versions of nc don't need the -N flag.   Unfortunately, nc does not print what version it is, so I don't know how to tell when -N is needed, or not.

opencog/nlp/learn/run/observe-launch.scm
```diff
@@ -0,0 +1,50 @@
...
+; counting pipeline. Starts the cogserver, opens the database,
+; loads the database (which can take an hour or more!)
+;
+(use-modules (opencog) (opencog persist) (opencog persist-sql))
+(use-modules (opencog cogserver))
+(use-modules (opencog nlp) (opencog nlp learn))
+(use-modules (system repl common))
+(use-modules (system repl server))
+
+(repl-default-option-set! 'prompt "scheme@(en-pairs)> ")
```
linas: well, the point of having the prompt in a config file is to have differrent prompts for different languages -- they are not all going to be 'en', and it can get very confusing if you are processing multiple languages at the same time. This really belongs back in the config file, like it used to be.

Also -- different stages -- parsing, computing the MI, performing the MST, performing the disjunct extraction, and performing the merges -- these all should get different prompts so tht one can tell which is which from the prompt. (so, for one language - english - there will be five different prompts.

opencog/nlp/learn/run/observe-launch.scm
```diff
@@ -0,0 +1,50 @@
...
+; Edit the below, setting it to the desired langauge.
+; This sets the cogserver port-number and the prompt-style.
+(start-cogserver "opencog-en.conf")
+
+; Open the database.
+(sql-open database-uri)
+(display "Opened database: ")
+(display database-uri)
+(display "\n")
+(display "Checking for empty database\n")
```
linas: why are you checking for an empty database???

opencog/nlp/learn/run/observe-launch.scm
```diff
@@ -0,0 +1,50 @@
...
+)
+
+(define atom-count (cnt-all))
+(if (> atom-count 9)
+    (begin
+        (display "CAUTION: Database has ")
+        (display atom-count)
+        (display " atoms\n"))
+    (begin
+        (display "Database is EMPTY and ready to observe.\n")))
```
linas: The database will typically never be empty, unless you just created it.  Why would you ever want it to be empty????  I don't understand. Empty==bad, so if you want to print a  caution, it should be exactly the other way around ...

opencog/nlp/learn/run/observe-launch.scm
```diff
@@ -0,0 +1,50 @@
...
+(load "utilities.scm")
+
+; Get the database connection details
+(define database-uri (get-connection-uri))
+
+; Start the cogserver.
+; Edit the below, setting it to the desired langauge.
+; This sets the cogserver port-number and the prompt-style.
+(start-cogserver "opencog-en.conf")
+
```
linas: just like beofer -- why are you starting a cogsrever?

opencog/nlp/learn/run/parse-launch.scm
```diff
@@ -0,0 +1,57 @@
...
+
+; Print the sql stats
+(sql-stats)
+
+; Clear the sql cache and the stats counters
+(sql-clear-cache)
+(sql-clear-stats)
+(print-matrix-summary-report star-obj)
+
+(display "Done fetching pairs.\n")
```
linas: Parsing was done earlier. This is the MST disjunct extraction stage, not the parse stage.

opencog/nlp/learn/README
```diff
@@ -253,21 +253,21 @@ So, set up the text-ingestion pipeline:
...
 
-11) Using the extracted mutual information for word-pairs, the MST
+10) Using the extracted mutual information for word-pairs, the MST
    parsing step can be started. See below. Run this for several days.
    At the end of this, you will have a bunch of statistics on
    "pseudo-connector sets": pairs which associate a word to a
    disjunct of pseudo-connectors.
 
-12) Perform clustering research.  You're on your own, now.
+11) Perform clustering research.  You're on your own, now.
```
linas: There are now several additional steps after this.  After obtaining a collection of disjuncts, you will probably want to compute the marginals and the MI for those.  I have an assortment of half-finished, half-working scripts that attempt to consolidate or "cluster" the disjuncts -- and those require assorted addtional scripts to be run.  But its still cutting edge, I guess...

linas:
OK, I left a bunch of comments -- they should get cleaned up. The #1 and #1 most important fixes is that each different stage really needs to have a distinct prompt, for both the guile shell and the cogserver (and all of the different cogservers need to have distinct port numbers, so that they don't collide -- and these also are per-langauge)  and everything needs to fit into just one byobu window (so one byobu per language per processing stage.) without this, getting things done is  .. impossible .. as most processing steps take days or longer, an one has to baby-sit all of them all at once.

glicerico:
Great, thanks for all the comments, will work on them. I didn't realize the importance of using byobu before; I will fix that.

linas:
Yeah, the process takes so long that its just easier to run the various stages in parallel, but then it becomes like baby-sitting kindergarten, where each pipeline is getting into some trouble of one kind or another.   Or maybe more like just doing your laundry in 5 different machines all at the same time at the laundromat.

ngeiswei:
@glicerico why did you merge the master into your branch? (see https://github.com/opencog/opencog/pull/3022/commits/aee1aab2948b1163558a177bd4067033975cf81f).

There is no reason to do so unless your branch needs a feature present in the master, otherwise it will just make the git history look more convoluted than it should.

opencog/nlp/learn/README
glicerico: Nice, please include instructions and scripts when you have them ready :)

opencog/nlp/learn/README
glicerico: I changed the guile file to do something easy like just running the commands you provided :)

opencog/nlp/learn/README
```diff
@@ -344,24 +340,22 @@ where type 7 is `CountTruthValue`.
...
+
+The code for computing word-pair MI is in `batch-word-pair.scm`, and called
+by compute-mi-launch.scm. 
+It uses the `(opencog matrix)` subsystem to perform the core work.
 
 Batch-counting might take hours or longer, depending on your dataset
 size. The batching routine will print to stdout, giving a hint of
 the rate of progress.
 
+When finished, stop the cogserver
```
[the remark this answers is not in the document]
glicerico: This is now done in compute-mi.scm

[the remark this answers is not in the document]
glicerico: @linas I understand that batch-pairs stores all data, and I corroborated by looking at the database after running it, where I see the following new data:
27 |     1 |   96 |      0 | *-Entropy Key-*       | 
   28 |     1 |   96 |      0 | *-Mutual Info Key-*   | 
   29 |     1 |   96 |      0 | *-Norm Key-*          | 
   30 |     1 |   96 |      0 | *-FrequencyKey-*      | 
   31 |     1 |   96 |      0 | *-Total Entropy Key-* | 
   32 |     1 |   96 |      0 | *-Right Norm Key-*    | 
   33 |     1 |   96 |      0 | *-Total MI Key-*      | 
   34 |     1 |   96 |      0 | *-Left Norm Key-*     | 

is this enough? should (sql-store) be executed before closing the database? How do I check the database is flushed? You don't have these steps in your README, I am not sure how to do it.

opencog/nlp/learn/run/compute-mi-launch.scm
```diff
@@ -0,0 +1,33 @@
...
+(add-to-load-path ".")
+(load "utilities.scm")
+
+; Get the database connection details
+(define database-uri (get-connection-uri))
+
+; Start the cogserver.
+; Edit the below, setting it to the desired langauge.
+; This sets the cogserver port-number and the prompt-style.
+(start-cogserver "opencog-en.conf")
```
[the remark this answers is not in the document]
glicerico: Got it, removed.

glicerico:
@ngeiswei sorry, I didn't realize this. Will avoid it later on.

ngeiswei:
Thanks @glicerico

opencog/nlp/learn/scm/mst-parser.scm
```diff
@@ -399,6 +399,47 @@
 
 	(close-port file-port)
 )
+
+(define-public (export-mi)
+"
+  Export the word-pairs fMI for
+  every possible pair in the corpus, to allow analysis
+  of the MI-based MST-parser
+
```
linas: It would be better if we did not place cheap ugly hacks in the main body of the code. Something like this really should go into a separate directory, where you are maintaining your personal research tools. I have been using the `nlp/learn/learn-lang-dairy/` directory for messy stuff like this.  If you feel you need to share this with others, then it should be placed in some collaborative tools directory. It doesn't belong in the main body of processing code.

opencog/nlp/learn/README
```diff
@@ -141,30 +144,44 @@ So, set up the text-ingestion pipeline:
...
+   to your working directory:
+
+   - run-server-parse.sh
+   - observe-launch.scm
+   - opencog-en.conf
+   - utilities.scm
+   - wiki-ss-en.sh
+   - ss-one.sh
+   - submit-one.pl
+   - nonbreaking_prefixes (the complete folder)
```
linas: is there a reason you are not copying the entire directory? There are various useful scripts in there, such as renice.sh that can make a big difference on a running system. Similar remarks for the shutdown scripts, which are needed to reboot containers/VM's/machines without corrupting the data in them.

opencog/nlp/learn/README
```diff
@@ -141,30 +144,44 @@ So, set up the text-ingestion pipeline:
...
+   up by default to avoid conflicts and confusion, and to allow multiple 
+   languages to be processed at the same time. 
+   You need to pass the language and database credentials as arguments; 
+   see next step.
+   `wiki-ss-en.scm` starts the process of observing the text files contained
+   in the beta-pages folder, more on that later (see Bulk Text Parsing).
+   The rest of the files are explained later.
+
+5) Start the REPL server. Eventually, you can use the
+   `run-server-parse.sh` file in the `run` directory to do this;
```
linas: There is no `run-server-parse.sh` file!

opencog/nlp/learn/README
```diff
@@ -414,44 +444,62 @@ information between word-pairs (so we are maximizing MI). Thus, MST
 parsing cannot be started before the above steps to compute word-pair
 MI have been accomplished.
 
-Minimum spanning tree code is in `mst-parser.scm`. The current version
+Minimum spanningutree code is in `mst-parser.scm`. The current version
```
linas: typo here.

opencog/nlp/learn/README
```diff
@@ -414,44 +444,62 @@ information between word-pairs (so we are maximizing MI). Thus, MST
...
 To perform MST parsing in bulk:
-* Copy `run/mst-count-en.scm`, `run/mst-one.sh` and `run/wiki-mst-en.sh`
-  to your working directory.
+* Review the file `opencog/nlp/learn/run/README`, and then copy the
+  scripts into your working dir, if you didn't do it before.
+  `run-server-mst.sh` starts the cogserver and sets a default prompt: set 
+  up by default to avoid conflicts and confusion, and to allow multiple 
+  languages to be processed at the same time. 
+  You need to pass the language and database credentials as arguments; 
+  see next stepsu
```
linas: another typo

opencog/nlp/learn/run/run-server-mst.sh
```diff
@@ -0,0 +1,76 @@
...
+elif [ "$1" = "lt" ]
+then
+  PORT=19002
+elif [ "$1" = "pl" ]
+then
+  PORT=19004
+elif [ "$1" = "yue" ]
+then
+  PORT=19006
+elif [ "$1" = "zh" ]
```
linas: All this is crazy overkill. There are thousands of languages; we won't be listing them all here. If and when Anton wants to run the pipeline on russian, it does not make sense to edit this file to add Russian.   Its very important in software to keep things as simple as possible. This is adding complexity for no reason. 

I'm also nervous about the lang-username-password flags in the scripts as well. They make the scripts long, verbose, difficult to understand, difficult to debug.  The system is already complex, we should not be adding additional complexity, when there is little or no point to it.

linas:
OK, I merged this.  However:  in the future, do NOT create such super-giant pull requests. Pull requests should be bite-sized, easy-to-chew and digestable, so that they can be easily reviewed and corrected when there is a problem.  This pull request had a huge number of problems with it, which turns it into a textbook example of how not to do pull requests. When you do it like this, it just delays or tanks the merge, and it makes the review much harder, as it requires a second, third review, which just uncovers yet more problems, causing more delays. Do not write code like this. Its bad progrmming practice. 

Anyway: there are still several minor problems that need to be corrected: there are several errors in the README, which look like your text editor or IDE is somehow corrupting the contents. (A smaller pull req would make it easier to spot this).

Also: please move the export-mi function to a personal experimental directory. 

Create distinct pull requests for each block of changes. Do not ever again lump three of four different, unrelated changes into one single giant massive pull request.

opencog/nlp/learn/scm/mst-parser.scm
glicerico: I put the offending code in a separate file, as suggested. I put this file in the nlp/learn/run folder... is that OK? It felt unnecessary to add a separate directory for only one file

opencog/nlp/learn/README
glicerico: Just following your README instructions, where you first instruct to copy only some files. But no problem, changed to copy all directory to your workdir.

opencog/nlp/learn/README
glicerico: Yes, it is: https://github.com/opencog/opencog/blob/master/opencog/nlp/learn/run/run-server-parse.sh

opencog/nlp/learn/README
glicerico: fixed

opencog/nlp/learn/README
glicerico: fixed

opencog/nlp/learn/run/run-server-mst.sh
glicerico: Linas, I just added them for completeness for the languages that you already had pre-set. I agree that there is no point in having this, but you were the one who introduced it.
About the lang-username-passwords, it really makes it easier to introduce these in command line, rather than going into a few files and changing them manually each time I want to test a new corpus or a different database. The language part is, again, to comply with your multi-language requirements.

Issue #3033, opened 2018-04-03
Clique pair counts disregard repeated words in a sentence.


glicerico:
In /opencog/nlp/learn/scm/link-pipeline.scm, lines 435-437, when choosing 
```
    (update-clique-pair-counts sent 6 #f)  
```
over 
```
    (update-lg-link-counts sent)
```
the tokens don't distinguish different word instances. This has the effect of getting wrong counts, and not counting word-pairs that have the same word. E.g. In the sentence "I saw a saw", I get the following words (missing the second saw):
```
(WordSequenceLink (ctv 1 0 1)                        
   (WordNode "###LEFT-WALL###" (ctv 1 0 24))         
   (NumberNode "0.000000")                           
)                         
(WordSequenceLink (ctv 1 0 1)                        
   (WordNode "a" (ctv 1 0 24))                       
   (NumberNode "3.000000")                           
)                         
(WordSequenceLink (ctv 1 0 1)                        
   (WordNode "saw" (ctv 1 0 24))                     
   (NumberNode "2.000000")                           
)                         
(WordSequenceLink (ctv 1 0 1)                        
   (WordNode "." (ctv 1 0 24))                       
   (NumberNode "5.000000")                           
)                         
(WordSequenceLink (ctv 1 0 1)                        
   (WordNode "I" (ctv 1 0 24))                       
   (NumberNode "1.000000")                           
)
```
Which results in the pair saw-saw not being counted. When using (update-lg-link-counts sent), this pair does get counted.

linas:
Yes, this appears to be a bug. The output should include
```          
(WordSequenceLink
   (WordNode "saw"                     
   (NumberNode 4)               
)
```
I don't know why it's not there.

linas:
This is fixed in #3034 right?  I'm closing this since I assume its fixed.

glicerico:
Yeah, fixed there.
a.

PR #3034, opened 2018-04-03
Fix repeated words

glicerico:
Fixed bug that ignored repeated words in clique-pair counting, reported in https://github.com/opencog/opencog/issues/3033
Also, only create LinkPairs when the words are withing allowed distance. This had caused me problems before in some tests.


linas:
thanks

PR #3035, opened 2018-04-03
Minor changes to Ghost

leungmanhin:
- Make the rules slimmer
- Use the existing API `psi-get-rules`
- Update the README


opencog/ghost/README.md
```diff
@@ -109,3 +111,76 @@ One can quickly test if a rule can be triggered by using `test-ghost`:
...
+             (opencog ghost))
+```
+
+7) Before creating any rules, run
+
+```
+(ecan-based-ghost-rules #t)
+```
+
+Note, rules being created after running this will be slimmer (preferred) and can only work with this ECAN approach. They are NOT backward-compatible with the `test-ghost`.
```
AmeBel: what does `slimmer` mean? And why two types of rules?

leungmanhin: It has fewer atoms, the ones that are for `DualLink` are not needed for this ECAN approach

PR #3047, opened 2018-04-11
Optimize pattern miner

ngeiswei:
The URE pattern miner performances are now decent.

These optimizations are entirely based on reason, as opposed to experimental profiling using standard tools like valgrind, however the speed-up is tremendous (from +inf to 30s). It's still much slower than its C++ cousin that passes all unit tests in 4s (it has also been optimized), but it's decent and I believe usable. I do not know what accounts for the slow down, I think it could entirely be explained by the overhead of the URE itself.

What remains to be done:
1. Improve documentation (both README.md and wiki)
2. Experiment with heuristics, as well as surprisingness
3. Probably more optimization

### DIFF
--- a/opencog/learning/PatternMiner/PatternMiner.cc
+++ b/opencog/learning/PatternMiner/PatternMiner.cc
@@ -2025,7 +2025,7 @@ unsigned int PatternMiner::getAllEntityCountWithSamePredicatesForAPattern(const 
             return eit->second;
         }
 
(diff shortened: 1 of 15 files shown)


ngeiswei:
Oops, there's one problem, opencog needs to be compiled and installed in order to pass the URE pattern miner unit test. I don't know if that's a big problem, I vaguely recall a debate of that issue but I forgot the conclusion (whether we should care or not).

PR #3068, opened 2018-04-27
Ecan spread only to atoms in the present atomspace

AmeBel:


opencog/attention/ImportanceDiffusionBase.cc
```diff
@@ -255,8 +255,17 @@ HandleSeq ImportanceDiffusionBase::incidentAtoms(Handle h)
...
-    // Add the incoming set
-    h->getIncomingSet(back_inserter(resultSet));
+    // Add the incoming set only found in the present atomspace, because
+    // if on another thread the incoming-set is being modified, for example
+    // a query that uses transient atomspaces is being processed, we don't
+    // want to diffuse to transient atoms created.
+    // TODO: How to handle cases when the other atomspaces are not transient
+    // but are a child or parent of the present atomspace?
+    IncomingSet hIncomingSet = h->getIncomingSet(_as);
+    for (const auto i : hIncomingSet)
```
misgeatgit: Could you change the syntax to ```resultSet.insert(resultSet.end(), hIncomingSet.begin(), hIncomingSet.end());``` so that it has similar syntax to line 273 and  avoids copy overhead?

PR #3089, opened 2018-05-08
Add Ghost predicates for checking if anyone started/stopped talking in a given time.

AmeBel:


AmeBel:
This also add utilities for tracking the true and false transition times, that should be provided by the time-server.

PR #3092, opened 2018-05-08
Add some linked libraries back to fix unit test compilation

ngeiswei:


ngeiswei:
@linas This fixes something you have apparently introduced in https://github.com/opencog/opencog/commit/befe9d55ceeead8b5f37310781f5c295661e0518 This is completely trial-n-error based. Please review.

linas:
I'm fixing the tests directly, rather than the cogserver. The cogserver doesn't have any missing symbols; you can check with `ldd -r ./opencog/cogserver/server/libserver.so`  so does not need any more libraries to satisfy the link step.  However, I see now some of the unit tests are missing required libraries; stupidly I forgot to check before pushing.

linas:
Closing, I think #3095 accomplished the same thing, but in a more modular fashion

ngeiswei:
I'm still having a linking error with it
```
/usr/bin/ld: CMakeFiles/AtomSpacePublisherModuleUTest.dir/AtomSpacePublisherModuleUTest.cpp.o: undefined reference to symbol '_ZN5boost6system15system_categoryEv'
//usr/lib/x86_64-linux-gnu/libboost_system.so.1.62.0: error adding symbols: DSO missing from command line
tests/persist/zmq/events/CMakeFiles/AtomSpacePublisherModuleUTest.dir/build.make:121: recipe for target 'tests/persist/zmq/events/AtomSpacePublisherModuleUTest' failed
```

Adding `${Boost_SYSTEM_LIBRARY}` to `AtomSpacePublisherModuleUTest` solves the problem, but that doesn't seem right because that test does use boost.system or any other boost library that may need it (it uses property tree that doesn't seem to depend on boost.system).

Adding `${Boost_SYSTEM_LIBRARY}` to `server` solves the problem as well, however `server` does use some libraries, such as boost.filesystem or boost.asio, that may depend on boost.system. I don't understand why these library wouldn't load boost.system to begin with but after reading a bit boost documentation it seems it doesn't necessarily, see for instance https://www.boost.org/doc/libs/1_67_0/doc/html/thread/build.html

ngeiswei:
Bingo! If I remove the `boost.asio` code from the `server`, specifically from classes `NetworkServer` and `ServerSocket`, that is I remove the `<boost/asio.hpp>` include and everything that depends on it. Then the linking error disappears. Meaning I believe the best fix is to add `${Boost_SYSTEM_LIBRARY}` to `server`.

linas:
This still doesn't feel right, because `ldd -r  ./opencog/cogserver/server/libserver.so` does not display any missing symbols. Continuing discussion on #3092

linas:
This still doesn't feel right, because `ldd -r  ./opencog/cogserver/server/libserver.so` does not display any missing symbols. So the missing symbols have to be coming from somewhere else.  I'm looking now ....

linas:
So `ldd -r opencog/cogserver/shell/libpy-shell.so` shows missing symbols...  so does
```
opencog/learning/PatternMiner/libPatternMinerAgent.so
opencog/nlp/wsd/libwsd.so
opencog/visualization/dotty/libdotty.so
```

The atomspacePublisher never builds on my system, I am investigating now

linas:
By the way, the order in which link libraries are specified actually matters; later libraries resolve symbols missing in earlier ones, assuming that they have not yet been resolved in some different fashion.

linas:
So, after building ... `ldd -r ./opencog/cogserver/modules/events/libatomspacepublishermodule.so` shows a vast number of unresolved symbols. -- the cogserver, the logger, the atomspace, truth values, json, bunch of different boost libraries -- It looks like it never got properly linked.  Fixing this now.

linas:
... and its actually libjson_spirit that had the missing boost symbols ...

PR #3097, opened 2018-05-09
Fix linking error related to boost.system

ngeiswei:
I believe this properly addresses issue #3092


ngeiswei:
Please @linas review as, although I feel confident about it, I'm not a lib linking guru either. The justification for the fix is further explained in that comment https://github.com/opencog/opencog/pull/3092#issuecomment-387728326

linas:
This still doesn't feel right, because `ldd -r  ./opencog/cogserver/server/libserver.so` does not display any missing symbols. Continuing discussion on #3092

linas:
closing, again, #3099 should do the trick. Its was json_sprit that was using boost, but not linking to it, and then the atomspacepublisher used json_psirit, again, without specifying boost

PR #3099, opened 2018-05-10
resolve all symbols in all shared libraries

linas:
Many old/crufty CMakefiles had failed to specify needed libraries dependencies, leading to confusing linking errors. The root cause of #3092 appears to have been our home-grown libjson_spirit which failed to link the needed boost libraries.  I believe this obsoletes #3097


ngeiswei:
It all works for me, thanks!

PR #3115, opened 2018-05-14
Update nlp/atom_types.script

rTreutlein:
Added descriptions to SetTypeLink and SetSizeLink

### DIFF
--- a/opencog/nlp/types/atom_types.script
+++ b/opencog/nlp/types/atom_types.script
@@ -154,5 +154,5 @@ LG_LINK_INSTANCE_LINK <- LIST_LINK
 // Types used in the Lojban Parser
 // Allows one to specify a set by the Concept that all members of it inherit
 SET_TYPE_LINK <- ORDERED_LINK
-// Allows one to specify a set by the Number of Members it has
+// Allows one to specify a set by the number of members it has
 SET_SIZE_LINK <- ORDERED_LINK
--- a/opencog/nlp/types/atom_types.script
+++ b/opencog/nlp/types/atom_types.script
@@ -152,7 +152,7 @@ LG_LINK_INSTANCE_LINK <- LIST_LINK
 
 // ---------------------------------------------------------------
 // Types used in the Lojban Parser
-// What do thse types do? XXX FIXME - these names are excessively
-// generic, and quite mysterious.
+// Allows one to specify a set by the Concept that all members of it inherit
 SET_TYPE_LINK <- ORDERED_LINK
+// Allows one to specify a set by the Number of Members it has
 SET_SIZE_LINK <- ORDERED_LINK


linas:
This does not solve the fundamental problem: what are these link types, what do they do, who uses them, and for what?  Why are they needed?

linas:
Also -- what do they have to do with NLP? Why wouyld they be NLP types? And, given that I'm working really hard to get rid of SetLink, I'm concerned that these might encourage new uses of Setlink.  FYI, SetLink is evil; it is a major source of problems in atomese. This is reviewed in opencog/atomspace#1502

linas:
Part of  the confusion is that there already exists a TypeSetLink:  https://wiki.opencog.org/w/TypeSetLink  which is used to specify sizes and ranges of sizes.

More generally, if you want to assign types, there is an entire infrastructure for that: see https://wiki.opencog.org/w/Type_constructor for that.

rTreutlein:
Well the reason why we need them is that in lojban you can talk for example about a set where each member is a dog.
SetTypeLink
    ConceptNode "set123"
    ConceptNode "dog"

I guess it could be turned into:
ForAll
    VariableNode "elem"
    AndLink
        MemberLink
            ConceptNode "set123"
            VariableNode "elem"
        InheritanceLink
            VariableNode "elem"
            ConceptNode "dog"

But you can also talk about a set with 10 elements:

SetSizeLink
    ConceptNode "Set456"
    NummberNode "10"

can't come up with a way to rewrite that with other atoms.

If they are basic enough they could be added to some more general atom-types.script or as ben said be put into their own. I just though since lojban is nlp makes sense to add it there.

linas:
What you are calling a "SetTypeLink" seems to be a special case of a TypedAtomLink  https://wiki.opencog.org/w/TypedAtomLink which allows you to assign a type to any kind of an atom, and not just concepts.

What you are calling a SetSizeLink seems to be a special case of a TypeSetLink https://wiki.opencog.org/w/TypeSetLink which allows you to specify not only a single numerical value, but an range of values (min and max bounds) -- its also applicable not just to sets, but to anything that can have a range of numerical values.

I'm assuming that you invented SetTypeLink and SetSizeLink because you were unaware that these other types already existed, and not because you looked at them, and found them to be somehow inadequate or unsuitable.  If they are inadequate or unsuitable, could you explain why? It would be better to fix the current definitions, rather than to invent brand-new link types to work around bugs.

rTreutlein:
Ben and I came up with those during a discussion on how to represent the above examples.

Maybe I don't quite understand what those links do but I don't see how SetTypeLink is a special case of a TypedAtomLink. TypedAtomLink specifies the type of an Atom. But a SetTypeLink specifies what Concept all the Members of the set inherit. The Type of this Set is just a ConceptNode.
Maybe using "Type" in the name is causing confusion maybe it should be called SetElemsInheritFromLink or something else.

Same for the TypeSetLink. I guess if one somehow listed all elements of the Set you could specify that this List should have a specific length using TypeSetLink. But that seems kind of imperative "do x then the result has to be y" and not declerative. "Set Z has N elems"

Perhaps if you could show me how the above examples would look using TypedAtomLink, TypeSetLink that would help.

linas:
"specifies what Concept all the Members of the set inherit" -- how does that statement differ from what the wiki page says:  "associate a type to an atom." where the "type" would be `ConceptNode "set123"` (from your example) and the "atom" would be `ConceptNode "dog"` (from your example). That is, how does 
```
TypedAtomLink
     ConceptNode "set123"
     ConceptNode "dog"
```
differ from a SetTypeLink?   Doesn't the above TypedAtomLink state exactly the same thing as the SetTypeLink? What is the difference?

linas:
And if you want to say "Set Z has N elements", why cant you say
```
TypedAtomLink
     ConceptNode "Set Z"
     TypeSetLink
          IntervalLink 
              NumberNode 3
              NumberNode 3
```
if you want to say that "Set Z has exactly 3 elements" ... If you want to instead say N for some generic, unspecified N, then ... I don't think we have anything in the existing code base that does this. But this should be solvable without too much difficulty.

linas:
Note, currently, the syntax of TypedAtomLink is such that if you want to say that "dog" belongs to both set 123 and also set 456, you would have to write:
```
TypedAtomLink
     TypeChoiceLink
            ConceptNode "set 123"
            ConceptNode "set 456"
     ConceptNode "dog"
```

As I look on this now, this seems to actually be a design mistake (per issue opencog/atomspace#1502) but back in 2015, when TypedAtomLink got added, it seemed to make sense to do it this way.   If there was any way in which TypedAtomLink was inadequate, it would be this: its unique, and the menu choices have to be explicitly enumerated.

ngeiswei:
Wow, I wasn't aware of these type constructs, that's really cool!

Minor remark, I would have thought that `TypeCoInhNode` would need to exclude its argument type, in order to be the true complementary of `TypeInhNode`, that is
```
TypeCoInhNode "ConceptNode"
```
would group all base types of `ConceptNode` excluding itself.

linas:
https://wiki.opencog.org/w/TypeInhNode and https://wiki.opencog.org/w/TypeCoInhNode are not meant to be "complementary", they are meant to be "opposite", in the standard sense of "opposite" meaning "reverse the direction of the arrow".   I don't know of any standard formal definition of  "complementary". 

At any rate, `TypeInhNode` and `TypeCoInhNode` are just shallow wrappers around the C++ `classserver().isA() method`, valuing to `true` whenever S isA T, conversely, when T isA S. 

This is distinct from assigning types to sets, or assigning other type-like descriptive properties to sets.  I don't particularly have any strong feelings about `SetTypeLink`, etc. except for --  (a) the name suggests something generic, when in fact its very narrow-purpose (b) its undocumented  (c) it might be re-inventing existing capabilities, although, given the rather narrow focus, perhaps its better to let them lay low for now, and see what happens.

AmeBel:
@linas In  https://wiki.opencog.org/w/TypeCoInhNode it states "...such as UnorderedLink, Link, Atom or Value would be considered to be matches." What does it mean to match against a value?

linas:
Hmm. Atoms are a special case of Value; where ever you can put a value, you can put an atom. `classserver().isA(ATOM,VALUE) ` will return true.  So you can say `TypeCoInhNode "Value"` and it won't crash. But I guess its not actually useful for anything at the moment. 

It might be useful in the future, to use the ArrowLink express something like  "when you evaluate this EvaluationLink, you get a TV back".  So maybe for example:
```
ArrowLink
     TypeNode "EvaluationLink"
     TypeNode "TruthValue"
```
or something like that.  Its not really usable, right now, but it leaves the door open for that possibility.

PR #3119, opened 2018-05-15
Support for filtering out certain types of atoms from receiving STI

misgeatgit:
This changes are introduced in response to issue #3107.  One can insert set of atom types that should not receive STI by specifying them [here](https://github.com/opencog/opencog/blob/master/opencog/attention/default-param-values.scm#L166).

### DIFF
--- a/opencog/attention/ImportanceDiffusionBase.cc
+++ b/opencog/attention/ImportanceDiffusionBase.cc
@@ -185,7 +185,7 @@ void ImportanceDiffusionBase::diffuseAtom(Handle source)
         Handle target = it->first;
         auto ij = std::find_if(hsFilterOut.begin(), hsFilterOut.end(),
                 [target](const Handle& h){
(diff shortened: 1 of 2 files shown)


linas:
Umm, OK, issue #3107 was not meant to be urgent: it was a remark that @ngeiswei made in an email, and it seemed like a good idea, so I copied it to an issue to not forget about it.

Unless, of course, it is urgent, and you do need it right away, for other reasons. At any rate, before just jumping in and writing code, you might want express (in #3107) which STI flow to which types, where, why, some general outline of what it could be?  I mean, I won't be using it, but I guess Nil and Amen (pln and ghost) will be, so may as well figure out exactly what nil & amen need?

misgeatgit:
@linas yeah, the implementation is based on our ongoing discussion with the ghost team at the HK office.

misgeatgit:
I'm closing this since the commits are kinda repetitive. Needs some squashing.

PR #3132, opened 2018-05-18
Synchronization with Atomspace: proto library

konstantin-t:
CogServerUTest fails on my computer. But I checked, that it failed before ```proto``` library appeared.


linas:
Thanks. This keeps the opencog repo in sync with opencog/atomspace#1698

PR #3136, opened 2018-05-22
Some inference control learning experiment improvements

ngeiswei:
Mainly add document about distributing the experiments and beyond.


ngeiswei:
That's the document in question https://github.com/opencog/opencog/blob/master/examples/pln/inference-control-learning/distributed-inference-control-learning.md

PR #3145, opened 2018-05-24
Add time windows to some ghost predicates 

AmeBel:
Same us https://github.com/opencog/opencog/pull/3141. This is just syncing


AmeBel:
This pull-request is just correcting work-flow mistake.

PR #3164, opened 2018-05-30
python2 AND 3 bindings 

SomeUserName1:
[Refering to atomspace #1707](https://github.com/opencog/atomspace/issues/1707)

### DIFF
--- a/lib/FindPython2Interp.cmake
+++ b/lib/FindPython2Interp.cmake
@@ -1,0 +1,42 @@
+# Find the Python2 interpreter.
+#
+# NB: Generally, :code:`FindPythonInterp.cmake` as shipped by cmake
+# is able to find a python3 interpreter by working with version numbers. This find module
+# does some gymnastics to be able to find BOTH python2 and python3
+# interpreters within the same project, which does not seem possible
+# with the modules provided from upstream.
+#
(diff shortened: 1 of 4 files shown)


lib/FindPython2Interp.cmake
```diff
@@ -0,0 +1,42 @@
...
+find_package_handle_standard_args(Python2Interp
+                                  REQUIRED_VARS PYTHON_EXECUTABLE)
+
+# Set all those variables that we promised
+set(PYTHON2_EXECUTABLE ${PYTHON_EXECUTABLE})
+set(PYTHON2_VERSION_STRING ${PYTHON_VERSION_STRING})
+set(PYTHON2_VERSION_MAJOR ${PYTHON_VERSION_MAJOR})
+set(PYTHON2_VERSION_MINOR ${PYTHON_VERSION_MINOR})
+set(PYTHON2_VERSION_PATCH ${PYTHON_VERSION_PATCH})
+
```
linas: Why are you even copying this file? It does almost nothing that you need - for example, you never use the above three lines. 

In addition, the the documentation clearly sates that it can find both python2 and also python3, so I don't understand why yo need to make a copy at all?

PR #3195, opened 2018-06-07
nlp_learn port automatization

cuayacas7:
Changes in this commit:
-added an alpha-pages directory with two test text files so that a new person that copies the 'run' directory can test if the pipeline is working
-renamed the cogserver configuration files and moved them all to a subdirectory. This will be needed for standardization for future modifications
-added a script that determines the port number given the processing mode (pair-counting, MI computing or mst-parsing) and the language (same 6 previously existing)


opencog/nlp/learn/run/config/det-port-num.sh
```diff
@@ -0,0 +1,59 @@
...
+      ;;
+   zh)
+      port_end=7
+      ;;
+   *)
+      echo "Usage: ./det-port-num.sh <mode> <language>"
+      echo "<language> must be one of: en, fr, lt, pl, yue, zh"
+      exit 0
+esac
+export PORT=$(($port_ini + $port_end))
```
linas: Missing newline in file.

linas:
Thanks you, this looks very reasonable.

Issue #3212, opened 2018-06-19
Tokenizer breaks MST-parser


glicerico:
During the MST-parse part of the NLP learn pipeline, the tokenizer removes spaces between dots and dashes:
https://github.com/opencog/opencog/blob/60fec1c292b89d913b3cd3200c3e76dd0ee0a0b4/opencog/nlp/learn/scm/mst-parser.scm#L179
The tokenizer in LG parser in "any" mode doesn't include this feature. This causes lines in the text like ". . . . . ." to be transformed into "......" and the MST-parse fails because a word composed of six consecutive dots wasn't counted during the pair-counting phase.

glicerico:
Proposed solution in PR: https://github.com/opencog/opencog/pull/3213

glicerico:
Sorry, I forgot to include how to reproduce this problem. If you use a simple file with three lines like the following:

" mom is a human.

  . . . . . . .

  dad is a human."

And do the whole pair-counting, mutual information, and mst-parsing processes specified in in nlp/learn/README.md, you'll see an error when mst-parsing the second line of the example file ". . . . . . ."

linas:
Sorry, I don't understand the problem.  If all text goes throuogh the tokenizer, how can some text sometimes not go through the tokenizer?

linas:
Oh, I see -- you actually had a string of white-space separated dots in the input.

glicerico:
There are two tokenizers in the NLP pipeline: the LG tokenizer used while pair-counting with LG "any" mode, and the tokenizer (included in mst-parses.scm) used during mst-parsing. This issue is reporting a situation when those tokenizers treat a sentence differently.
Yes, the sentence (coming from a Gutenberg book) is just a number of dots separated by spaces.

linas:
The tokenizer in LG is very complex: it creates a graph of possible parallel paths through a sentence.  The canonical example is "seasand beach" which could be "seas and beach", "sea sand beach" or (after typo correction) "seasoned beech" or simply "(unknown word) beach"  -- 

There's no particularly easy way to port this to opencog; the MST side of things does not know about word-graphs, even if it was ported.  I was hoping to defer action on this until work on morphology begins. 

[...]

PR #3213, opened 2018-06-19
removed merging of dots and dashes

glicerico:
Proposed solution to Issue https://github.com/opencog/opencog/issues/3212

Delete feature to remove spaces between "." and "-" in tokenizer for MST-parsing.

### DIFF
--- a/opencog/nlp/ull-parser/scm/mst-parser.scm
+++ b/opencog/nlp/ull-parser/scm/mst-parser.scm
@@ -152,22 +152,6 @@
 		(if (null? ifx-list) str
 			(pad-dash (pad-a-dash str (car ifx-list) 0) (cdr ifx-list))))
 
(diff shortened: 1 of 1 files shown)


linas:
Well, this "fix" won't work, because it ignores everything that happens between lines 92 and 154 -- that code is inserting the white-space. It inserts too much white-space, and so some of it needs to be removed.   For example, with your change, it would convert "this ... is" into "this . . . is"   and "a--test" into "a - - test"  neither of which are wanted. 

To get the desired behavior, you'd have to create a more sophisticated tokenizer.  The problem with that is sophisticated tokenizers are actually morphological analyzers, and morphology was being put off for a later stage of the project.

glicerico:
Understood why this fix wouln't work, thanks.

PR #3214, opened 2018-06-19
Export mst

glicerico:
Simplified the function that exports mst-parses to file, since we're not using ParseNodes now. I left Manhin's function (commented out), in case we need the parses later. @linas , please advice if this is appropriate.

### DIFF
--- a/opencog/nlp/ull-parser/scm/mst-parser.scm
+++ b/opencog/nlp/ull-parser/scm/mst-parser.scm
@@ -258,154 +258,224 @@
 
 (define-public (observe-mst-extra plain-text CNT-MODE DIST-MODE)
 "
-  observe-mst-extra -- update pseduo-disjunct counts by observing raw
-  text.
-
-  See also 'observe-mst'.
+  observe-mst -- update pseduo-disjunct counts by observing raw text.
 
   This is the second part of the learning algo: simply count how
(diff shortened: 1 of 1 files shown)


linas:
(1) Can you make your changes in the `run-poc` folder instead? This change looks very similar to the code that is already in `run-poc/mst-tools.scm`  that came from Claudia. 

(2) instead of commenting out large blocks of code with semi-colons, try using block comments instead. Block comments are encases in #! ... !#

This folder: https://github.com/opencog/opencog/tree/master/opencog/nlp/learn/run-poc

the file:  https://github.com/opencog/opencog/blob/master/opencog/nlp/learn/run-poc/mst-tools.scm

linas:
I'm closing this because you seem to be unaware of the `run-poc` directory, have forgotten that you have created similar code before, and have ignored previous suggestions, comments and guidelines on what kind of code to submit, and where to make changes.

In the future, before creating new pull requests, please review your changes with either @AmeBel or with @leungmanhin  or with @elggem for general appropriateness an, content and style.

glicerico:
Ah, you're right, I wasn't aware that you removed from your folder the parse-exporting code that @leungmanhin had created before, since you approved those last December. Thanks for the suggestion about block comments! :)

I talked to the guys and I'll try again

PR #3228, opened 2018-06-27
load modus-popens rule from rules/propositional/modus-ponens.scm

noskill:
rules/wip/modus-ponens.scm - current path - does not exist


ngeiswei:
Thanks.

PR #3233, opened 2018-06-28
Swap MemberLink arguments

ngeiswei:
Cousin of https://github.com/opencog/atomspace/pull/1798


linas:
No, this is fundamentally not possible. There are datasets sitting on disk that you don't have access to, that have the MemberLink using the old order. It is not possible to go into those datasets, and alter that order. There are jobs, running right now, that are adding more MemberLinks, at the rate of maybe one every 10-30 minutes or so.

Invent a new link type, say `NewMemberLink` or `SetMemberLink` or `MembershipLink` or `ElementOfLink` that has the arguments in the desired order. However, the old MemberLink is carved in stone, and cannot be changed.

PR #3243, opened 2018-07-02
Various fixes

roman-khimov:
Some things found by static analysis tools and a little wandering around the code. Be careful as I have no experience with this codebase.


opencog/learning/PatternMiner/PatternMinerDistributedWorker.cc
```diff
@@ -505,12 +505,12 @@ bool DistributedPatternMiner::sendRequest(http_request &request, http_response &
             return false;
         }
     }
-    catch (http_exception e)
+    catch (http_exception &e)
```
linas: I'm surprised that your tools do not recommend a `const` here ...

opencog/nlp/lg-parse/LGParseLink.cc
```diff
@@ -298,9 +298,7 @@ Handle LGParseLink::cvt_linkage(Linkage lkg, int i, const char* idstr,
 		char idstr2[37];
 		uuid_unparse(uu2, idstr2);
 		char buff[801] = "";
-		strncat(buff, wrd, 800);
-		strncat(buff, "@", 800);
-		strncat(buff, idstr2, 800);
+		snprintf(buff, sizeof(buff), "%s@%s", wrd, idstr2);
```
linas: It probably does not matter much, but the idea is that the three strncats are faster than the snprintf. Or rather, back in the good old days, they would have been. These days, with compiler magic, who knows.

linas:
Thanks, these changes all look very reasonable.

opencog/nlp/lg-parse/LGParseLink.cc
roman-khimov: I think it would be even better to use plain C++ `string`s here because performance-wise you probably won't notice it anyway (otherwise this code would probably be using `stpncpy()` already) but the code would be even simpler than this one. But it's more invasive and debatable change and I wanted this branch to only contain simple technical fixes. I do have a pair of more invasive changes already in the queue.

opencog/learning/PatternMiner/PatternMinerDistributedWorker.cc
roman-khimov: They usually don't recommend `const`s, but I should've spotted that myself, true.

PR #3244, opened 2018-07-02
cmake: add UUID library dependency for NLP

roman-khimov:
LGParseLink.cc fails to compile without it so a proper check for this library
is needed. FindUUID.cmake taken from CCMPred project which also is AGPL-3.0
licensed (with minor modification).


linas:
Huh. OK, thanks

PR #3246, opened 2018-07-03
Drop HTree class and PatternMinerBF

roman-khimov:
`PatternMinerBF` seems to be broken and unmaintained while `HTree` isn't used properly (and also technically broken).


linas:
I agree, but am leaving this for @ngeiswei as this is someething he's been working on.

ngeiswei:
I'm hesitant to merge that PR as that code will "soon" be removed (when the new pattern miner can do everything it can and more). Before removing it will be tagged so that it can easily be accessed in case some ideas it contains need to be recalled. Thus removing potentially informative code (like PatternMinerBF) even if unmaintained doesn't seem ideal.

I'll probably close it soon. Sorry for the unnecessary work on your part @roman-khimov and thanks anyway.

roman-khimov:
Well, it's not that this removal makes it totally unaccesible, it would be there in the git history, but it won't be stealing anyone's mental energy trying to understand what's all that about. In any case it's more important here to remind that there is a general problem with stale and obviously broken code in the project. I understand that there is a lot of research being done here and all of that has at least some historic value (like "we tried this, but it was bad because of smth, then we tried that but it also failed due to whatever, so now we're doing something completely different, but you can refer to these old attempts if there is a need to"), but that could easily be converted into some README that would contain exactly that --- information about previous approaches to the problem and references to respective git commits that removed them (so that it would be easier to find them). Maybe the project needs to be a bit more aggressive about removing stale/broken/unmaintained things.

Anyway, I'm fine with any kind of decision wrt this PR. It's not hard to remove things and you obviously know better which parts when and how to be removed if removed at all.

ngeiswei:
I don't know well this code, that the other problem, ideally @shujingke should provide her feedback on that since she is the author. As you allude, some mention in some `README.md` of the deleted code would be fine. Maybe @shujingke will show up, let's wait a few days.

linas:
Its been almost two weeks. I see nothing wrong with merging this, and @shujingke  is not responding.  Can we merge this?

PR #3261, opened 2018-07-18
Merge ull back into mainline branch

linas:
All of the ull code is now in a directory called "run-poc". I believe that this is the latest and the best code. 

Please review this code for correctness.


linas:
Hi @glicerico Can you look at this, please?

linas:
So should this be merged?

glicerico:
I have no objection that you merge it (we can't use it as is, but feel free to merge it, it won't affect us).
Since you're waiting for a PR, I'll send an example of how it could work out for both sides, to see if you agree with it.

PR #3264, opened 2018-07-23
changed to link chasing instead of string matching for finding word n…

SomeUserName1:
…odes where possible

### DIFF
--- a/opencog/nlp/sureal/SuRealPMCB.cc
+++ b/opencog/nlp/sureal/SuRealPMCB.cc
@@ -238,15 +238,20 @@ bool SuRealPMCB::clause_match(const Handle &pattrn_link_h, const Handle &grnd_li
         auto it = m_words.find(hPatNode);
         if (it == m_words.end())
         {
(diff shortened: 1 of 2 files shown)


linas:
Thanks. I did not study this carefully, but it seems correct.

What happens if you put `OC_ASSERT(neighbor_win.size() != 0, "Reference Link is missing")` into the code?  Would the assert ever trigger?  If so, then I think that would be due to a bug in RelEx, or due to a bug in `LGParseLink`, since I believe that it should always be the case that a ReferenceLink is present.  Those bugs should be fixed.   You would then be able to remove the if-test `if(neighbor_win.size() != 0)` from the code.

leungmanhin:
The reason of doing it is because there can also be other atoms of the same type being passed to the callbacks. For example `(Concept "apple@123")` is the one that has a `ReferenceLink` linking to it while `(Concept "apple")` doesn't, so the if-test tries to catch the second case

linas:
Ah. OK.

PR #3273, opened 2018-07-27
Fix the load path in PLNRulesUTest

leungmanhin:
It has been failing because some rules couldn't be found


ngeiswei:
Up to this day I'm till not sure how to properly handle Scheme paths. But, although it still requires to `make install` it works. Thanks.

leungmanhin:
Yes actually, none of those `/usr/local/share/*` is needed... and they require `make install`. https://github.com/opencog/opencog/pull/3278 removes them and uses `PROJECT_SOURCE_DIR` instead.

PR #3275, opened 2018-07-27
Dont delete the SystemActivityTable on the end of the serverloop

SomeUserName1:
[See Amens comment on PR 120, cogutil](https://github.com/opencog/cogutil/pull/120)
Fix non-passing unit test

Note: as the SystemActivityTable is erased (SystemActivityTable::halt) in the destructor of the CogServer, this shouldnt cause memory leaks or equal

### DIFF
--- a/opencog/cogserver/server/CogServer.cc
+++ b/opencog/cogserver/server/CogServer.cc
@@ -201,7 +201,6 @@ void CogServer::serverLoop()
 
     // No way to process requests. Stop accepting network connections.
     disableNetworkServer();
-    _systemActivityTable.halt();
 }
 
 void CogServer::runLoopStep(void)


linas:
Can you explain why this change fixes the unit test?

SomeUserName1:
https://github.com/opencog/opencog/blob/master/tests/server/CogServerUTest.cxxtest
Arround line 160 cogserver.serverloop is called 
As it is synchronous SystemActivityTable::halt is called and by that all AgentActivityTable is erased before the code continues
9 lines later the AgentActivityTable is accessed to check the entry size for each instance ran by the cs.serverloop

linas:
Ah, OK, thanks.

Please be aware that the entire cogserver design and concept of agents and activity is terribly wrong, and implemented in just an awful way. It is trying to re-invent operation system concepts, specifically, the concept of "cooperative multithreading", and its doing a bad job of it.  All of the agent code needs to be removed from the cogserver, and placed in it's own module, and redesigned to use threads. The entire Agent API needs to be rethought, from the ground-up.

PR #3280, opened 2018-07-30
Parametrize observe-text

glicerico:
Modified observe-text function in NLP tools, to take as argument the observing mode:
- "lg" mode keeps current default functionality, but takes an integer specifying how many linkages to use for word-pair counting
- "clique" mode observes word-pairs using existing clique-pair counting capabilities, taking an integer to specify window size for observation.

Also included wrapper to old observe-text function for backwards compatibility
@linas , plase review and let me know if this is acceptable

### DIFF
--- a/opencog/nlp/learn/scm/link-pipeline.scm
+++ b/opencog/nlp/learn/scm/link-pipeline.scm
@@ -450,8 +450,10 @@
  an integer parameter:
  - tree: counts pairs of words linked by the LG parser in 'any' language.
  	   'count-reach' specifies how many linkages from LG-parser to use.
- - clique: pairs each word with all neighboring words located less than
-           'count-reach' spaces away.
+ - clique: itearates over each word in the sentence and pairs it with
+           every word located within distance 'count-reach' to its right.
+           Distance is defined as the difference between words positions
+           in the sentence, so neighboring words have distance of 1.
 
  This is the first part of the learning algo: simply count the words
  and word-pairs observed in incoming text. This takes in raw text, gets
--- a/opencog/nlp/learn/scm/link-pipeline.scm
+++ b/opencog/nlp/learn/scm/link-pipeline.scm
@@ -448,7 +448,7 @@
  
  There are currently two observing modes, set by observe-mode, both taking
  an integer parameter:
- - tree: counts pairs of words linked by the LG parser in 'any' language.
+ - any: counts pairs of words linked by the LG parser in 'any' language.
  	   'count-reach' specifies how many linkages from LG-parser to use.
  - clique: itearates over each word in the sentence and pairs it with
            every word located within distance 'count-reach' to its right.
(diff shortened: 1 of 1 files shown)


opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -441,10 +441,17 @@
...
-; ---------------------------------------------------------------------
-(define-public (observe-text plain-text)
+; --------------------------------------------------------------------
+(define-public (observe-text-mode plain-text observe-mode count-reach)
 "
  observe-text -- update word and word-pair counts by observing raw text.
+ 
+ There are currently two observing modes, set by observe-mode, both taking
+ an integer parameter:
+ - lg: counts pairs of words linked by the LG parser in 'any' language.
```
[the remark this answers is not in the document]
glicerico: No problem. Maybe to make it more informative and coherent with the other mode called "clique", how about calling this mode "tree" or "planar-tree", although the second may be too long.

opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -441,10 +441,17 @@
...
+ 	   'count-reach' specifies how many linkages from LG-parser to use.
+ - clique: pairs each word with all neighboring words located less than
+           'count-reach' spaces away.
```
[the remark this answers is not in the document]
glicerico: Let's see if this is clearer ;)

opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -441,10 +441,19 @@
...
-(define-public (observe-text plain-text)
+; --------------------------------------------------------------------
+(define-public (observe-text-mode plain-text observe-mode count-reach)
 "
  observe-text -- update word and word-pair counts by observing raw text.
+ 
+ There are currently two observing modes, set by observe-mode, both taking
+ an integer parameter:
+ - tree: counts pairs of words linked by the LG parser in 'any' language.
+ 	   'count-reach' specifies how many linkages from LG-parser to use.
```
linas: please change this to "any".  There are three different kinds of trees, today: any, ady and amy. Ady splits words into two morpheme (dual) and  amy splits words into multiple morphemes. Because of the way amy works, it might get redesigned in the future to work differently, so we have to leave room for yet other random languages.

In addition, there are theoretical reasons to explore non-tree random parses. These might be useful in handling finnish, turkish texts. Also, the poper handling of semitic texts (hebrew, arabic, ethioprian) is unclear.   That's why the language is called "any" and not "tree".

linas: also, in case its not clear, for the ady random tree language, exploring the clique would be incorrect.  You could make a weak argument that a clique-based approach to the morphology of finnish, turkish might be OK.  Maybe. But a properly tuned version of amy will outperform clique.  I've thought about this more heavily with Lithuanian, where there would be lots of clique-link links between morphemes, but only one link between words.  This would also be the case for the slavic languages.   From what I can tell, all of the romance languages are best handled by ady, and the clique language would be an utter disaster.   English is very special and unusual as a language.

glicerico: Sounds convincing ;)

linas:
thanks

PR #3281, opened 2018-07-31
Fix CogServerUTest take 2

ferrouswheel:
This is similar to #3275 - except that was only a partial fix.

The AgentRunnerBase was relying on a global singleton `cogserver()` to get the SystemActivityTable... this is a different cogserver instance to the one that is run in the test.

PR #3283, opened 2018-08-02
Also pass SystemActivityTable to threaded agent runners

ferrouswheel:
Agents either get run by the SimpleRunner in the CogServer thread or in an AgentRunnerThread.

I previously (#3281) had only dealt with the SimpleRunner case.

This should fix the failure of HebbianCreationModuleUTest and possible others https://github.com/opencog/opencog/pull/3281#discussion_r207089904


linas:
@misgeatgit can you review and merge if this works?

misgeatgit:
Thanks!

PR #3285, opened 2018-08-04
Gradient Descent vs. Graphical Models

linas:
New and improved version that provides more detail, precise formulas, better narrative flow.

Compared to previous version: this spends more time explaining why:

* k-means clustering is identical to matrix factorization (not my result, its an old result)
* Why bicategorization is the right thing to do (because that is what the link-grammar dicts already do!)
* Why an information-theoretic divergence is much better than a cosine distance.

See the PDF for more.


linas:
Arghhh. Forgot the PDF ...

PR #3296, opened 2018-08-17
Function description improvement

glicerico:
- Corrected and added to documentation of observe-text-mode.
- Removed try-catch wrapper. @linas , is it OK to remove this since relex is no longer used?


opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -461,36 +464,19 @@
...
-	; (see documentation for `word-inst-get-word`), the function
-	; `update-clique-pair-counts` might throw.  If it does throw,
-	; then avoid doing any counting at all for this sentence.
-	;
-	; Note: update-clique-pair-counts commented out. If you want this,
-	; then uncomment it, and adjust the length.
-	; Note: update-disjunct-counts commented out. It generates some
-	; data, but none of it will be interesting to most people.
-	(define (update-counts sent)
-		(catch 'wrong-type-arg
```
linas: Have you actually tested, to make sure that there are no other code-paths that might result in a throw?  Large parts of opencog use exceptions in lots and lots of places, for performance:  It is much much cheaper to catch an exception every now and then, than it is to make millions of if-statement checks of return values. That is, relex is not the only thing that throws, there are other blocks of code that also use exceptions to report errors.

opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -461,36 +464,19 @@
...
 	; Count the atoms in the sentence, according to the counting method
 	; passed as argument, then delete the sentence.
+
+	; Note: update-disjunct-counts commented out. It generates some
+	; data, but none of it will be interesting to most people.
 	(define (process-sent SENT cnt-mode win-size)
 		(update-word-counts SENT)
 		(if (equal? cnt-mode "any")
 			(update-lg-link-counts SENT)
 			(update-clique-pair-counts SENT win-size #f))
```
linas: Can you turn this into a case-statement? Currently, you wrote this so that if the mode is "any", then link-counting is used; otherwise, clique-pair counting is used as the default. It would be nicer to make this into an explicit case statement, with the two cases being "any" and "clique". 

Also -- weren't you planning on having other modes besides just these two? Or was this all that was envisioned?

opencog/nlp/learn/scm/link-pipeline.scm
```diff
@@ -59,6 +59,9 @@
 ; as a clique of edges drawn between all words. This is explored further
 ; in the diary, in a section devoted to this topic.
 ;
+; Update: Additional counting modes can be requested by calling
+;         observe-text-mode. See function's documentation for details.
```
linas: Please remove this comment; it serves no purpose.

opencog/nlp/learn/scm/link-pipeline.scm
glicerico: I have parsed a few corpora without this exception, and I haven't found a throw here. Any other particular test you recommend?
The arguments passed to the functions that go inside the exception are a linkage from LG "any" (no throws detected in my experiments), and parameters for the whole run; if the latest are the wrong type, I don't want to skip one sentence, but getting a throw to correct my parameter.
Also, there's another catch one function higher in local-process, so I feel it's safe to remove this. Am I right?

opencog/nlp/learn/scm/link-pipeline.scm
glicerico: Yes, this was only an example of how I was planning to do the function wrappers for backwards compatibility. My next PR will implement the clique-mode counting which accounts for distance, and will be with a case-statement ;)

opencog/nlp/learn/scm/link-pipeline.scm
linas: OK. The word "Update" isn't needed. You could instead say, `Additional counting methods can be specified, including clique counting within a fixed window size. See the function description for additional modes.`

opencog/nlp/learn/scm/link-pipeline.scm
linas: OK

opencog/nlp/learn/scm/link-pipeline.scm
linas: OK

opencog/nlp/learn/scm/link-pipeline.scm
linas: Anyway, the clique-pair counting mode has always been there, from the very begining, and is already described by the docs. So this is not a new mode.  Also, I just now realized that this mode-setting style is not a good API; in the past, you could do both clique-counting, distance counting and "ANY" counting, all at the same time, with just one pass of the text.  With the new API, you would have to run the code three times, to get these three different modes.

You should probably think about some other API -- its a pointless waste of CPU time to run `observe-text` three difrerent times, to get three different counting modes, when running it just once is enough.

glicerico: This makes a lot of sense. I will try to implement it later on. Thanks!

Issue #3303, opened 2018-09-13
atomspacepublisher mis-handling of atoms results in crash


Dagiopia:
the cogserver crashes with the following message in the log when the attention module is loaded after loading the atomspacepublisher
```
[2018-09-13 15:18:14:695] [INFO] Loading module "/usr/local/lib/opencog/modules/libatomspacepublishermodule.so"
[2018-09-13 15:18:14:702] [INFO] [AtomSpacePublisherModule] constructor
[2018-09-13 15:18:14:702] [INFO] Initializing AtomSpacePublisherModule.
[2018-09-13 15:18:16:102] [INFO] Loading module "/usr/local/lib/opencog/libattention.so"
[2018-09-13 15:18:16:129] [ERROR] Not a node! (/home/dagim/opencog/atomspace/opencog/atoms/base/Atom.h:240)
	Stack Trace:
	2: Logger.cc:512	  opencog::Logger::logva(opencog::Logger::Level, char const*, __va_list_tag*)
	3: Logger.cc:524	  opencog::Logger::Error::operator()(char const*, ...)
	4: exceptions.cc:55	  opencog::StandardException::parse_error_message(char const*, __va_list_tag*, bool)
	5: exceptions.cc:82	  opencog::StandardException::parse_error_message(char const*, char const*, __va_list_tag*, bool)
	6: exceptions.cc:142	  opencog::RuntimeException::RuntimeException(char const*, char const*, ...)
	7: Atom.h:240	  opencog::Atom::get_name[abi:cxx11]() const
	8: basic_string.h:109	  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_Alloc_hider::_Alloc_hider(char*, std::allocator<char> const&)
	9: AtomSpacePublisherModule.cc:254	operator()()
	10: [0x26fdd] ??() ??:0
	11: [0x200dc] ??() ??:0
	12: [0x1efd3] ??() ??:0
	13: [0x1aa91] ??() ??:0
	14: [0x1acf9] ??() ??:0
	15: [0x76ba] ??() ??:0
	16: [0x10741d] ??() ??:0
```

I was experimenting with the publisher module so I just went in there to print the message it pushes to the queue and saw the following message:

```
BFD: /usr/lib/x86_64-linux-gnu/libtbb.so.2: warning: loop in section dependencies detected
BFD: /usr/local/lib/opencog/modules/libatomspacepublishermodule.so: warning: loop in section dependencies detected
[...]

PR #3310, opened 2018-09-30
fix for #3303 and other changes on atomspacepublishermodule

Dagiopia:
- fix for #3303 
- changed the library from json_spirit to jsoncpp because the latter is available on all recent ubuntu versions. 
- removed the json_spirit library folder in lib/


linas:
Looks reasonable to me; I did not test. I suspect that neither @vsbogd or @ngeiswei have much to say about this.

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
```diff
@@ -81,91 +81,91 @@ class AtomSpacePublisherModule;
...
+
+		static const char *id(void);
+		virtual void init(void);
+
+		void atomAddSignal(Handle h);
+		void atomRemoveSignal(AtomPtr atom);
+		void AVChangedSignal(const Handle& h,
+							 const AttentionValuePtr& av_old,
+							 const AttentionValuePtr& av_new);
+		void TVChangedSignal(const Handle& h,
```
linas: The indentation here looks incorrect. I think you want to use tabs, three, I guess, to line up with `void AVChangedSignal(const Handle& h,` and then use spaces the rest of the way, so that `const AttentionValuePtr& av_old,` lines up correctly, no matter what size the tabs are.

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
```diff
@@ -81,91 +81,91 @@ class AtomSpacePublisherModule;
...
+							 const TruthValuePtr& tv_old,
+							 const TruthValuePtr& tv_new);
+		void addAFSignal(const Handle& h,
+						 const AttentionValuePtr& av_old,
+						 const AttentionValuePtr& av_new);
+		void removeAFSignal(const Handle& h,
+						 const AttentionValuePtr& av_old,
+						 const AttentionValuePtr& av_new);
```
linas: Oh, and also, it would be great if this could be designed in some way to not use signals, especially the atomspace add/remove signals. The problem is that these signals run synchronously, in the thread adding/removing the atom.  This slows down performance, which is already poor. I don't yet know what a good substitute would be, but its something to think about, in the back of one's mind.

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
Dagiopia: that's what sucks about using tabs over spaces... changing tabstop results in inconsistent view. I was using tabstop=4 on my vim and it looked well aligned until the github viewer showed it with 8 spaces then it's ugly.

Dagiopia:
I haven't touched the overall design that existed. Only changed the library to one that is easily available and added a check for the atom type so that get_name isn't called on a link type.

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
ngeiswei: @Dagiopia, please see http://vim.wikia.com/wiki/Indent_with_tabs,_align_with_spaces

ngeiswei: My recommendation is that you set tabstap to 3 or 5, that way if you stumble on a space indented file (that still concerns a third of the code), you'll immediately catch it.

ngeiswei:
Oh, it's already been merged...

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
Dagiopia: pushed the changes on #3311 

thanks @linas and @ngeiswei

opencog/cogserver/modules/events/AtomSpacePublisherModule.h
Dagiopia: Replacing the signals seems like it would be a whole lot of change. 
But since the signals are basically callbacks to a user-defined function, the performance of the user-defined function ultimately affects the performance of the atomspace (while waiting for the user function to finish execution) 
it should be something that is called and the caller can just go on with the rest of its activities. 
filling a queue like zmq does.

PR #3323, opened 2018-10-11
fix inference-control-learning example

noskill:
rename cog-tv-conf -> cog-tv-confidence in inference-control-learning


ngeiswei:
@noskill, I'm closing because in the up-to-date version that lives in `singnet` it is already corrected.

I would love to update `opencog/opencog` with it, but I'd rather wait till the conflicts between `opencog` and `singnet` get resolved (hopefully after a couple weeks).

PR #3332, opened 2018-11-02
Lojban - Lot's of improvments

rTreutlein:


bgoertzel:
wow!

opencog/nlp/lojban/HaskellLib/src/OpenCog/Lojban.hs
```diff
@@ -46,6 +46,8 @@ lojbanToAtomese rstate seed text = (fmap wrapAtom) . fst <$> evalRWST (apply loj
                         ,sNow = now_here
                         ,sCtx = [now_here]
                         ,sJAI = Nothing
+						,sDA  = id
+						,sDaM = M.empty
                         ,sXU = []}
```
linas: whitespace!

opencog/nlp/lojban/HaskellLib/src/OpenCog/Lojban.hs
```diff
@@ -63,6 +65,8 @@ atomeseToLojban rstate seed a@(LL [_an,s]) = sText . fst <$> execRWST (unapply l
...
```
linas: whitespace!

opencog/nlp/lojban/HaskellLib/src/OpenCog/Lojban/Syntax/Util.hs
```diff
@@ -273,6 +291,11 @@ getFlagValueIso flag = Iso f g where
 setFlagValueIso :: Flag -> SynIso String ()
 setFlagValueIso flag = inverse (getFlagValueIso flag)
 
+setFlagValueIso2 :: Flag -> String -> SynIso a a
+setFlagValueIso2 flag val = Iso f g
+    where f a = setFlagValue flag val >> pure a
+          g a = rmFlag flag >> pure a
+
```
linas: So almost all of your code seems not to contain any comments that explain what it actually does. I don't want to be a pill and explain why writing comments for code is good.  For example, it helps future-you remember what something does. and future-others. And, rarely, having to explain something in comments helps you realize that there's a bettter design.

linas:
Hey Roman, this is a weird question, but I have to ask...

The above pull req shows twenty commits. However, when I do a `git log`, I see only a single commit, and none of the twenty.  If I look at the commits, above, I see, for example, https://github.com/opencog/opencog/pull/3332/commits/9e9b8785b47728cdd8a7649af29bb3c8690175a7   but this number never actually appears in the `git log`. 

If I do a `git diff accec5509a5f160e5e97cb6da5f959cd8f886ee2 1832571ffda017099a9507d54a52d6d1c33872a6`  then I see the changes in this pull request, instead of what I should normally see: all of the other changes that everyone else made, from April 5th (your first commit) until now, excluding your changes. 

Any clue how this rather bizarre state of affairs arose? This makes me nervous: this is not how git normally works.  Is this something you did? Something that Microsoft did? How can this happen?

@ngeiswei Any clue what Roman did here? Is this something to be concerned about?

rTreutlein:
@linas GitHub selected "Squash and Merge" as the default option. I wasn't quite sure if that's the right thing to do but in the end, I just went with that. But I guess from your confusion that was the wrong choice.

If you want I can try to reverse that an apply a normal merge.

linas:
Oh, that's .. different. I've never seen a "squash and merge" menu entry/button/selection anywhere on github before ... I just now crawled through account settings, and see nothing like that either.  I'm looking now ...  found it. Been using github for 10 years, never saw that button before. Wow.

Squashing: There are two minor(?) problems with it: if one is bisecting to find a bug, its much easier to bisect many small patches, than to have to deal with one big one, because you don't know where the bug is, in a big patch.  However, I bisect very infrequently, and certainly not haskell code, so not really a problem.

Other minor problem: summary total of commits. For example: `git log |grep Treut |wc` says 170 commits. Without the squash, it would be 185.  But there are other issues:  https://github.com/opencog/opencog/graphs/contributors  shows that you have four commits, and not 170...  

I guess I don't really care; except that I monitor user activity to see who is doing what, just to keep track of things.  The squash-merge makes you kind-of disappear from the radar of activity.

rTreutlein:
@linas I figured out why GitHub only showed 4 commits from me. As this started as my personal account a long time ago I had only added my personal email address to GitHub. And since GitHub only attributes commits to you that have the same Email it didn't count all the work where I used my hansonrobotis or now singularity.net Emails to configure git.
That's fixed now but it still only shows 93 commits. Not quite sure were that discrepancy comes from I don't think I have done that money squash&merges. At least the additions/deletions are the right numbers.

rTreutlein:
@linas Okay, some more checking your command `git log |grep Treut |wc` was counting some things that should have been filtered. If you do `git log | grep Treut | sort -u` then you can see what. Only occurs starting with Author: should have been counted. Which brings the number of my commits down to 139. And if I did squash and commits for some bigger Pull Requests then this seems plausible.
Going to avoid doing that from now on. But I think additions/deletions are more representatives, though I guess it all depends. I tend to pack too many changes into 1 commit which I am trying to get better at but it leads to commits that might contain the work of 5 commits of a reasonable size.

linas:
I'm fairly sure that github does NOT count merge-commits; it only counts those that are not merges. I think. Not entirely sure.

Re; the desirable size of a commit. Up to you, but .. let me explain my workflow. I break up "big ideas" into small ideas, that are typically 1-5 method calls, 1 or 2 classes, 1 or 3 wrappers, whatever. Code those as a unit in about 5-15 minutes. Compile. (Maybe/usually) run unit tests. Works? great! Now `git diff` and perform a code review. Did I forget something? Some obvious typo? Did I really mean to do it like that? Looks good? Commit. Then do it again. 

Thus all of my commits tend to be 10-50 lines of code changed, never more. Why? If its bigger, then the git diff is too big, too hard to review, has too many complex moving parts. 

Here's another reason for small commits: I often start doing something, and realize its a bad idea. Throw it away. This is really easy: `git checkout file` but only  if I want to throw away EVERYTHING in that file. As long as I commit frequently, and all changes are small changes, then throwing away everything is easy, cheap, because its small. But if the file has a dozen changes in it, not yet committed, and I want to remove just one or two ... yuck. Backtracking is hard. Have to remove by hand, with an editor, which is error-prone, and likely to screw something else up in the file. Dangerous, painful, time-consuming. Must avoid.

So: for me, a commit is a fire-wall -- it protects everything that came before from whatever bad idea I'm going to have at this moment. It provides easy undo. But this works only if you have lots of small commits. Which is why I have .. lots of small commits.

ngeiswei:
I would not recommend squashing-and-merge (didn't know it existed either), for the same reason Linas highlighted.

For branches that go very far in the past (few months), I would recommend rebasing onto the master *before* creating the PR (rather than using github option rebase-and-merge), to fix possible conflicts (some could be indirect and not detectable by git, BTW), as well as mitigate parallel git histories (though this one is less important).

Regarding commit size, as Linas said, they should be self-contained, each capturing an idea, or sub-idea. Ideally the code should be able to compile after each commit, convenient for bisecting. As perhaps Linus Torvalds might have said, the more intricate the idea the shorter the commit should be, see Section 6 of https://www.kernel.org/doc/html/v4.10/process/coding-style.html where I've substituted "function" by "commit"

" The maximum length of a [commit] is inversely proportional to the complexity of that [commit]."

Note that `git -p add` (or likely your IDE git front) allows you to craft your commit with surgical precision.

PR #3339, opened 2018-11-22
Complete inference control experiment

ngeiswei:
DON'T MERGE!


ngeiswei:
This is a PR test to understand what is gonna be the best path to merge `singnet/opencog` back to `opencog/opencog`.

Obviously naively creating a PR out of the head of that branch brings with it all the preceding commits not merged yet into `opencog/opencog`.

So I see 2 options

1. I cherry pick my commits, create a PR with that, merge, done (then redo that with all my work living on singnet)
2. Wait for the authors to merge their work. I can do that for obvious stuff like circle config files. But I cannot do that for less trivial stuff like NLP which likely already contain conflicts.

PR #3340, opened 2018-11-22
Merge initial singnet circle config files

ngeiswei:


ngeiswei:
This seems completely harmless so I'm merging. Obviously it will not affect the `opencog` repo since circle ci is not configured there.

linas:
Is there a reason circleci is not configured? We used to run it many years ago, I liked it a lot, but then it got broken, and was stripped out. It would be cool if it could be made to work again, or if its still borken, then some replacement for it could be found. 

Very recent discussions about code stability and process clearly shows that we need this stuff again, that we lost a lot of good stuff, when circleci broke, and the build servers broke. We really should get these working again, and now that the teams are bigger, this is getting to be more important.

linas:
Well, but circleci had some major train-wreck some years ago, which is why it got turned off for opencog. Is it now fixed and operational?

ngeiswei:
It's slow (10 to 15m to compile check a single repo) but it works.

linas:
Oh; Hmm. Why would speed matter? 

One of the issues with Jenkins or CircleCI, I forget which, was a fu'ed up config which prevented a merge until the compile is done. That was a freakin nightmare and I have no desire to revisit that nightmare.

The point being: Tech should always make it easier for the programmer to do their work, and never harder. Complexity always generates bugs, makes it harder for the programmer to understand the system, and damages programmer productivity.  It's moronic to make programmers miserable.  

If the states and history queue is running behind that's OK; as long as the status gets updated in 24-48 hours, I don't care. If it's blocking some action as a result, that's just plain wrong.    (15 minutes is mega- faaaast; in large corps, there's usually a 1-2 day backlog sometimes hitting a week. It gets annoying when its that long, but its not fatal. The circleci/jenkins config was accidentally blocked on this queue, and I had a very very very long hostile discussion about this, that was quite unpleasant. Lets not go there again.)

ngeiswei:
I'm fine having unblocking CI. That can easily be disabled in the repo settings (you probably have the admin rights to control that BTW).

The idea I believe would be to have unblocking CI on the opencog repos, and blocking CI on the singnet repos. At least that makes sense to me, opencog would be for development, singnet for production (at least that's the way I see it).

linas:
Yes, that makes sense.

linas:
@ferrouswheel I'm guessing that you never actually hit the problem I hit, probably because you have a .. rather different development process.  The normal development process is to make many small commits, and do frequent merges.  Having to wait even one or two minutes for a merge is just outright unacceptable; having to wait 10-15 minutes is outrageous. It defeats the whole point of the development process!  No sane software development shop would use a process like that.

However, this logic is reversed when dealing with stable branches: the stable branches should pass unit tests, conformance tests and system tests. Unit tests typically take 20 minutes-hour to pass (they are very fast on opencog, because opencog is a rather tiny project) (and opencog does not have conformance or system tests; these typically take 1-2 days, and 1-4 weeks, respectively, when they do exist)  So yes, for release  candidates, the conformance tests should pass. For releases, the system tests should pass. However, for development, its just insane to delay anything by longer than it takes a browser window to refresh.

The point of having an automated build system is to avoid the pain of bisecting, when the build is broken. You can look at teh build history, see when the blinker turned from green to red, see which merge did the breakage, and then torture that developer until they fix it, and/or revert that merge, if they don't fix it.  It's a very useful tool for supervising, managing and debugging broken masters.  Make it a useful tool, people will use it. Make it into a painful tool, and people will do end-runs around it, find ways to avoid, skip that step.   The deployment of CI a few years ago turned development into a world of pain.   A rational, reasonable development process should make development smooth, easy, painless.

vsbogd:
Usually review takes more time than building the CI and it also happens before merging to master it is the reason I also don't think CI time is not such a big issue. 

I worked in setup where large work is being done in dev branches, merging each small change requires green CI build and CI runs only unit tests. It is fast enough (10 mins) to be faster than most of reviews. To merge whole large dev branch to the master functional tests should be green. Running functional tests takes more significant amount of time: about 1 day to run them all. And this system allows dev team releasing new version each week.

For opencog as many people work on relatively compact set of components it could be merging to the master requires review and green unit tests. Releasing the master and making release branch requires green functional tests. If patches are needed for the stable branch they can be merged to stable branch and master in parallel.

linas:
What would happen a lot, de-facto, in opencog, -- and I mean, happen more than just a few times, but lots of times, was that people would change something, break a unit test, disable the unit test, get a green circle-CI build, not wait for review, and just merge. There were 3 or 4 developers who would do this *all the time*.

Since the circle-CI was always green, you have no clue that in fact the system is actually broken. Typically, you would not discover the break until half-a-year later, during some difficult debug session.

There are two solutions:  One is: don't give everyone write permission. I have argued strongly for this, and I have always lost. and these have always been loud, angry and hurtful arguments.  I still think it is completely wrong to give the most junior and inexperienced programmer write permissions. But it's political, who is a friend of who type situation.

If everyone is allowed to break anything for no reason, then Option two is to allow broken unit tests. This removes the incentive to cheat. This also results in an obvious red light, so you can immediately see who cheated, and what they did.

I don't see a third solution. Given the laissez-faire attitude, that anyone can do whatever they feel like for no reason and without review, management, oversight or supervision, then I think that option two is much better.

vsbogd:
In case which I described above the solution was PR merging policy. PR cannot be merged without green CI and approve from someone else (PR author could not approve it himself). And, yes, nobody had rights to merge to the master skipping the policy.

linas:
That would be a reasonable policy, I suppose; but, again, when I tried to implement a similar policy, it was met extremely heavy resistance. Several years ago, I threatened to quit, literally, and that did not work to get such a policy enacted. If you recall, I brought up this issue about a month ago, and we had almost exactly the same argument, and I was told to shut up and stop arguing. Yes, what you suggest is a reasonable policy. Unfortunately, there are some people who are against it.  I don't know how to solve this problem.

PR #3343, opened 2018-11-22
Complete inference control experiment 3

ngeiswei:
DON'T MERGE!


ngeiswei:
Damn, git isn't able to tell that my commits have already been merged. I was hoping for a nice
```
This merge commit was added into this branch cleanly.
```
but instead we have two duplicates
https://github.com/opencog/opencog/pull/3343/commits/929655989f93e567f12de9b56fc8c1f40e4a51eb from singnet
https://github.com/opencog/opencog/commit/4e23033f1685118bdca0fbb968327504fb55d3cc from opencog

Maybe if I would merge git would somehow solve it at this stage but I don't think so. I'm gonna try locally but that doesn't look good...

ngeiswei:
Actually locally it works fine! My commits are not duplicated.

I suppose the reason is because they are conflicts and so github didn't carry out the merging (or completely) at this point and thus could tell that these commits are redundant.

I suppose that if I merge upstream/master into that fake branch, my commits are gonna disappear from the PR, let's try.

ngeiswei:
Wait! My commits have not disappeared on github and have not disappeared on my local PR. They are duplicated on that branch. That makes me a bit nervous. I'll locally merged that PR into my master to see how it goes.

ngeiswei:
Nope, git was unable to detect the redundant commits. :-(

Yet when I locally merged that PR to master, prior to merging master into it, it did detect the redundant commits.

I don't know git well enough to understand why it fails in that case. But the moral of the experiment is that we have to be extremely careful with merging stuff in parallel (just as I had feared).

Now maybe it's because I know how to use git well enough, but in the mean time, since we have no choice, for that PR at least. Please, @leungmanhin and @glicerico cherry picks your commits of that PR to create a new each a new PR. Then for the rest we'll see, maybe can try to do a batch PR, or cherry picking. Obviously we should absolutely avoid merging that PR (we didn't want it in the first place anyway), we should absolutely avoid merging batches (I'm whole singnet to opencog, or while opencog to singnet) because we have started merging PRs to opencog and singnet in parallel, and apparently git may have our back.

ngeiswei:
Sorry let me rephrase that
> we should absolutely avoid merging batches

We should be extremely careful to not introduce redundant commits (trying locally first before pressing the merging button, that sort of things).

ngeiswei:
Obviously if we introduce redundant commits it's not the end of the world, just some mess in the git history, I just hate that sort of mess.

ngeiswei:
Hmm, one can probably avoid these duplicates by playing with rebasing or something. Anyway, late, I'll get to the bottom of this tomorrow. I refuse to live in fear!

vsbogd:
@ngeiswei, have you tried to reproduce it using two small experimental branches?

ngeiswei:
@vsbogd no, I will try tomorrow if I have too. Anything it takes to be a free man!

vsbogd:
I would like to understand the scenario to reproduce. At the moment my understanding it is:
- create fork of the singnet
- make branch from singnet/master
- add commit
- merge branch to singnet/master
- merge branch to opencog/master
- merge opencog/master to singnet/master

Is it correct?

ngeiswei:
I could be, I guess.

vsbogd:
Unfortunately it is not that simple. I have tried the scenario on test repository and cannot reproduce commit duplication.

ngeiswei:
@vsbogd it might have something to do with the fact some of the duplicates required resolutions that I provided different ones during the 2 mergings. Not sure of course, just an hypothesis.

linas:
I haven't thought very hard about this, but I don't see any particular way out of this pickle, except to cherry-pick commits one by one.

In the long run, I would be a lot happier is you picked a process that is simpler, easier and maintainable. 
You're making things hard on yourself, and also damaging code quality and stability.  A simpler, better process will lead to higher-quality code, fewer bugs, and fewer maintenance headaches.

linas:
Also: side note: I do not want to see any of the ull code in the opencog repo. It would need a major redesign to become acceptable.   Its OK for a sandbox, but it's not production quality.

PR #3344, opened 2018-11-23
Complete inference control experiment 4

ngeiswei:
Identical to https://github.com/opencog/opencog/pull/3343 but it has been rebased into upstream/master before creating the PR.

What I did is just skip the duplicates while rebasing. So the duplicates are gone, great. But it's not all that glorious, there is basically no way to tell for sure if a commit is a duplicate while rebasing, that is because there is no way (or I guess it's generally hard) to tell if a duplicate is "legit" or not. For instance consider the following code on master:
```c++
call_function_A();
```

And the following commits in a feature branch:

commit B:
```c++
call_function_B();
```
commit C:
```c++
call_function_A();
```

If you rebase the feature branch onto the master no commit is empty, and there is no way git can know by looking them individually that the whole sequence is idempotent. Even if it is idempotent, maybe the user wanted to take that path. I guess the best git could do is look for that commit in the git history and warn the user an identical one.

I'll dig a bit deeper into what git can and cannot do, but so far the moral of the story is pretty clear:

*Be careful*


ngeiswei:
I tried to see if one can automate finding duplicate commits based on

https://stackoverflow.com/questions/11605703/git-finding-duplicate-commits-by-patch-id

but even this doesn't reliably work. Duplicates are not always identical either because they may require conflict resolution and depending on who resolves and on top of what commits the results may differ...

So, anyway, it seems there is no easy way out.

For the opencog repo I think it's best we all cherry pick our work, making sure we don't introduce duplicates.

Once we're done cleaning this, well, working on parallel branches obviously add extra inconveniences, we should reconsider that IMO, and if we keep doing that we need to be disciplined. As I said either we always double PR (one `opencog` and `singnet`) or we always sync opencog<->singnet by batches, not a mixture of both.

linas:
OK, please do not bother to merge the ULL code. It has half-a-dozen issues with it; its borken, and is not in a state that is acceptable for merging into opencog. One obvious issue with this pull req is that it duplicates existing files, e.g. the non-breaking prefixes, which already exist in other directories. It also modifies some exiting files in such a way that bugs are introduced.

The ULL code is not acceptable for merge, without significant restructuring and redesign.

PR #3345, opened 2018-11-23
Remove useless warning from miner

ngeiswei:
From https://github.com/singnet/opencog/pull/24


ngeiswei:
As well as optimize and simplify the pattern miner code.

PR #3362, opened 2018-12-06
Make TimeOctomap work with Values WIP  

misgeatgit:
This PR depends on https://github.com/opencog/atomspace/pull/1937. Here is an example how it works:
```
(use-modules (opencog octomap))

(define znode_0 (Node "zitem_0"))

; One octomap object to store locations of items
(define zmap (OctomapNode "zmap"))

; Insert items and their location to the octomap
(define zvalue_0 (OctoValue  (ListLink znode_0 zmap) 1.12 2.34 3.564))

(define key (PredicateNode "spacetime") )

; setting values
(cog-set-value! znode_0 key zvalue_0)

; getting values
(cog-value znode_0 key)

; list ref
(define v (cog-value znode_0 key))
(cog-value->list v)
(cog-value->ref  v 1)
```
Next step will be  modifying the pointmemory API to make it work with OctoValue and some unit tests.


linas:
Re this: `(define zvalue_0 (OctoValue (ListLink znode_0 zmap) 1.12 2.34 3.564))` in the example. -- The three floating point numbers are presumably the 3D coordinates. But I don't think that they would be needed, here. The actual 3D coordinates are going to come from the octomap itself; I see no reason to pass hem into the constructor.

In fact, it would be kind-of-a-pain to pass them into the constructor, since you might not actually know what they are, when building the OctoValue for the first time. The actual x,y,z location comes later, from some ROS node somewhere.

opencog/spacetime/octomap/OctoMapNode.cc
```diff
@@ -0,0 +1,73 @@
...
+
+Handle OctoMapNode::factory(const Handle& base)
+{
+    if (OctoMapNodeCast(base)) return base;
+    Handle h(createOctoMapNode(base->get_name()));
+    return h;
+}
+
+/* This runs when the shared lib is loaded. */
+static __attribute__ ((constructor)) void init(void)
```
linas: Please use `DEFINE_VALUE_FACTORY` for this. See `FloatValue.cc` for an example.

linas: whoops, wrong comment, deleting.

opencog/spacetime/octomap/OctoMapNode.cc
```diff
@@ -0,0 +1,73 @@
...
+{
+    classserver().addFactory(OCTOMAP_NODE, &OctoMapNode::factory);
+}
```
linas: Please use `DEFINE_NODE_FACTORY` for this. See `NumberNode.h` for an example. see `ClassServer.h` for the definition.

opencog/spacetime/octomap/OctoMapValue.h
```diff
@@ -0,0 +1,71 @@
...
+static inline OctoValuePtr OctoValueCast(const ValuePtr& a)
+{
+    return std::dynamic_pointer_cast<const OctoValue>(a);
+}
+
+static inline std::shared_ptr<OctoValue> createOctoValue(const HandleSeq& hseq,
+        const std::vector<double> values)
+{
+    return std::make_shared<OctoValue>(hseq, values);
+}
```
linas: Per other comments, its probably best if the constructor doesn't require initlial float values.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,69 @@
...
+        snprintf(buf, 40, "%.17g", v);
+        rv += std::string(" ") + buf;
+    }
+    rv += ")\n)\n";
+
+    return rv;
+}
+
+// Adds factory when the library is loaded.
+static __attribute__ ((constructor)) void init(void)
```
linas: Please use `DEFINE_VALUE_FACTORY` for this; see `FloatValue.cc` for example.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,69 @@
...
+
+    return false;
+}
+
+std::string OctoValue::to_string(const std::string& indent) const
+{
+    std::string rv = indent + "(" + nameserver().getTypeName(_type);
+    rv += ("\n " + _item->to_short_string() + " ");
+    rv += (_octo_atom->to_short_string() + " (");
+    for (double v : _value) {
```
linas: You probably want to call update, first, as otherwise, you will probably print out some old, stale value...

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,69 @@
...
+#include <opencog/atoms/value/ValueFactory.h>
+
+using namespace opencog;
+
+
+void OctoValue::update()
+{
+    //point3d p = (_om->get_locations_of_atom(_octo_atom)).front();
+    //_value = {p.x(), p.y(), p.z()};
+    _om->insert_atom(point3d(_value[0], _value[1], _value[2]), _item);
```
linas: OK, well, this is clearly wrong -- it needs to be the other way around: when `update` is called, it should ask the octomap what the correct values should be. That is, it should look something like this:
```
void OctoValue::update() {
     _value = _om->get_current_location(_item);
}
```

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,69 @@
...
+                     std::vector<double> values) : FloatValue(OCTO_VALUE, values)
+{
+
+    if( hseq.size() != 2) {
+        throw InvalidParamException(TRACE_INFO, 
+                "Required 2 arguments. Provided %d.", hseq.size());
+    }
+
+    _item = hseq[0];
+    _octo_atom = hseq[1];
```
linas: Could you rename `_octo_atom` to `_octomap_node`?  This would be a little less confusing, because right now, everything is octo-this and octo-that, and its hard to keep in mind what is what.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,69 @@
...
+    _om = OctoMapNodeCast(_octo_atom)->get_map();
+
```
linas: and, here, of course, you need to say something like
```
   _om->insert_atom(_item);
```
so that the octomap knows it needs to start tracking the location of `_item`.  That is the whole point of `OctoValue` -- to force the octompa to track the atom, and to ask for the updated location, whenever it is needed.

opencog/spacetime/octomap/OctoMapValue.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace opencog
+{
+/** \addtogroup grp_atomspace
+ *  @{
+ */
+
+// OctoValue holds 3d coordinates of an atom. When it is
+// set, it stores in the TimeOctoMap instance provided by
+// OctoMapNode.
```
linas: This comment is wrong. The OctoValue does NOT hold the 3d coordinates! The OctoMap holds the 3D coordinates. The OctoValue provides a way to obtain the current 3D coordinates from the OctoMap.

opencog/spacetime/octomap/OctoMapValue.cc
misgeatgit: So, you are suggesting OctoValue should only be used as a getter? and whenever **cog-set-value!** is called, all it does is get latest location from the octomap and update the _value vector object?  That makes sense to me.

linas: Yes, it is only a "getter". In fact, `atomtypes.script` should probably have `OCTO_VALUE <- STREAM_VALUE`  -- the idea here is that stream values are values that "automatically" change, under the covers, and the atomspace doesn't have to do any work when they change. Examples would be audio or video streams -- so the for example, a stream would store a URL for a webcam or a microphone; it wouldn't actually store the video frames at any given moment; it would just return them, but only if someone asked for them.

See RandomStream.cc as a (really super-simple) example of  stream. It returns a different random number every time you ask for its value.

So, for all streams, the `cog-set-value!` is done only once, to attach the stream to the atom.

linas: Oh, and since the constructor does `_om->insert_atom(_item);` the destructor needs to do a `_om->remove_atom(_item);`

Hopefully, it will never happen that there are two different values tracking the same atom in the same octomap. That would be weird and buggy.  So lets just ignore that for now?

linas: And part of the idea of using streams in atoms is that the atoms could maybe represent how stream values are hooked together. So maybe there might be some "output stream" that could be a phone or tablet, and to send the video from webcam to the phone, you would attach `(ConceptNode "my webcam")` to `(ConceptNode "that tablet")`  or something like that. Details TBD, no one has ever done this; I haven't thought this through very much.

For the octomap, we would need something like
```
EvaluationLink
     GroundedPredicateNode "scm:is-to-the-left-of"
     ListLink
           Concept "Visible Person 42"
           Concept "visible person 69"
```
and "scm:is-to-the-left-of" would dig out the OctoMapValue for both, compare x,y,z, and return true or false. But it would do this only at the time that the evaluationlink is actually evaluated.

linas: And just two more short comments to clarify: 

> `OctoValue` should only be used as a getter? and whenever `cog-set-value!`

You can add setter-methods to the c++ `OctoValue` class. I don't care. It might be an OK way to update the contents of the octomap.  How do you plan to get ROS data from ROS to the octomap?

>  and whenever `cog-set-value!`

Please note that `cog-set-value!` only changes the value on the atom. It does NOT change the value itself! You cant to avoid calling this, as much as possible, because it is a performance killer. Using it requires you to run the constructor for a value, malloc memory, update the KVP database in the atom, and then run the destructor for the old value. This would be an extremely bad idea, if the only goal is to change three floats.

Dagiopia: the entire point of having values is to be able to change them as frequently as possible right? is there any way in which the cog-set-value! function could be written where changing a value simply comes down to just indexing the vector and changing the value right there? or maybe implementing the virtual function `update()` ? 
also maybe `cog-set-value!` shouldn't be able to create new values at all... like `set!` if the variable isn't already defined, it should give out an error. It only calls the update function of each value type to change the actual value and `cog-new-value` could create new values taking the key pair atoms (because what cog-new-value does is basically the same thing as calling the functions FloatValue StringValue....)

Dagiopia: on `cog-set-value!` not creating new values(objects) I see that a call to it (for instance on a FloatValue) means a call to `createFloatValue` to attach to the key pair atoms which essentially creates a new FloatValue object (thereby a new vector of floats with the new value) 
Do you think it's possible to use the object's referece to the vector of floats in order to change a specific value from 0.2 to 0.4 without allocating a new memory space for the new FloatValue object? 

I see that the current implementation of values is much much faster than atoms when manipulating them but if there is a way to avoid the above scenario, it would be as fast as it can be.

linas: Well, `cog-set-value!`, the scheme code, does have to convert a list of SCM smobs into an instance of a C++ `class Value`, so, yes, it calls the factory to call `createValue` to do this.  After this, it calls the c++ code `Atom::set_value()` which takes a `ValuePtr`, and so that had to come from somewhere.

The SCM smobs are opaque data types managed by guile, they are not floats or doubles or vectors; they're a bucket of bitflags and weird encodings, including "boxes" (read about "boxing" in guile) and other strange stuff. 

The Atomspace Values themselves are designed to be "readonly" (immutable) so as to avoid multi-threading race conditions. If you made them read-write, then you would have to take a lock every time you read them, which would make them less efficient.  My goal here was to write lock-free code. 

I haven't read the following, but I assume it contains valuable explanations: https://softwareengineering.stackexchange.com/questions/171253/does-immutability-entirely-eliminate-the-need-for-locks-in-multi-processor-progr

The general background:  https://en.wikipedia.org/wiki/Immutable_object

linas:
@Dagiopia I wanted to say: thank you for thinking about what the best implementation might be. I kind-of wish everyone did that. Finding the best implementation is hard: there are usually 3 or 4 or 5 different ways of doing it, and its hard to see which is "best" (simplest, smallest, fastest, easiest-to-use). The usual guides, of existing use-cases or existing systems to measure and compare, are absent, so its hard to guess at what "best" is. So you're questions are good. I gave quick answers, but that's only because I got to think hard about this,  over the last few years.

opencog/spacetime/octomap/OctoMapValue.cc
linas: The OctoValue should not track. Instead, the `update()` function should call `TimeOctoMap` to obtain the current position.  Basically, OctoValue should use zero CPU time, unless someone is actually calling the `get_value()` method.

linas: The original comment, stating that ` _om->insert_atom(_item);` still seems to not have been done (or, if its not needed, I don't see why)

linas: @misgeatgit yes, the implementation of `update()` now looks correct.

misgeatgit: Insertion is going to be made by using pointmemory library (https://github.com/opencog/opencog/tree/master/opencog/spacetime/pointmemory). So, the next step will be modifying pointmemory API to return Values instead of Handles. That is what I was thinking. As you said, the other way would be, also doing the insertion using OctoValue. But I am not sure how we can reuse the functionalities provided in pointmemory if we go that way.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,58 @@
...
+{
+
+    if( hseq.size() != 2) {
+        throw InvalidParamException(TRACE_INFO, 
+                "Required 2 arguments. Provided %d.", hseq.size());
+    }
+
+    _item = hseq[0];
+    _octo_node = hseq[1];
+    _om = OctoMapNodeCast(_octo_node)->get_map();
```
linas: I'm fairly certain you want to have `_om->insert_atom(_item);` here.  Maybe you are hoping that "someone else" will do the insert, but that is a poor thing to hope for, because "someone else" might forget.  In that case, you end up with the confusing situation where the octomap isn't actually tracking the node that you are given, here.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -0,0 +1,58 @@
...
+using namespace opencog;
+
+
+void OctoValue::update() const
+{
+    auto it = (_om->get_locations_of_atom(_item)).rbegin(); //We need the last.
+    _value = {it->x(), it->y(), it->z()};
+}
+
+OctoValue::OctoValue(const HandleSeq& hseq) : FloatValue(OCTO_VALUE)
```
linas: This is maybe nit-picking, but it would probably be better if this took two `Handle`s, instead of a `HandleSeq`. The reason for this is to minimize API complexity: If all code everywhere always uses Handles for everything .. then we're done. But if all code everywhere has to support both Handles and handleSeq and HandleSeqSeq, and HandleFooMap and ... well, that just ends up being more code, more complexity, more bugs, and everything is harder to understand.  So its best to avoid HandleSeq most of the time.  This seems like a good example where its not really needed.

linas:
OK, I marked everything resolved, except for one old comment, and two new comments. Hit browser "refresh" to see the easy-to-read version of the page.

opencog/spacetime/octomap/OctoMapValue.cc
linas: never mind. I see that the octomap API doesn't allow that.

opencog/spacetime/octomap/OctoMapValue.cc
linas: Yes, that's right, about the insert.  I'd forgotten the pointmemeory API.

Yes, PointMemory needs to be converted to use mostly FloatValues for most things, and maybe OctoValue in one or two places.  That conversion is going to be icky ... it won't always be obvious what to convert, when.   Convert the xyz first, make that a pull request, and then do the rest in a different pull req. That should make things simpler to review and test.

linas:
OK, how abot this: do the one remaining `DEFINE_NODE_FACTORY ` change, and thenn I'll merge, and we can handle everything else in other pull reqs.

opencog/spacetime/octomap/atom_types_init.cc
```diff
@@ -0,0 +1,6 @@
+#include "opencog/spacetime/octomap/atom_types.definitions"
+
+#define INHERITANCE_FILE "opencog/spacetime/octomap/atom_types.inheritance"
+#define INITNAME octomap_types_init
+
+#include <opencog/atoms/proto/atom_types.cc>
```
linas: this is now `<opencog/atoms/atom_types/atom_types.cc>`

misgeatgit: Oh, I thought I was using the latest atomspace. Let me fix that.

opencog/spacetime/octomap/OctoMapNode.cc
```diff
@@ -0,0 +1,69 @@
...
+{
+}
+
+/*Add factory.*/
+DEFINE_NODE_FACTORY(OctoMapNode, OCTOMAP_NODE)
+
+/* This allows guile to load this shared library */
+extern "C" {
+    void opencog_spacetime_octomap_init(void) {}
+};
```
linas: You don't actually need this, because there's no scheme for any of this. At any rate, it wouldn't go here.  Just remove this, and remove the call to it.

opencog/spacetime/octomap/octomap.scm
```diff
@@ -0,0 +1,8 @@
+(define-module (opencog octomap))
+
+(use-modules (srfi srfi-1) (opencog) (opencog exec))
+
+; Load the C library that calls the nameserver to load the types.
+(load-extension "libatom_octomap" "opencog_spacetime_octomap_init")
```
linas: and you don't need this call to `opencog_spacetime_octomap_init` (because it doesn't do anything)

linas:
and circleci is failing with a weird error I don't understand .. does this build for you?

misgeatgit:
Yes, it builds for me.

opencog/spacetime/octomap/OctoMapValue.h
```diff
@@ -0,0 +1,69 @@
+#ifndef  _OPENCOG_OCTOMAP_VALUE_H_
+#define  _OPENCOG_OCTOMAP_VALUE_H_
+
+
+#include <opencog/atoms/value/Value.h>
+#include <opencog/atoms/value/FloatValue.h>
+#include <opencog/atoms/value/atom_types.h>
```
linas: and this one, and the one below: `opencog/atoms/atom_types/atom_types.h`

linas: oh, there is no second one. Hmm.

opencog/spacetime/octomap/OctoMapNode.cc
```diff
@@ -0,0 +1,64 @@
...
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this program; if not, write to:
+ * Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+
+#include <mutex>
+
+#include <opencog/atoms/value/NameServer.h>
```
linas: and this one: `opencog/atoms/atom_types/NameServer.h`

Sorry for all the trouble. This really should be the last of it.   We are going to move truth values a bit later, but after that, there will be no more big code reorgs that affect everybody.

linas:
phew

PR #3369, opened 2018-12-11
Quick hack to fix #1944

linas:
hack around a long-failing unit test.


linas:
I'm just going to merge this; this has been a blocking bug for far too long. If you want a different design .. please implement it.

PR #3371, opened 2018-12-11
Add a test for ghost pln demo

AmeBel:


tests/ghost/test-ghost-procedures.scm
```diff
@@ -0,0 +1,66 @@
...
+(define gpt "ghost-procedure-tests")
+
+;------------------------------------------------------------------------------
+; Setup
+
+(use-modules
+  (opencog)
+  (opencog nlp)
+  (opencog nlp chatbot)
+  (opencog nlp relex2logic)
```
ngeiswei: Is `(opencog ghost)` missing? If I don't have it, I get
```
Unbound variable: perceive-word-hook
```
and then `nlp-parse` is not defined.

ngeiswei: After importing `(opencog ghost)`, `nlp-parse` works, however then I hit
```
opencog/ghost/procedures/pln-trail-3.scm:105:15: In procedure configure-pln-rbs-3:
In procedure module-lookup: Unbound variable: pln-load-rules
```
If I add modules `(opencog rule-engine)` and `(opencog pln)`, I get passed that, however then I hit
```
opencog/ghost/procedures/pln-reasoner.scm:88:8: In procedure pln->sureal:
In procedure module-lookup: Unbound variable: filter-for-sureal
```
If I add `(opencog sureal)`, I get passed that, however then I hit
```
opencog/ghost/procedures/pln-reasoner.scm:97:17: In procedure pln->sureal:
In procedure car: Wrong type argument in position 1 (expecting pair): ()
```

AmeBel: Have you run `sudo make install`? If you have, then run `rm -r ~/.cache/guile/*`

ngeiswei: @AmeBel, I have installed and cleared guile's cache, but I'm still having the same problem.

AmeBel: Assuming you are using the latest atomspace, also remove the old directory for guile module installation `sudo rm -r /usr/local/share/opencog/scm`

ngeiswei: It works, thanks!

PR #3377, opened 2018-12-13
General cleanup of the CMakeile

linas:
As with other components, this removes cmake policy CMP005 regarding backslashes.

This requires pull req opencog/atomspace#1951


linas:
Merge despite a circle-ci breakage; the breakage appears to be due to a an unrelated change. Debugging now.

Issue #3349, opened 2018-11-29
Clean up PLN scheme module, improve documentation and API


ngeiswei:
The pln scheme module needs some cleaning.

1. It's clear `focus-set.scm` doesn't belong to pln and should be moved somewhere else, perhaps to `opencog/ghost/procedures`.
2. I'm not sure `filter-for-pln` is still useful. It doesn't seem to be used anywhere.
3. I suspect that `simple-forward-chain` is no longer useful as the same behavior can be probably obtained by setting a high complexity penalty (see `ure-set-complexity-penalty`) so that the forward chainer behaves like breadth first search.

With that the README.md needs to be dramatically improved, with a proper usage section, etc.

The API can probably be improved as well, to make it easier to load some subset of PLN, etc.

ngeiswei:
@AmeBel could you provide me some feedback on the cleaning part of the pln scheme module? Like where should `focus-set.scm` be moved, etc. I don't mind to do it, just want to be sure I'm not creating destructive interference with your work.

ngeiswei:
@AmeBel @leungmanhin I'm about to carry out that cleaning. How can I test if I haven't screwed up anything? Will the unit tests catch it (that would be ideal)? Or should I run some examples, if so which?

ngeiswei:
Thanks @leungmanhin, OK, I'll make a unit test out of it prior to applying with my changes, unless @AmeBel has a better suggestion.

AmeBel:
@leungmanhin It is the same as [eg-trail-3](https://github.com/opencog/opencog/blob/3358f01a0e3e4cbb2d879e8268fc968af2f3d4da/opencog/ghost/procedures/pln-trail-3.scm#L149). Testing now

AmeBel:
@ngeiswei https://github.com/opencog/opencog/pull/3371 adds the above as a test. Before running `cmake ..` you will need to run relex server using [opencog-server.sh](https://github.com/opencog/relex/blob/e8e3ef40b8bafb8dbadd477a477c5f4801c45371/opencog-server.sh). 

Will configure circle-ci to run the test tomorrow.

ngeiswei:
Awesome.

I can't run it yet (in spite of having relex running), see https://github.com/opencog/opencog/pull/3371#discussion_r240564593 and https://github.com/opencog/opencog/pull/3371#discussion_r240568708 if that helps.

AmeBel:
[...]

PR #3383, opened 2018-12-14
Clean pln module

ngeiswei:
Take care of the cleaning part of #3349


ngeiswei:
@leungmanhin @AmeBel I've moved `focus-set.scm` under `ghost/procedures` and explicitly load it at the end of `procedures.scm`. Unit tests pass so I think it's all right but you may want to double check just in case.

PR #3384, opened 2018-12-17
UPdate cpack configuration for debian packaging

AmeBel:


linas:
How is this related to the debian packaging that @mhatta is working on, in the (newly created) opencog/opencog-debian repo? Is this stuff duplicated? How does this work?

AmeBel:
It is not related. Run `make package`  and get the result in `build/packages`. The results will be available at https://bintray.com/opencog. A test repo can be found at https://bintray.com/octool/deb-pkgs and will be part of the ci process on master and for building docker images.

I will likely use the relex packaging script for opencog-debian though.

AmeBel:
And it is not duplicated, I was working on this per the request of Cassio. That is why I did the cmake guile install update, and will be doing some for python and possibly other cmake configurations. Will also be making link-grammar available on by using https://salsa.debian.org/debian/link-grammar/tree/master.

linas:
so, lets see, you get installable .deb packages, which in some sense "should just work", but since they don't go through the standard Debian project build servers and testing environments, they can't ever be "official".  

Can @mhatta use the cpack stuff to make his life easier, or is that a different process?

linas:
oh, debian might be dropping support for link-grammar on 12 January, unless a maintainer steps up. @mhata said he might do that. Or something like that; I might be misunderstanding something.

mhatta:
@linas I'm not familiar with CPack, I'll check it.  My understanding is the same as yours, .debs generated by CPack can't be uploaded to the official Debian archive even if I gpg-signed it. but I might be wrong.

Anyway, I believe having up-to-date and working packages is REALLY important for OpenCog.  That's the whole point.

PR #3385, opened 2018-12-17
Revert some of the changes in opencog/opencog#3383

AmeBel:
The removal was a mistake made as there are no tests that use the function. These functions
are needed for some of the previous pln-trails and also for pln->sureal.


ngeiswei:
@AmeBel please note that `simple-forward-chain` are now part of the rule-engine module https://github.com/opencog/atomspace/blob/master/opencog/scm/opencog/rule-engine/rule-engine-utils.scm#L556-L575

PR #3390, opened 2018-12-19
Update CMakeLists.txt to export extra headers #52

Ophien:
Recently, I have tried to deploy a session based GHOST service for SNET that uses several CogServers to manage MindAgents for exclusive sessions. Consequently, the included headers are needed in order to allow CogServer object to be created.

Need the following headers to use the CogServer object directly inside a c++ program.
-AgentRunnnerBase.h
-AgentRunnerThread.h

### DIFF
--- a/opencog/cogserver/server/CMakeLists.txt
+++ b/opencog/cogserver/server/CMakeLists.txt
@@ -93,6 +93,8 @@ INSTALL (TARGETS server DESTINATION "lib${LIB_DIR_SUFFIX}/opencog")
 
 INSTALL (FILES
 	Agent.h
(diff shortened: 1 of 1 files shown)


ngeiswei:
Thanks @Ophien

linas:
Umm, but agents and agent runners are kind-of deprecated, yeah?  They have some utterly horrid design that breaks all the common-sense rules of programming... we no longer live in the era of MS-DOS and MacOS 9 ... !?

linas:
I mean, seriously -- don't use agents ... if this feature is really really needed, then we need to put some warm body on it, to redesign this stuff from scratch.

PR #3394, opened 2018-12-27
Sync singnet ull changes 

AmeBel:
These are ull changes restructured and merged to master. 

- **Why is this done?**
  - For syncing singnet with opencog.
  - For enabling diverging work on ull with as little as possible duplication of code.

- **What is done?**
  - `common.scm`, `link-pipeline.scm`, and `mst-parser.scm` are copied to `run-poc` folder, and changes made to them in singnet are applied as patches, so as to conserve the authorship history and avoid modifying of code in `(opencog nlp learn)`.
  - The changes are merged to opencog master for testing.
  - `run-poc/launch-cogserver.scm` is modified to use the code in `(opencog nlp learn)` by loading the files of the modules from load-path. This is done so as to redefine some of the functions in the copied files without making changes to `(opencog nlp learn)` code.
  - Reviewed by @leungmanhin and @glicerico . Looks good to them.

- **What is next?**
  - Due to copying of files, there are duplication of code. This will be cleaned up by @glicerico


linas:
This duplicates a large number of files that do not have any changes in them; they are just byte-for-byte copies of existing files, placed in directories which even have the same name, but occur in different locations in the file hierarchy. 

There is another set of files that are almost-duplicates, and have minor or trivial changes that don't change the function, but merely copy the files, and alter the whitespace, or have other non-functional changes, e.g. to the documentation. Some of these copies are in triplicate.  

There is a third set of files that make functional changes. Some of the functional changes are reasonable. Some are mis-designed, but are not hard to fix. Some do not seem reasonable, and were changes that were made based on a misunderstanding of how the system works. 

I would strongly prefer it if all of these issues were addressed.  I firmly believe that all of these issues are resolvable. None of these issues are terribly hard to resolve.  Can this please be done?

AmeBel:
@linas ok will do

linas:
OK, so, just to be clear: the entire directory `nonbreaking_prefixes` seem to be duplicated. I think that's several dozen worth of files.

AmeBel:
@linas I have removed duplicated codes and add a cmake target for configuring in build directory the setup  required for running roc-poc. Running `make run-ull` copies files to the `build/run-ull/` directory. 

I will update the readmes for setup instruction after you review the present state.

ngeiswei:
I'm merging since @linas has given his approval.

ngeiswei:
Thanks to @AmeBel and everybody else involved to make that possible.

PR #3401, opened 2019-01-04
Documentation update

AmeBel:
This is update for changes made in opencog/opencog#3394. Also whitespace corrections

### DIFF
--- a/opencog/nlp/learn/run-poc/README-run.md
+++ b/opencog/nlp/learn/run-poc/README-run.md
@@ -13,7 +13,7 @@ Connecting to Cogserver:
                       takes as language the processing mode and the language
                       (inside config folder).
 
- -- `opencog-pairs-??.conf` contains configuration settings for the cogserver, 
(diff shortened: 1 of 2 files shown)


linas:
Umm, making these whitespace changes might make it hard to eventually fix the duplication problem. -- Cladia just made a copy of `opencog/nlp/learn/README.md` made some random (but I think minor??) changes to it, and now it sits in `opencog/nlp/learn/run-poc/README.md`  -- we really shouldn't have two copies of almost-the-same-README with mystery-differences between them...

linas:
Aiee. I just did a side-by-side diff. Apparently Claudia reformatted the entire README, so that although it contains almost exactly the same text, all line breaks are different, various paragraphs appear in differing order, and removed a large quantity of some parts of the text.  Merging these back into one is going to be a chore. :-(   The fact that there's now a lot of work needed to resolve the differences is precisely why I was strongly requesting that the cleanup be done earlier, while it was still relatively easy to do.

PR #3422, opened 2019-02-02
General AtomSpace cleanup

linas:
Twin pull request from opencog/atomspace#2035 -- one reason for the circular shared library dependency was an accumulation of unused cruft, and also some very special-case code not used by the atomspace (e.g. agent code, NLP code) .. move that code to here, to the users who need it.


linas:
Unit tests won't pass until opencog/atomspace#2035 is merged; I'm merging now.

PR #3428, opened 2019-02-15
Misc pattern-miner scheme code cleanup

linas:
Simplify the temp-atomspace logic.


ngeiswei:
I had forgotten about the atomspace stack. Thanks.

PR #3429, opened 2019-02-15
Fix use-after-free bug in unit test

linas:


linas:
exposed by pull req opencog/atomspace#2074

PR #3430, opened 2019-02-16
fix CogserverUTest double-free crash

linas:
The AttentionBank is another abandoned, unused subsystem that needs love and attention from a dedicated maintainer.


linas:
exposed by pull req opencog/atomspace#2074

PR #3437, opened 2019-02-19
Move attentionbank examples from atomspace.

linas:
Part one of three, issue #3432 - consolidate all attention-bank code to one repo.


linas:
@misgeatgit please review and merge when ready.

PR #3447, opened 2019-02-27
Use the "set" api, not the "list" api.

linas:
Minor change needed for general cleanup of atomspace.


linas:
requires opencog/atomspace#2112

PR #3453, opened 2019-03-05
Fix cog-number applied to TimeNodes

eddiemonroe:
(cog-number (TimeNode "12345")) leads to an error.
```
ERROR: In procedure cog-number:
In procedure cog-number: Wrong type (expecting NumberNode): (TimeNode "12345")
```


linas:
circle-ci failures are because the tested branch did not have master merged into it. Master passes unit tests. Be sure to pull the latest.

PR #3455, opened 2019-03-06
Fix pattern miner

### LINKED ISSUE #3454: PatternMiner SCM API returns "Invalid handle"
Attempt to run pattern miner example finishes with error:
```
~/projects/opencog/examples/learning/miner/simple$ guile simple.scm 
Backtrace:
           7 (apply-smob/1 #<catch-closure 7fd5c735bb80>)
In ice-9/boot-9.scm:
    705:2  6 (call-with-prompt _ _ #<procedure default-prompt-handle…>)
In ice-9/eval.scm:
    619:8  5 (_ #(#(#<directory (guile-user) 7fd5c6deb140>)))
In ice-9/boot-9.scm:
   2312:4  4 (save-module-excursion _)
  3822:12  3 (_)
In /home/vital/projects/opencog/examples/learning/miner/simple/simple.scm:
    24:16  2 (_)
In opencog/miner/miner-utils.scm:
    41:14  1 (desc-sort-by-tv-strength #<Invalid handle>)
In unknown file:
           0 (cog-outgoing-set #<Invalid handle>)

ERROR: In procedure cog-outgoing-set:
In procedure cog-outgoing-set: Wrong type argument in position 1 (expecting opencog atom): #<Invalid handle>
```

The reason is that after miner returns the result handle its original atomspace is cleaned up and atomspace link in atom set to `NULL`. After that `scm_to_handle` procedure checks that atomspace is `NULL` and returns `Handle::UNDEFINED`.

ngeiswei:
Fix #3454 + add `none` surprisingness measure, as surprisingness is currently only defined for conjunctions of components.

### DIFF
--- a/opencog/learning/miner/miner-utils.scm
+++ b/opencog/learning/miner/miner-utils.scm
@@ -481,8 +481,9 @@
      enable incremental conjunction expansion, see the
      #:incremental-expansion option.
 "
-  (let* (;; Create a temporary child atomspace for the URE         
-         (parent-as (cog-push-atomspace))
+  (let* (;; Create a temporary child atomspace for the URE
+         (tmp-as (cog-new-atomspace (cog-atomspace)))
+         (parent-as (cog-set-atomspace! tmp-as))
          (texts-concept? (and (cog-atom? texts)
                               (eq? (cog-type texts) 'ConceptNode)))
          (texts-cpt (if (not texts-concept?)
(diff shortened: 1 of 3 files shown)


opencog/learning/miner/miner-utils.scm
```diff
@@ -514,21 +517,29 @@
...
-
-          (desc-sort-by-tv-strength isurp-results)))))
+               (patterns-lst (cog-outgoing-set patterns)))
+
+          (if (equal? surprisingness 'none)
+
+              ;; No surprisingness, simple return the pattern list
+              (begin
+                (cog-set-atomspace! parent-as)
+                ;; TODO: delete tmp-as but without deleting its atoms
```
linas: ahh, OK, sorry.

linas: I think the root bug was that BIT was holding on to a temp atomspace even when parent atomspace is gone.this might be?? variant of that.

PR #3456, opened 2019-03-07
Update apiatomcollection.py

StrangeTcy:
Made print python3-compatible


linas:
Please not that the API implemented there is out-of-date, and needs to be modernized, in general...

linas:
Anyway, thanks!

PR #3459, opened 2019-03-15
time-octomap

misgeatgit:
Added guile bindings which accept and return locations as Values.

### DIFF
--- a/opencog/spacetime/octomap/OctoMapNode.h
+++ b/opencog/spacetime/octomap/OctoMapNode.h
@@ -63,6 +63,12 @@ public:
     {
         return octomapPtr;
     }
+
+    void set_map(TimeOctomapPtr& map)
+    {
+      octomapPtr = map;
+    }
+
     static Handle factory(const Handle&);
 };
 
--- a/opencog/spacetime/octomap/OctoMapValue.cc
+++ b/opencog/spacetime/octomap/OctoMapValue.cc
@@ -14,6 +14,11 @@ void OctoValue::update() const
     _value = {it->x(), it->y(), it->z()};
 }
 
+void OctoValue::update(const std::vector<double>& loc) const
+{
+  _value = {loc[0], loc[1], loc[2]};
+}
+
 OctoValue::OctoValue(const HandleSeq& hseq) : FloatValue(OCTO_VALUE)
 {
 
@@ -25,6 +30,7 @@ OctoValue::OctoValue(const HandleSeq& hseq) : FloatValue(OCTO_VALUE)
     _item = hseq[0];
     _octo_node = hseq[1];
     _om = OctoMapNodeCast(_octo_node)->get_map();
+    update();
 }
 
 bool OctoValue::operator==(const Value& other) const
--- a/opencog/spacetime/pointmemory/PointMemorySCM.cc
+++ b/opencog/spacetime/pointmemory/PointMemorySCM.cc
@@ -31,9 +31,13 @@
 #include <opencog/atoms/base/Link.h>
 #include <opencog/atoms/base/Node.h>
 #include <opencog/atoms/core/NumberNode.h>
+#include <opencog/atoms/value/LinkValue.h>
 #include <opencog/guile/SchemePrimitive.h>
 #include <opencog/spacetime/atom-types/atom_types.h>
+
 #include <opencog/spacetime/octomap/TimeOctomap.h>
+#include <opencog/spacetime/octomap/OctoMapNode.h>
+#include <opencog/spacetime/octomap/OctoMapValue.h>
 
 namespace opencog
 {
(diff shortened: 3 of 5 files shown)


opencog/spacetime/octomap/OctoMapNode.h
```diff
@@ -63,6 +63,12 @@ class OctoMapNode : public Node
     {
         return octomapPtr;
     }
+
+    void set_map(TimeOctomapPtr& map)
```
linas: Atoms are immutable. There must not be such a call. The reason for that immutability is to guarantee thread safety.

opencog/spacetime/octomap/OctoMapValue.cc
```diff
@@ -14,6 +14,11 @@ void OctoValue::update() const
     _value = {it->x(), it->y(), it->z()};
 }
 
+void OctoValue::update(const std::vector<double>& loc) const
+{
```
linas: Values are meant to be externally immutable. So again, no calls of this form. Again, one of the primary drivers is thread safety.

opencog/spacetime/pointmemory/PointMemorySCM.cc
```diff
@@ -47,13 +51,16 @@ class PointMemorySCM
 	// Check if the named map exists.
 	void have_map(const Handle&);
 
-	std::map<Handle, TimeOctomap<Handle>*> tsa;
+	std::map<Handle, std::shared_ptr<TimeOctomap<Handle>>> tsa;
 	time_pt get_map_time(const Handle& map_name, const Handle& elapse);
 
 public:
+    static std::map<Handle,Handle> handle_octomap_map;
```
linas: From what I can tell, you don't actually need this. If you already have the OctoMap Node, you don't need to look it up a second time!  In the code below, `map_name` is already the map that you want.  The indirection is not needed ...

opencog/spacetime/pointmemory/PointMemorySCM.cc
```diff
@@ -47,13 +51,16 @@ class PointMemorySCM
...
 	// Create a map
 	Handle create_map(Handle map_name, Handle resolution);
 
+    Handle create_map_ov(Handle map_name, Handle resolution);
```
linas: Indentation.

linas:
OK, I cannot figure out at all what you are trying to do here. The things that the octomap is supposed to be tracking are Atoms -- the location of "things". It does not make sense to track "values" because values are not "things".  So, for example, the octoMap needs to track the three-D location of `(Concept "chair")` or `(Concept "person standing in front of me")`.  It does not make sense to track the 3D-location of  `(FloatValue 42 33 68 72)`

Can you open a new issue, and explain what the problem is, that you are trying to solve, here?  Because this sure does not feel like the correct solution. 

For now, I'm going to just close this pull req; that seems like the easiest thing to do, until the meta-issue is resolved, first.

misgeatgit:
I thought that was the whole point of designing the OctoValue. If we don't really care about updating the xyz values contained in the OctoValue, FloatValue would have been sufficient enough for solving our main issue, which is the expensiveness of creating atoms as a return value for the xyz location of an object in the Octomap. 
With FloatValues, there is no way to get the latest position of the object unless we explicitly call  the `get-current-position`(this is not the exact syntax) where as with OctoValues, it gets updated everytime when the values(i.e the xyz) locations are accessed.

opencog/spacetime/octomap/OctoMapValue.cc
misgeatgit: Noted.

linas:
OK, what you implemented is not the design we had on the white-board. Look at `examples/atomspace/stream.scm` -- it uses RandomValue as a generic example of an OctoValue (its also a generic example of a neurla-net value).  Next, look at the implementation of `RandomValue.cc` -- it uses a random-number generator to generate random number streams.

The correct octovalue design replaces the random-number generator by x,y,z values obtained from the OctoMap.  So, whenever someone asks the OctoValue "what numbers do you have right now?", the C++ code gets the correct xyz location from the octomap.

The contructor for OctoValue tells it which octomap to use (to find those values) AND also which object is being tracked. -- for example. `(ConceptNode "person in front of me")`.  So the OctoValue knows about the map, already, and it knows which object in the map it needs to track. It just needs to return the correct xyz for that object ... that's all. 

Anyway, I'm still confused, because I thought you already finished writing that code. Its done. I thought it worked already. There was nothing more to do there.  Except maybe write unit tests?  There is more to do, but just not here.  Next post.

linas:
So anyway, I think you are done with the basic octomap code. I don't think anything more is needed. (maybe I'm wrong?)  Here's what needs to be done:  First, unit tests for the existing OctoValue and the existing OctoMapNode   This is not hard, but you do need to be through and complete.

The hard part that is not done is to implement "prepositions" -- near, far, next-to, in-front-of, behind, above, below.  These preps need to take *two* octovalues, and answer yes/no for each.  It needs to work so that both PLN and also ghost can use it.  So they need to be PredicateNodes of some kind.  We need to have this discussion in some other github issue. Isn't there an open github issue for this, already?

linas:
well,even before the unit tests -- we need several working examples of the current octovalue design. So -- create theat example -- create `(Concept "wandering human")` and push that into the octomap.   Use a random-number generator to update the xyz coords of `(Concept "wandering human")` . In real life, the 3D coords would be coming from a ROS node that @jdddog  would provide -- some SLAM system or whatever. But for the example, just use a random number generator.

The main body of the example then shows how to use the OctoValue, to obtain the current x,y,z coords.

The example could even be a super-simple PredicateNode that answers the question "is in front of", for  two different OctoValues. Maybe/probably should use https://wiki.opencog.org/w/PredicateFormulaLink for the formulas.

To repeat: the example pretneds to be a ROS node, jams random-walk x,y,z locations into the octomap, uses octovalues to get xyz, and PredicateFormula to figure out in-front/behind relations.   I think that everything you need to write this example is already done - it already exists in the codebase.

PR #3463, opened 2019-03-22
Change the formula used in inheritance-to-member

leungmanhin:
@ngeiswei I suppose it should be `inheritance-to-member-side-effect-free-formula` instead of `member-to-inheritance-side-effect-free-formula` in the `inheritance-to-member-formula` ?


ngeiswei:
Correct.

Although I would argue that the rule is wrong to begin with, the premise should be
```
; InheritanceLink
;   Set B 
;   C
```
either this, or C should be restricted to a certain class (such as https://wiki.opencog.org/w/Soggy_Predicates though probably something different).

Beware if you use rules in the wip directory, some are very very sketchy.

PR #3473, opened 2019-04-19
Replace old, deprecated functions with modern ones

linas:
These old functions were quite slow and inefficient, and were
deprecated many years ago. Stop using them.


linas:
Merging despite unit-test failure -- the same unit test -- `GhostUTest` is also failing in the master branch.

Issue #3484, opened 2019-04-26
Running Ghost sample crashes with Segmentation fault from time to time


stellarspot:
I run the code [sample-ghost-notebook.scm](https://github.com/stellarspot/opencog-tutorial/blob/1de8396546993ccc54afd6a3877c13e4203d7298/samples/ghost/sample-ghost-notebook.scm) and it crashes with Segmentation fault error sometimes.

Here is the result of call `catchsegv guile samples/ghost/sample-ghost-notebook.scm`
```
[2019-04-26 15:03:52:974] [WARN] [GHOST] Did you forget to link a goal to the rule?
[2019-04-26 15:03:53:008] [WARN] [GHOST] Did you forget to link a goal to the rule?
[2019-04-26 15:03:53:067] [INFO] [say] (softmegacorp is a fantastic company !)
[2019-04-26 15:03:53:110] [INFO] [say] (i work in SoftMegaCorp company.)
[2019-04-26 15:03:54:138] [INFO] [say] (hardmegacorp is a fantastic company !)
*** Segmentation fault
Register dump:

 RAX: 00007f2e6b6134a8   RBX: 00000000012c9200   RCX: 00007f2e61ef3440
 RDX: 0000000000000006   RSI: 0000000000000000   RDI: 000000000111fe80
 RBP: 0000000000000000   R8 : 00000000012c9200   R9 : 00007f2e6b613138
 R10: 00007f2e72e969a0   R11: 00007f2e66425fe0   R12: 00007f2e61ef34e0
 R13: 00007f2e61ef3600   R14: 00007f2e663fde30   R15: 00007f2e61ef35f0
 RSP: 00007f2e61ef3488

 RIP: 00007f2e6b613450   EFLAGS: 00010246

 CS: 0033   FS: 0000   GS: 0000

 Trap: 0000000e   Error: 00000015   OldMask: 00000000   CR2: 6b613450

 FPUCW: 0000037f   FPUSW: 00000420   TAG: 00007f2e
 RIP: 68bb9217   RDP: 00000000

 ST(0) ffff ffffffffb5829040   ST(1) ffff c0a15421076a650b
 ST(2) ffff e000000000000000   ST(3) ffff 8000000000000000
 ST(4) ffff 96b81373055aef41   ST(5) ffff 8000000000000000
 ST(6) ffff 8000000000000000   ST(7) 8000 8000000000000000
 mxcsr: 1fa0
[...]

PR #3487, opened 2019-04-30
#3484 Join to psi thread in pshi-halt function to avoid segmentation fault

stellarspot:
Psi starts a new thread and checks the provided atom value if the thread should stop.
When Scheme exits it can destroy the Atomspace first so when psi checks the atom on a separated thread the atom or its values may not exist anymore.

The fix stores the created thread object with the provided atom in a list and uses the stored atom to join to the thread in the psi-halt function.

### DIFF
--- a/opencog/openpsi/main.scm
+++ b/opencog/openpsi/main.scm
@@ -23,6 +23,10 @@
   opl
 )
 
+; list of pairs (psi-thread component)
+; to close psi threads before exiting
+(define psi-threads '())
+
 ; --------------------------------------------------------------
 ; Define the component category. Components are like mind-agents
 ; but there activities are defined using openpsi-rules and action-selectors
@@ -102,8 +106,10 @@
   (if (not (psi-running? component))
     (begin
       (cog-set-value! component (Predicate "run-loop") (StringValue "#t"))
-      (call-with-new-thread
-        (lambda () (cog-evaluate! (cog-value component (Predicate "loop"))))))
+      (save-thread
+        (call-with-new-thread
+          (lambda () (cog-evaluate! (cog-value component (Predicate "loop")))))
+        component))
   )
 )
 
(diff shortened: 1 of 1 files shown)


opencog/openpsi/main.scm
```diff
@@ -134,15 +140,36 @@
...
+  save-thread THREAD COMPONENT
+
+  Save thread in the psi-threads list to properly terminate it
+  before exit.
+ "
+ (set! psi-threads (cons (cons thread component) psi-threads))
+)
+
+; -------------------------------------------------------------
+(define* (psi-halt component #:optional (timeout 3))
```
linas: I'm thinking that an infinite default timeout would be better. That would be guaranteed to always have the correct behavior (although it might hang if there are other bugs).

opencog/openpsi/main.scm
```diff
@@ -134,15 +140,36 @@
...
   psi-halt COMPONENT
 
   Halts COMPONENT's previously-started psi loop thread. The thread is
   started by calling `(psi-run COMPONENT)`.
 "
+  (define (psi-join lst)
+    (if (not (null? lst))
+      (let
+        ((thread (car (car lst)))
+         (comp (cdr (car lst))))
```
linas: So, `(car (car lst))` is the same as `(caar lst)` and `(cdr (car lst))` is the same as `(cdar lst)`  and there's a large collection of utilities here: https://srfi.schemers.org/srfi-1/srfi-1.html

linas: ... and for accessing elements of a pair, you can use `first` and `second` .. or maybe even `list-ref`  ... here, it does not really matter, but names like that are just slightly more readable for conventional programmers coming from other languages.

AmeBel:
Never mind got a better context after reading https://github.com/opencog/opencog/issues/3484

PR #3490, opened 2019-05-09
Possible fix for miner bug opencog/atomspace#2159

linas:
I have not tested this, but based on problem description in opencog/atomspace#2159, this should fix the problem.  Please review, test and merge at leisure.

### DIFF
--- a/opencog/learning/miner/miner-utils.scm
+++ b/opencog/learning/miner/miner-utils.scm
@@ -488,10 +488,9 @@
      enable incremental conjunction expansion, see the
      #:incremental-expansion option.
 "
-  (let* (;; Create a temporary child atomspace for the URE
-         (tmp-as (cog-new-atomspace (cog-atomspace)))
-         (parent-as (cog-set-atomspace! tmp-as))
-         (texts-concept? (and (cog-atom? texts)
+  ;; Create a temporary child atomspace for the URE
+  (cog-push-atomspace)
+  (let* ((texts-concept? (and (cog-atom? texts)
                               (eq? (cog-type texts) 'ConceptNode)))
          (texts-cpt (if (not texts-concept?)
                         ;; Construct a temporary concept containing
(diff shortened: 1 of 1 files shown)


ngeiswei:
@linas, thanks but it doesn't solve the problem because, although it prints the solutions, as soon as the temporary atomspace is popped then the solutions disappear again.

But I don't think it is a bug now that I understand how guile deals with atomspace references. Instead I should redesign the API, like probably copying the results in the user's atomspace or something.

I'm fine with the behavior itself (deleting unreferenced atomspaces) I just think it should be better documented. To that end I have added a word of warning in https://wiki.opencog.org/w/Scheme#cog-set-atomspace.21, and will do that in `cog-set-atomspace` and `cog-pop-atomspace` docstrings.

ngeiswei:
BTW, @linas, could have slightly more descriptive branch names (than `mi` for instance), it would make browsing your branches a bit easier. And branch names are stored in the git history so it's an extra reason to have them minimally informative.

ngeiswei:
Closing as it doesn't properly address the problem. But it helped, thanks!

linas:
What's the problem? `miner-utils.scm` has an obvious bug in it. This fixes the obvious bug. 

Are you telling me that after applying this patch, you still get some other crash?  If so, maybe a new bug report is in order?

linas:
> how guile deals with atomspace references

This is how *any* garbage-collected or reference-counted language or library deals with *any* reference; its nothing particular to the atomspace.  Whenever there's some object, and there are no remaining references to it, it is fair game for finalization.

linas:
The pull req opencog/atomspace#2164 is attempting to force the issue.  Currently, for me, all unit tests pass, so I'm guessing this particular utility is not currently tested.

ngeiswei:
This fix has 2 problems
1. It doesn't return the solution set
2. Even if modified to return it then the solution set disappears after the temporary atomspace is popped.

The problem requires to modify the user API, so it is best left for another PR that I'll take care of. I know where I'm going now. Thanks.

Issue #3492, opened 2019-05-15
Child AtomTable is released later than parent one when using Scheme in Python


stellarspot:
Calling OpenPsi from Python leads to the error (see the code below):
```
terminate called after throwing an instance of 'opencog::RuntimeException'
  what():  AtomTable - deleteing atomtable with subtables! (/home/stellarspot/opencog/master/atomspace/opencog/atomspace/AtomTable.cc:128)
Aborted (core dumped)
```
The problem is that Python releases main atomspace before it releases OpenPsi in Scheme.

Here is what happening during the program execution.

The parent atomspace is created in Python.
GetLink creates child AtomTable (_num_nested is increased to 1) which is stored in the cache when it is not necessary (_num_nested is decreased to 0).

OpenPsi creates OpenPsiImplicator which is a static variable in the `OpenPsiImplicator& opencog::openpsi_implicator()` method.
OpenPsiImplicator extends DefaultPatternMatchCB which grabs the AtomTable from the cache (_num_nested is increased to 1).

When the program finishes Python calls  the main AtomSpace destructor which frees its AtomTable and throws the RuntimeException (the child AtomTable has not been released).

```python
from opencog.type_constructors import *
from opencog.utilities import initialize_opencog, finalize_opencog

# Initialize AtomSpace
atomspace = AtomSpace()
initialize_opencog(atomspace)

PARENT_RECT = ConceptNode("rectangle")
KEY_CORNES = ConceptNode("key-corners")


def add_rect(id, name, x1, y1, x2, y2):
    rect = ConceptNode(name)
    rect.set_value(KEY_CORNES, FloatValue([x1, y1, x2, y2]))
    return SetLink(
        InheritanceLink(rect, PARENT_RECT))


# Called by OpenPsi Action
def find_the_same_rects(params):
    rect1 = params.get_out()[0]
    rect2 = params.get_out()[1]
    # Mark rectangles as handled
    return SetLink(
        EvaluationLink(PredicateNode("handle-rect"), rect1),
        EvaluationLink(PredicateNode("handle-rect"), rect2))


# OpenPsi components

def check_the_same_rects(rect1, rect2):
    if rect1.get_value(KEY_CORNES) == rect2.get_value(KEY_CORNES):
        return TruthValue(1.0, 1.0)
    else:
        return TruthValue(0.0, 1.0)


# OpenPsi context
def context_find_the_same_rects():
    return AndLink(
        InheritanceLink(
            VariableNode("$RECT_1"),
[...]

PR #3496, opened 2019-05-17
#3492 Move Satisfier from OpenPsiImplicator so it is possible to release AtomTable

stellarspot:
See the issue #3492 

The idea of the fix is to move the Staisfier from the OpenPsiImplicator class hierarchy and create it only by demand in `check_satisfiability()` method and remove in `imply()` method.

This relies on the methods `check_satisfiability()` and `imply()`  order  execution.
For example if `check_satisfiability()` is called and returns TRUE_TV but `imply()` is not called the AtomTable belonging to `DefaultPatternMatchCB` will not be released and the crash described in #3492 happens again.


vsbogd:
As `imply()` and `check_satisfiability()` methods share state I don't see good solution without changing scheme API. For instance PR #3496 is a fragile solution which will have different side effects when methods `imply()` and `check_satisfiability()` are called in different order.

Straightforward solution to instantiate `OpenPsiSatisfier` at the beginning of each method and delete it at the end has performance impact obviously. More performant options:
- to pass instance of `OpenPsiSatisfier` to the client with responsibility to delete it;
- to add one method with single callback argument and pass `OpenPsiSatisfier` to this callback thus allowing external user writing any logic in the callback;
Both options require change in API.

linas:
> Straightforward solution to instantiate OpenPsiSatisfier at the beginning of each method and delete it at the end has performance impact obviously

It should have a rather very small performance impact - negligable, I think.  All that Satisfier is doing is allocing a struct, and initializing two or three pointers, and nothing more.   It should be very fast to create and destroy.  Most other code, outside of openpsi, just allocs it on stack (block scope) - that skips the overhead of new/malloc.

linas:
> All that the Satisfier ctor is doing

I mean .. all that the Satisfier ctor is should be doing is something very minimal. I haven't checked recently.  It was originally designed to be very small and efficient; maybe that design went sideways.

linas:
This cannot be merged until the `GhostUTest` and the `OpenPsiImplicatorUTest` both pass.

linas:
Hmm. Now seven different unit tests fail ...

stellarspot:
The problem with the GhostUTest was the following.
There are usually paired calls: OpenPsiImplicator::check_satisfiability(...) and if it passes OpenPsiImplicator::imply(...).

The fix created
* OpenPsiSatisfier in the  OpenPsiImplicator::check_satisfiability(...) and called its check_satisfiability(...) method.
* OpenPsiSatisfier in the  OpenPsiImplicator::imply(...) and called its both check_satisfiability(...) and imply(...) method.

The OpenPsiSatisfier.check_satisfiability(...) first removes the given query from the satisfiability_cache and then uses grounding to find solution.

Usually its fine because between OpenPsiImplicator::check_satisfiability(...)  and OpenPsiImplicator::imply(...) the pattern in question located in AtomSpace.

This is not true for the Ghost rejoinder because after  OpenPsiImplicator::check_satisfiability(...) call it is removed from the AtomSpace so consequent OpenPsiImplicator::imply(...) removes it from the cache and is not able to find it again.

I updated the code to not call OpenPsiSatisfier.check_satisfiability(...)  inside the  OpenPsiImplicator::imply(...) method.

stellarspot:
There are 3 failed tests in [CircleCI](https://circleci.com/gh/opencog/opencog/2107):
```text
The following tests FAILED:
	 17 - AnaphoraTest (OTHER_FAULT)
	 19 - SuRealUTest (Failed)
	 20 - MicroplanningUTest (Failed)
```
SuRealUTest and MicroplanningUTest tests pas on my local system with the current fix.
AnaphoraTest still fails.

linas:
Thanks!  Seems reasonable, merging.

Issue #3500, opened 2019-05-21
Cython tests do not run


stellarspot:
Here is the link to the CircleCI run and I can't find there Cython tests: CythonServer and CythonStatistics:
https://circleci.com/gh/opencog/opencog/2024

On my local machine the PYTHONINTERP_FOUND is not set and it is not defined in the opencog project.

PR #3501, opened 2019-05-21
Add OpenPsi Python wrapper

stellarspot:
This is the proposal to add OpenPsi Python wrapper methods and classes.

OpenPsi has OpenPsiRules and OpenPsiSCM classes.
OpenPsiSCM class was used to bind it in Cython because it already caches some results. OpenPsiSCM methods access has been changed from private to public to allow to call them by Cython. It seems that there should be a class that uniformly provides its methods to Scheme and Python bindings.

The OpenPsi Python test is added to the tests/cython/openpsi dir. It is possible to run it from its dir by Python.
There is an issue that cython tests do not run: #3500 

OpenPsiRules class has methods that takes a rule as an argument like `Handle get_action(const Handle rule)`. It may be better to have a separate OpenPsiRule class which contains a rule as a field and allows to call these methods without the rule argument.


stellarspot:
run, halt, and init_component methods are added to OpenPsi Python wrapper which just makes call to scheme.

linas:
The call to `get_openpsi_scm()` strikes me as hacky, but .. whatever. I see why you did it.

@AmeBel is there any reason not to merge this? I try not to poke my fingers into open-psi-related matters....

PR #3507, opened 2019-05-25
Avoid use of deprecated function

linas:
There is no benefit from using this function.


linas:
```
The following tests FAILED:
	 17 - AnaphoraTest (Failed)
	 26 - PythonModuleUTest (SEGFAULT)
```
but not due to this pullreq

PR #3503, opened 2019-05-22
Run python tests on  python3

AmeBel:
Fixes #3500

PR #3508, opened 2019-05-25
Fix one reason that python unit test is crashing

linas:


linas:
This attempts to fix the crashes due to #3503 and 3509 - one was a null-pointer deref; the other an obviously bad RPATH  Mergine this anyway, even though additional bugs remain.

Issue #3509, opened 2019-05-25
Unit test crashes due to python3


linas:
After pull req #3503 two unit tests have been crashing, fouling up circleci:
```
	 17 - AnaphoraTest (Failed)
	 26 - PythonModuleUTest (SEGFAULT)
```
My best guess is that `AnaphoraTest` needs to be ported to python3. Likewise for the other.

I looked around, I hit this:
```
Running cxxtest tests (4 tests)[2019-05-25 05:30:14:637] [DEBUG] registering opencog::Request "loadpy"
[2019-05-25 05:30:14:637] [DEBUG] Creating opencog::Request instance with "loadpy"
[2019-05-25 05:30:14:637] [DEBUG] [loadpy Request] execute
Exception ignored in: 'opencog.agent_finder.load_req_agent_module'
TypeError: Can't convert 'bytes' object to str implicitly
Got >>><<<
.Exception ignored in: 'opencog.agent_finder.load_req_agent_module'
TypeError: Can't convert 'bytes' object to str implicitly
[2019-05-25 05:30:14:637] [DEBUG] [Request] destructor
```
This typical python3 error is this:
```
TypeError: Can't convert 'bytes' object to str
```
The bug is probably located in `opencog/cython/opencog/agent_finder.pyx` somewhere after line 44.

I suggest disabling both unit tests.  I don't think anyone is using agents. I don't really think agents should be fixed anyway, they're a goofy design that doesn't really make sense, so I'm fine with just disabling that code.

Anaphora -- well, at the current rate, it seems unlikely that anyone is going to be using that code any time soon. So, it too, is more-or-less an unmaintained experiment, and can be disabled.  Surely the python3 port can't be hard... but why?

linas:
Ohhh ... the Anaophora code does use agents ... see `opencog/nlp/anaphora/agents

ferrouswheel:
I'm investigating this since I've dealt with str/bytes difficulties and 2/3 incompatibility before.

Even though it's unrelated to what I should be doing, it's very difficult to make progress when a test suite is broken and you have no idea if incoming changes have broken stuff or if it's always been broken.

ferrouswheel:
The test doesn't show any TypeError if you run it separately - though 3 of tests fail: `test_conjunctions`,`test_pleonastic_if`, and `test_propose`.

```
~/src/opencog $ python3 --version
Python 3.6.7
[...]

PR #3512, opened 2019-05-29
Move AttentionValue type definition out of AtomSpace

linas:
Move it so that it lives with all the atom AttentionValue code.


linas:
The usual unit tests are failing, per bug report #3509

PR #3514, opened 2019-06-04
Update the Ghost example

leungmanhin:


leungmanhin:
The new `OpenPsiTest` in python is failing, apart from the two reported in https://github.com/opencog/opencog/issues/3509, but they are unrelated to this example update...

Issue #3513, opened 2019-06-04
Cython build broken if you don't enable SpaceTime component


ferrouswheel:
The file `opencog/cython/opencog/cogserver_type_constructors.pyx` assumes that `opencog/spacetime/octomap/octomap_types.pyx` is generated [here](https://github.com/opencog/opencog/blob/master/opencog/cython/opencog/cogserver_type_constructors.pyx#L21).

But this isn't generated when the SpaceTime component is disabled.

Either cython bindings need to be disabled when SpaceTime component is disabled, or some flag needs to be passed to this file to avoid importing the missing file. It wasn't immediately obvious how to pass cmake flags to pyx files, but I'm happy to try to fix it if someone points me in the right direction.

linas:
Does `#ifdef` work with pyx files?  I suppose not, but it's worth a try.

There are two other solutions: 
* generate octomap types, even if spacetime is disabled
* generate  an empty ampty types file if spacetime is disables (i.e. generate octomap_types file from an **empty** `atom_types.script` file.)

Either of the last two seem reasonable. It's presumably easier than trying to figure out how to add conditional compile statement to pyx.

linas:
A third solution -- the long-run solution, is to split out space-time into it's own repo, per issue #3391  

The only reason that the cogserver is loading spacetime types is that it was an "easy hack" back in the day.  Extracting spacetime into it's own stand-alone component would help makke the code more modular.  Of course, this is a larger task than the above fix suggestions...

ferrouswheel:
Opted for the second option as I couldn't find any info about passing a macro definition to influence pyx includes.

vsbogd:
I would just move spacetime server into separate library and export octomap types from it instead of cogserver_types, but leave it in opencog repository.

PR #3515, opened 2019-06-04
Generate empty types file so cython build works without octomap

ferrouswheel:
Fix for #3513

### DIFF
--- a/opencog/spacetime/CMakeLists.txt
+++ b/opencog/spacetime/CMakeLists.txt
@@ -1,6 +1,10 @@
 
 ADD_SUBDIRECTORY (atom-types)
+
 IF (HAVE_OCTOMAP)
 	ADD_SUBDIRECTORY (octomap)
 	ADD_SUBDIRECTORY (pointmemory)
+ELSE (HAVE_OCTOMAP)
+	FILE(WRITE "${CMAKE_CURRENT_BINARY_DIR}/octomap/octomap_types.pyx" "\n")
 ENDIF (HAVE_OCTOMAP)
+


linas:
Hmm...  `OpenPsiUTest` now fails with `Failure: ImportError (No module named 'opencog.type_constructors') ... ` and I'm pretty sure that was not failing before this. 

I'm guessing that manually touching the pyx file means that the compile step failed to actually compile the pyx into valid code. (Recall that pyx files are auto-converted to C++ code, and then the C++ code is compiled to create the python bindings, and your change cause the auto-convert-to-C++ step to be skipped.)

Try instead creating an empty `atom_types.script` file, and then triggering it's compile as it normally would (i.e. copy the needed parts of that CMakefile)

ferrouswheel:
#3517 has the same error, and OpenPsiTest fails for me on master even without my change.

linas:
Yeah, but the changes being proposed there are even more broken. Look at some of the the closed pull reqs - (e.g. recent ones in atomspace)  they all have only two failing tests: PythonModules and Anaphora, both failing on missing Agents code (#3500 #3503 and #3509) ... This would be more clear if we had an actual python maintainer to keep that code up-to-date.  We need a python maintainer.

ferrouswheel:
I'm happy to look at those, and the python stuff does look like it needs major tidy up and someone to maintain it. And possibly that could be me in future.

... but that is all a separate issue to this PR.

leungmanhin:
I think the `OpenPsiTest` is a new test that's introduced two days ago when https://github.com/opencog/opencog/pull/3501 got merged?
The usual ones are `OpenPsiRulesUTest`, `OpenPsiImplicatorUTest`, and `OpenPsiSCMUTest` and they all pass.

linas:
Ahh. OK. Based on this, I'm merging. However, if #3501 introduces a new failing unit test, it should not have been merged...

PR #3517, opened 2019-06-05
Add some new methods to OpenPsi Python wrapper

stellarspot:
Add create_goal, create_component, set_action_selector and other methods to OpenPsi Python wrapper.


stellarspot:
From the error log I see message:
```
29/39 Test #30: OpenPsiTest ......................***Failed    0.20 sec
Failure: ImportError (No module named 'opencog.type_constructors') ... ERROR
```
It looks like the test can't find atomspace.so in the Python library dist-packages.

linas:
well, but OpenPsiUTest does pass, before your changes!?

linas:
Also -- can you take a look at the #3500 -- #3503 -- #3509 bug complex? Python got changed around, it looks like the python agents code got broken, and now, unit tests aren't passing ... this is becoming an urgent bug, we need to either fix Agents, or remove them, or something?

ferrouswheel:
#3518 Should fix the ImportError - someone messed up the test `PYTHONPATH`.

stellarspot:
I run the OpenPsiTest with python3 and using make test.
https://github.com/opencog/opencog/blob/master/tests/cython/openpsi/openpsi_test.py

The test passes when python3 is used because OpenPsiRules is created only once  so the default component is registered in OpenPsi:
https://github.com/opencog/opencog/blob/ded29ec1a9a627188b895f7a11da2893c1ab6adb/opencog/openpsi/main.scm#L35

The test fails when make test is used because OpenPsiRules is created twice.
In first time the default component is stored in OpenPsiRules _category_index.
The second time OpenPsiRules is created one more time with empty _category_index where the default component is not registered. That is why OpenPsi does not start its loop.

stellarspot:
The problem is that the fist time the OpenPsiTest interacts with OpenPsi only using scheme and scheme loads libopenpsi.so from the system path: `/usr/local/lib/opencog/libopenpsi.so`

The second time the test calls OpenPsi API directly using C++ and in this case the  libopenpsi.so is used from local build directory: `build/opencog/openpsi/libopenpsi.so`

The solution can be to pass for tests a Scheme env variable to make Scheme using  libopenpsi.so from local build directory.

stellarspot:
I tried to add LD_LIBRARY_PATH to the OpenPsiTest environment variable and found the following things:
1) Using SET_TESTS_PROPERTIES with several environment variables actually stores only the last one.
For example:
```cmake
	ADD_TEST(OpenPsiTest ${NOSETESTS_EXECUTABLE} -vs
		${CMAKE_SOURCE_DIR}/tests/cython/openpsi/)
	SET_TESTS_PROPERTIES(OpenPsiTest
		PROPERTIES
		ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_BINARY_DIR}/opencog/openpsi"
		ENVIRONMENT "PYTHONPATH=${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_BINARY_DIR}/opencog/openpsi:${PYTHON_ROOT}")
```

The file tests/cython/CTestTestfile.cmake contains generated code:
```cmake
add_test(OpenPsiTest "/usr/bin/nosetests3" "-vs" "/home/stellarspot/opencog/stellarspot/opencog/tests/cython/openpsi/")
set_tests_properties(OpenPsiTest PROPERTIES  ENVIRONMENT "PYTHONPATH=/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/cython:/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/openpsi:/usr/local/lib/python3/dist-packages")
```

The LD_LIBRARY_PATH env variable is missed.

There is the discussions that it is the known issue:
https://cmake.org/pipermail/cmake-developers/2013-June/019219.html

I tried the suggested workaround to use a semicolon to separate env variables but it also does not help in my case:
```cmake
	SET_TESTS_PROPERTIES(OpenPsiTest
		PROPERTIES
		ENVIRONMENT "LD_LIBRARY_PATH=${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_BINARY_DIR}/opencog/openpsi;PYTHONPATH=${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_BINARY_DIR}/opencog/openpsi:${PYTHON_ROOT}")
```
The generated tests/cython/CTestTestfile.cmake file:
```cmake
set_tests_properties(OpenPsiTest PROPERTIES  ENVIRONMENT "LD_LIBRARY_PATH=/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/cython:/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/openpsi;PYTHONPATH=/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/cython:/home/stellarspot/opencog/stellarspot/opencog/build-docker/opencog/openpsi:/usr/local/lib/python3/dist-packages")
```

Setting LD_LIBRARY_PATH in this way leads that Scheme still  loads libopenpsi.so from  `/usr/local/lib/opencog/libopenpsi.so` and not from `build/opencog/openpsi/libopenpsi.so`.

stellarspot:
The LTDL_LIBRARY_PATH is set to the OpenPsiCythonTest env variables to make Guile load *.so libraries from build dir.

Now the OpenPsiCythonTest test fails with error: `Fatal Python error: take_gil: NULL tstate`
https://circleci.com/gh/opencog/opencog/2093

This is known issue in Python3 which has been fixedin Python 3.6.
https://bugs.python.org/issue20891
The proposed solution is to Python 3.6 or higher in the CI docker build. 

Eventhough the problem with Python version is fixed the test sill fails because of the known issue: https://github.com/opencog/opencog/issues/3492
```
terminate called after throwing an instance of 'opencog::RuntimeException'
  what():  AtomTable - deleteing atomtable with subtables! (/home/stellarspot/opencog/master/atomspace/opencog/atomspace/AtomTable.cc:128)
Aborted (core dumped)
```

I propose to disable the OpenPsiCythonTest test and enable it again when both these solutions are fixed.

stellarspot:
I divided the OpenPsiCythonTest on two parts. The first  checks OpenPsi rule creations and passes.
The second checks how OpenPsi is running. It is disabled.

Now only the following tests fail in the Circle CI:
```text
The following tests FAILED:
	 17 - AnaphoraTest (OTHER_FAULT)
	 19 - SuRealUTest (Failed)
	 20 - MicroplanningUTest (Failed)
```

linas:
Anaphora is failing due to a garbage-collection overflow, probably triggered by an unusual python/scheme mix. See discussion on #3509 The work-around is to install a built-from-source garbage collector configured to handle "large memory". The test will probably have to be disabled for now.

@leungmanhin will look at SuReal and MicroPlanning again see #3509

ferrouswheel:
Discussion of Anaphora and GC error starts [here](https://github.com/opencog/opencog/issues/3529#issuecomment-500268143) (rather than the issue Linas linked to)

tests/cython/openpsi/openpsi_test.py
```diff
@@ -16,22 +18,24 @@ def eat_apple(apple):
     return ConceptNode("finished")
 
 
+__main__.eat_apple = eat_apple
```
ferrouswheel: Is this necessary to make Scheme see `eat_apple`?

tests/cython/openpsi/openpsi_test.py
```diff
@@ -69,14 +72,40 @@ def testopenpsi(self):
         self.assertTrue(new_category in categories)
 
         self.assertEqual(context, rule.get_context())
+        self.assertEqual(action, rule.get_action())
+
+    @unittest.skip("Issue https://github.com/opencog/opencog/issues/3492")
```
ferrouswheel: Thanks for documenting why the test is skipped :+1:

tests/cython/openpsi/openpsi_test.py
stellarspot: Yes. Here is the error message which is printed when the eat_apple is not defined in the main module:
`In procedure psi-imply: Python function 'eat_apple' not found in module '__main__'!`

stellarspot:
> The only thing I would change is: instead of calling `init_scm` in every method, I would add a method `OpenPsi.scheme_eval` that does the the initialisation check before calling the main `scheme_eval`. Then in each method you'd call `self.scheme_eval(...)`.

Yes. I updated the code to initialize the Scheme in OpenPsi constructor. 
The `scheme_eval(atomspace, ...)` will be called each time OpenPsi is created.
There is a question should it be done only once. It seems reasonable that it should be called oly once for each provided atomspace. So if OpenPsi is created with two different AtomSpaces the initialization should be called twice for each AtomSpace.

However I not sure that OpenPsi can work with several AtomSpaces. 
The OpenPsi initialization in file https://github.com/opencog/opencog/blob/74d4d402b7fff31efe6bf15d601b5228ead0d558/opencog/openpsi/main.scm#L35 puts main component into global categories list and it seems that it is initialized only once.

AmeBel:
OpenPsi doesn't work with multiple atomspaces

PR #3518, opened 2019-06-06
Python module fixes

ferrouswheel:
Fix for #3509

There was also some bad `PYTHONPATH` mangling which would have seriously messed up any python development: *Remember* the sys.path is ordered. If you put a system install dir before all the local ones, then python won't see the equivalent modules in your dev directories.

Includes a change to the pyx cmake rule to fix dependencies... for files without an equivalent `pyd` suffix, make wouldn't detect changes.

AnaphoraTest and OpenPsiTest still fail, but they give sensible output now instead of failing to find stuff or crashing.

I have not tested with Python 2. I'm not sure if that is something we still care about?


opencog/attentionbank/cython/opencog/CMakeLists.txt
```diff
@@ -22,10 +22,9 @@ SET(CYTHON_FLAGS "-f")
 # Cythonizes one .pyx file into a .cpp file
 # Additional arguments are dependencies
 MACRO(CYTHON_ADD_MODULE_PYX name)
```
linas: This and the below seem to be cut-n-pastes of one-another. This macro should almost surely be moved to the atomspace repo, `/cmake/OpenCogFunctions.cmake` which is where the macro for guile modules lives. (or `/cmake/OpenCogCython.cmake` or something like that. Maybe rename `OpenCogFunctions.cmake` to `OpenCogGuile.cmake` too ... not now but in some later fix.

linas:
Whoops!, I should not have been so quick with that merge button.  I'm too eager to please. :-)

The circle-ci build is now broken, with 
```

[ 18%] Cythonizing agent_finder.pyx

Error compiling Cython file:
------------------------------------------------------------
...
        vector[bool] req_is_shell
        string err_string 

cdef api requests_and_agents_t load_req_agent_module(string& module_name) with gil:
    """ Load module and return a vector of MindAgent names """
    cdef str mod_name = module_name.decode("UTF-8")
                                  ^
------------------------------------------------------------

/ws/opencog/opencog/cython/opencog/agent_finder.pyx:33:35: Object of type 'string &' has no attribute 'decode'
[ 18%] Linking CXX shared library libspacetime-types.so
opencog/cython/opencog/CMakeFiles/agent_finder.dir/build.make:65: recipe for target 'opencog/cython/opencog/agent_finder.cpp' failed
```

this pull req should probably be reverted until this is fixed; but it's past my bedtime so I am not reverting ...

ferrouswheel:
hrm, I verified it worked locally.

Immediate thoughts are it is a difference in either:
- Python version, I'm using Python 3.6.7
- Cython version, I'm using 0.26.1

I can look into the circleci config tomorrow.

opencog/attentionbank/cython/opencog/CMakeLists.txt
ferrouswheel: I thought the same thing. Thanks for the tip for where it should be placed.

ferrouswheel:
Yeah, circleci uses Cython 0.23.4 and Python 3.5.
It'd be nice if we could bump those to standard versions in Ubuntu 18.04 LTS.

vsbogd:
May be I am "Captain Obvious" but merging PR which breaks CI build is not a good practice.

ferrouswheel:
The CI build appears to always be broken due to test failures, so until we fix it it's always going to be non-obvious when a PR impacts the build.

vsbogd:
Usually one can at least check that number of failing tests is not changed but this time build was totally broken.

linas:
> Captain Obvious

Yeah, it was late at night and I forgot to check, and since tests have been failing I got used to seeing red X's.

PR #3521, opened 2019-06-07
Rely on common CYTHON_ADD_MODULE_PYX from atomspace cmake includes

ferrouswheel:
Based on comment here: https://github.com/opencog/opencog/pull/3518/files#r291023552

Requires opencog/atomspace#2207


linas:
Thanks!

Issue #2985, opened 2018-01-25
AnaphoraTest fail on clean ubuntu 16.04 (originally #2793)


mjsduncan:
```
8:AnaphoraTest

;;; compiled /home/mjsd/oc/guile/ccache/2.2-LE-8-3.A/home/mjsd/oc/opencog/tests/nlp/anaphora/data/pleonastic_it/data_#4.scm.go
;;; compiling FAIL

======================================================================
FAIL: test_conjunctions (test_anaphora.AnaphoraUnitTester)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mjsd/oc/opencog/tests/nlp/anaphora/test_anaphora.py", line 430, in test_conjunctions
    self.assertFalse(self.hobbsAgent.pleonastic_it(self.getWord('waitresses')))
AssertionError: True is not false

======================================================================
FAIL: test_pleonastic_if (test_anaphora.AnaphoraUnitTester)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mjsd/oc/opencog/tests/nlp/anaphora/test_anaphora.py", line 417, in test_pleonastic_if
    self.assertFalse(self.hobbsAgent.pleonastic_it(self.getWord('it')))
AssertionError: True is not false

======================================================================
FAIL: Testing the propose function
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mjsd/oc/opencog/tests/nlp/anaphora/test_anaphora.py", line 378, in test_propose
    filter_1()
  File "/home/mjsd/oc/opencog/tests/nlp/anaphora/test_anaphora.py", line 94, in filter_1
    self.assertTrue(self.hobbsAgent.propose(self.getWord('antecedent')))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 5 tests in 9.889s

FAILED (failures=3)
accepted 
[(WordInstanceNode "b") ; [7176733813256336297][1]
, (WordInstanceNode "b") ; [7176733813256336297][1]
, (WordInstanceNode "c") ; [1954536943936637625][1]
]
accepted 
[(WordInstanceNode "c") ; [1954536943936637625][1]
, (WordInstanceNode "b") ; [7176733813256336297][1]
, (WordInstanceNode "c") ; [1954536943936637625][1]
, (WordInstanceNode "d") ; [7920996617400663969][1]
, (WordInstanceNode "e") ; [811631232188714808][1]
[...]

PR #3524, opened 2019-06-07
Skip test for broken HobbsAgent code

vsbogd:
See analysis at: https://github.com/opencog/opencog/issues/2985#issuecomment-499839361
I propose skipping theese unit tests as code needs to be reworked to fix the issue.


linas:
thanks!

PR #3527, opened 2019-06-07
Actually fix the anaphora code

linas:
This actually fixes the assorted "obvious" bugs in the code; the unit test now passes.

The original code author had the right idea, and mostly had everything working right, but strangely misunderstood the bindlink.


linas:
Merged .. AnaphoraUTestnow passes!

PR #3535, opened 2019-06-12
Workaround for distutils bug in Ubuntu/Debian's Python3 versions

ferrouswheel:
To be paired with opencog/atomspace#2220 - which has additional context.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,7 +345,7 @@ IF(HAVE_PY_INTERP)
 		# Find python destination dir for python bindings
 		# because it may differ on each operating system.
 		EXECUTE_PROCESS(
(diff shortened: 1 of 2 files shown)


linas:
Given that there is a bi block of cut-n-paste code both here and in the atomspace, the best long-ter solution is to create a `cogutils/cmake/FindPython.cmake` that extracts all of that logic.   In the meanwhile, this can be merged as is, I suppose.

Why `cogutils` and not `atomspace`? No reason - just a historical accident. There are several dozen FindStuff.cmake files there; notable by its absence is python.

ferrouswheel:
Makes sense, there's is a bunch of scaffold code in the root CMakeLists.txt for Python detection that could also be moved (tho that may have been what you were also referring to)

PR #3533, opened 2019-06-11
Remove old pattern miner

ngeiswei:
The new pattern miner seems mature enough to replace it.

Tag of the old pattern miner: https://github.com/opencog/opencog/releases/tag/old-pattern-miner-11-Jun-2019

PR #3536, opened 2019-06-12
some remnants from pattern miner removal in #3533

ferrouswheel:


linas:
The new patternminer does not need custom types any more?

ferrouswheel:
@ngeiswei will have to confirm that. This just fixes the broken symbols, so that tests "work" again - i.e. are in the state they were in before #3533.

linas:
actually, this is a variant of the spacetime atoms bug - even if the new patternminer did need new atom types, this would not be the right place to initialize them,

ngeiswei:
The new pattern miner doesn't need new atom types, instead it relies on scoping and quotation for reflection, discovered patterns are essentially pattern matcher queries.

PR #3547, opened 2019-06-18
Remove my ubigraph example

ferrouswheel:
Ubigraph can't be downloaded anymore so this is just noise.

Surprised it was still around :laughing:


linas:
Unfortunate.

Issue #3551, opened 2019-06-24
SuRealUTest compilation error


ngeiswei:
I'm getting the following error with the last opencog 
```
[ 70%] Linking CXX executable SuRealUTest
/usr/bin/ld: ../../../opencog/nlp/lg-dict/liblg-dict-entry.so: undefined reference to `boolean_dictionary_lookup'
collect2: error: ld returned 1 exit status
```

ngeiswei:
In case it is related, my version of link grammar is (according to cmake)
```
-- Link Grammar (5.5.1 >= 5.4.0) found.
```

ngeiswei:
`boolean_dictionary_lookup` is an old function so the problem likely doesn't come from a link grammar version.

linas:
It's an old function, but apparently, it was never specified in the shared-library export list. Now, normally, this list only affects MacOS users; the glibc shared library system seemed to ignore it.   .. at least for me. Maybe something changed. I'm adding it to the export list now.

linas:
Oh, thaat's interesting .... `nm /usr/local/lib/liblink-grammar.so` shows that, indeed, only the exported symbols are in the shared lib. whereas `nm /usr/local/lib/liblink-grammar.a` shows everything.  I have no clue why your system is behaving differently than mine.  A different loader version, maybe?  What does `ldd --version` show?

Mine shows
```
ldd (Debian GLIBC 2.24-11+deb9u4) 2.24
```

linas:
The correct solution seems to be to publish an LH version 5.6.2 ASAP and export this and a few other missing symbols.

ngeiswei:
My loader is
```
ldd (Ubuntu GLIBC 2.29-0ubuntu2) 2.29
```
My OS is Ubuntu Studio 19.04.

linas:
I will publish link-grammar version 5.6.2 later today, which should fix this.  If you want to try this at home, you can edit `./link-grammar/link-grammar.def` and add `boolean_dictionary_lookup` anywhere.

linas:
New version here: https://www.abisource.com/downloads/link-grammar/5.6.2/link-grammar-5.6.2.tar.gz should fix this.

ngeiswei:
It works, thanks!

ngeiswei:
BTW, CMake still only requires `5.4.0` while I suppose it should require `5.6.2`, correct?

linas:
[...]

PR #3553, opened 2019-06-25
Bugfix for SuReal, Microplanning

### LINKED ISSUE #3544: MicroplanningUTest and SuRealUTest fail
From the test log, looks like SuReal fails to even generate a very simple sentence, and it fails when it's trying to match the disjuncts between two words, e.g. he <-> she, runs <-> eats as in the first test, which is very strange... and it has started failing when I switch to using Link Grammar 5.6.1.

@linas is there any change in 5.6.1 that may give words a different set of disjuncts or something like that?

BTW since the Microplanner uses SuReal, I believe fixing SuRealUTest will also "fix" the MicroplanningUTest.

linas:
It appars that the loop to read LG expressions was over-complex.
With this change, SuReal and MicroPlanning pass for LG version 5.6.1
I did not test earlier versions of LG.

Fixes #3544

### DIFF
--- a/opencog/nlp/lg-dict/LGDictReader.cc
+++ b/opencog/nlp/lg-dict/LGDictReader.cc
@@ -41,33 +41,13 @@ static LGDictExpContainer lg_exp_to_container(Exp* exp)
     if (CONNECTOR_type == exp->type)
         return LGDictExpContainer(CONNECTOR_type, exp);
 
-    // Whenever a null appears in an OR-list, it means the
-    // entire OR-list is optional.  A null can never appear
-    // in an AND-list.
     E_list* el = exp->u.l;
-
-    if (NULL == el)
-        return LGDictExpContainer(CONNECTOR_type, NULL);
-
-    // The C data structure that link-grammar uses for connector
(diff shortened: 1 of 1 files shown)


linas:
@stellarspot could you test this with earlier versions of LG? I'm 99% certain that this should work, but cannot say for certain.

linas:
ah, what the heck. all unit tests passed.  I'm merging. If something fancier is needed for older LG versions, then ... a distinct patch is needed.

stellarspot:
Tests MicroplanningUTest and SuRealUTest pass with LinkGrammar 5.6.0 and 5.6.2.

With LinkGrammar 5.5.1 SuRealUTest fails with error described in #3551
`../../../opencog/nlp/lg-dict/liblg-dict-entry.so: undefined reference to 'boolean_dictionary_lookup'`

linas:
> 'boolean_dictionary_lookup'

Yeah, that's a screw-up. It was a public API function that was missing a declaration to export it to the shlib.  For some (unkonw-to-me) reason, my system doesn't trip on this, but others do.

PR #3554, opened 2019-06-28
Use PresentLink

ngeiswei:
Have PLN and the pattern miner use `PresentLink`.


ngeiswei:
Wow, everything passes!!!

PR #3557, opened 2019-07-03
Merge opencog to singnet

stellarspot:


linas:
I think you have this going in the wrong repo; merge goes in other direction!?

stellarspot:
Thank you. It is wrong repository.

PR #3559, opened 2019-07-04
Use cog-execute! instead of cog-bind-af

ngeiswei:
since attention focus is irrelevant for these unit tests.


ngeiswei:
```
The following tests FAILED:
	 36 - GhostSyntaxUTest (Failed)
	 37 - GhostProcedureUTest (Failed)
	 38 - GhostUTest (Failed)
```

But it's certainly unrelated to that change.

PR #3565, opened 2019-07-06
Remove the remote-tracking submodule "learn"

linas:
It probably should never have been a submodules in the first place.

FYI @glicerico I believe this will not affect you, but please double-check and let me know.


glicerico:
@linas , this didn't cause any problems on my side

vsbogd:
@glicerico , could you please clarify, should `.gitmodules` file be removed from `singnet/opencog` as well?

linas:
the only reason it was put there as a module was to provide a stable location for the PDF URL's. But the PDF were never handed out with a singnet URL.

PR #3567, opened 2019-07-08
Link to the URE libraries

ngeiswei:
Sister PR of https://github.com/opencog/atomspace/pull/2256


ngeiswei:
The dependence on the ure is optional, which is why all checks pass even though circleci doesn't build the ure yet. Upgrading circleci is coming next.

PR #3570, opened 2019-07-09
Condition sureal and microplanning unit tests on HAVE_NLP

ngeiswei:
Based on

https://github.com/opencog/opencog/blob/1b3b4b02288d5be868246a2e429d5c9aa635d870/opencog/nlp/CMakeLists.txt#L16-L27

it's clear that the unit tests of `sureal` and `microplanning` should only be enabled when `HAVE_NLP` is true.


ngeiswei:
I'm merging because I'm close to certain that is correct.

PR #3571, opened 2019-07-09
Condition ghost on HAVE_NLP

ngeiswei:
According to

https://github.com/opencog/opencog/blob/cc3988ecfc01308c4741e91c32c1c5c235c20ebe/opencog/ghost/procedures/focus-set.scm#L5

it is clear that `ghost` relies on `nlp`.


ngeiswei:
It could perhaps be refined (submodules of ghost may rely on nlp while others not), but it shouldn't be incorrect nevertheless, so I'm merging.

PR #3572, opened 2019-07-09
Define HAVE_OPENPSI

ngeiswei:
Introduced because OpenPsi relies URE.


ngeiswei:
Turns out all openpsi unit tests pass without the ure. That doesn't invalidate that PR, it may still be useful to have `HAVE_OPENPSI`, but that's not that important.

CMakeLists.txt
```diff
@@ -688,5 +692,6 @@ SUMMARY_ADD("NLP" "Natural Language Processing" HAVE_NLP)
 SUMMARY_ADD("Haskell codes" "Logic via Lojban" HAVE_STACK AND BUILD_LOJBAN)
 SUMMARY_ADD("PLN" "Probabilistic Logic Network" HAVE_PLN)
 SUMMARY_ADD("Miner" "URE based hypergraph pattern miner" HAVE_MINER)
+SUMMARY_ADD("OpenPsi" "MicroPsi OpenCog implementation" HAVE_OPENPSI)
```
linas: its' nt really micropsi any more

PR #3574, opened 2019-07-11
Update README.md with URE as optional prerequisite

ngeiswei:


ngeiswei:
Damn it, circleci fails for opencog with the same `The specified paths did not match any files in /ws` error. I guess we probably just need to disable persistence for now, it's probably gonna slow down the thing but it should still be correct. Maybe an error on circleci side indeed.

PR #3575, opened 2019-07-11
Checkout URE in circleci

ngeiswei:


ngeiswei:
Fails because `The specified paths did not match any files in /ws`, works otherwise, so I'm merging.

PR #3580, opened 2019-07-12
fix URE-related regression.

linas:
NLP does not require URE!

This fixes one of the failures found in https://github.com/opencog/atomspace/issues/2267


linas:
tests pass, except for `The specified paths did not match any files in /ws` .. merging, this is a blocking bug.

ngeiswei:
I'm not so sure, some nlp code definitely imports the `ure` module. See

https://github.com/opencog/opencog/blob/ac48647239e62de0419ea7e9f62eb90f75d4e7b3/opencog/nlp/sentiment/loadSentiList.scm
https://github.com/opencog/opencog/blob/ee4bca0d198be503f9cdbbff9b2d38402d86eb52/opencog/nlp/relex2logic/relex2logic.scm
https://github.com/opencog/opencog/blob/ac48647239e62de0419ea7e9f62eb90f75d4e7b3/opencog/nlp/chatbot/chat-utils.scm
https://github.com/opencog/opencog/blob/ac48647239e62de0419ea7e9f62eb90f75d4e7b3/opencog/nlp/relex2logic/loader/gen-r2l-en-rulebase.scm

linas:
OK, so those things that depend on URE need to be fixed; right now, I'm getting multiple hard system failures. 

Anyway, its highly unlikely that any of those things actually require URE, someone just propagates some boilerplate code via cut-n-paste.

ngeiswei:
I think it's more than cut-n-paste boilerplate, but @leungmanhin should be able to tell.

I'm getting no hard failure, maybe you need to purge you system from opencog, for that I'm using

https://github.com/ngeiswei/ocbld/blob/19e74231a8e412aba02c6b69216d28da5271bbfe/ocbld#L73-L84

linas:
I purged my system. That is what provoked the hard fail!

ngeiswei:
How did you purge it? What did you do differently than my purging code?

linas:
seriously, relex2logic is unrelated to URE, it only needs relex. The chatbots never needed URE. Link-grammar does not need URE... and the sentiment analysis code should never have been committed to the repo. Anyway, the sentiment code is a python wrapper for some 3rd-party system.

linas:
I just manually `rm -r /usr/local/share/opencog` and also `lib`, `include`

ngeiswei:
I did that as well... Weird...

linas:
.. not weird at all, you have a different config than I do. I've got half-a-dozen different containers running language-learning code on ubuntu 16.04 ... you don't ... you would not have my scripts to launch and run the various jobs .. that are now failing, ... but I think I found the needed fixes

linas:
This is kind-of the generic issue with "stable" vs. "unstable" code.  I made the mistake of pulling the newest atomspace (well, I needed to, cause it has code I need), plus also doing the make install-clean step (which I did not need to do) The combination of these two converted a stable system into a failing system.   But we do no have system acceptance tests on opencog so ... that's to be expected, I guess.

AmeBel:
relex2logic uses ure and the chatbot uses relex2logic

linas:
A simple grep shows that relex2logic does NOT use ure:
```
grep -r forwar *
grep -r chain *
grep -r ure *sumably incorrect.
```
These yield zero hits, except for the README, which is presumably incorrect.  This is hardly a surprise -- there is no reason at all for relex2logic to actually use the URE for anything; it's an assortment of BindLinks applied to the relex output to rewrite them into a different form. There's no chaining, anywhere.

AmeBel:
The place it is used at
https://github.com/opencog/opencog/blob/5facc3b24a29a2919601cce593b5bce3f426b228/opencog/nlp/relex2logic/relex2logic.scm#L57 

You are right, cog-bind can be used as there is no chaining. One of the reason `cog-fc` was used at the time was for restricting the application of the rules to the atoms to a subset of atoms that satisfy the rules, through the foucs-set, rather than all atoms.

linas:
Oh, huh... that's ... unexpected.

First, the whole point of PARSE-NODE is that it *is* the defacto focus-set -- BindLink will never-ever search "all atoms" -- it merely pretends to; its smarter than that.  Also, the Relex output was explicitly designed so that it was well-connected, making it efficient for the pattern matcher to crawl it.  So there is no need for a "focus set", at least for this: -- there is always a built-in focus-set in the pattern mathcher, and the SentNode/ParseNode thing in both RelEx and Link-grammar are also "focus sets".  So somehow, someone never quite understood that.  Oh well. Too late ...

But then there is something I don't understand ... about the forward chainer. Normally, the way chaining works is that you have a set of premises to start with, and a set of rules, and a target that you want to reach.   With forward-chaining, you apply rules to the premises, over and over and over, hoping to stumble onto the target. If you do, you are done, else loop again.  In backwards chaining, you start at the target, apply rules in reverse, and hope to eventually hit one of the premises.  Backwards-chaing is usually more efficient, because there is only one target, and you might hit any of the premises.  Forward-chaining is usually very inefficient -- there's more to explore, and a smaller target to hit.

In relex2logic, what is the target?    Are there a bunch of targets, and you keep forward-chaining  until you hit all of them? Why not use the backward-chainer, as its usually more efficient? Or do you have more targets than premises?

AmeBel:
If I remember correctly, there isn't any mult-step forward chainging. Just one step rule application within the focus set. No backward chaining because it was being developed.

PR #3584, opened 2019-07-15
Reduce cogserver thrashing by reducing open socket count.

linas:


linas:
Fails with `The specified paths did not match any files in /ws`  merging anyway.

Issue #3391, opened 2018-12-20
Split opencog into multiple github repos!


linas:
This is an issue open for debate: there are various pros and cons. The proposal:

Split this repo into multiple parts. These would be:

* pln (*done* see https://github.com/opencog/pln #3622)
* attention allocation (*done* see https://github.com/opencog/attention #3610)
* spacetime server (*done* see https://github.com/opencog/spacetime #3593 )
* cogserver; mind-agents (*done* see https://github.com/opencog/cogserver #3608)
* pattern miner (*done* see https://github.com/opencog/miner #3586 )
* language-learning  (*done*, see https://github.com/opencog/learn #3444)
* language - link-grammar subsystem  (*done*, see https://github.com/opencog/lg-atomese #3670)
* atomspace-publisher (*done*, see https://github.com/opencog/atomspace-restful #3649)
* pattern-index (*done*, see  https://github.com/opencog/pattern-index #3595)
* visualization (*done*, see https://github.com/opencog/visualization #3598)
* python-attic  (*done*, see https://github.com/opencog/python-attic #3648)

What's left to be done:
* ghost
* openpsi
* relex2logic
* chatbots
* robot-embodiment

The above are fairly tightly knit together in the current code base, so it might not be possible to split them up cleanly into distinct components. So maybe we can leave these, as they are!?

Pros:
* It makes it easier to find interesting bits, and see how they fit together.
* Work gets focused on the component that matters, without distraction from other component maintainers
* Issue list easier to track.
* Simpler git history
* Less overall complexity
* Different repos can have different owners, different merge policies.
* Developers in one repo don't have to be experts in the other repos.
* Test breakages in pre-alpha components don't affect mature components

Cons:
* Its gonna take work to split things apart.
* Some dependencies might get complicated.
* API changes in one module (e.g. openpsi) trickle down into ghost, and thus require coordination.

If you have new pros & cons to add to this list, please edit this first top-post.  Other comments, including proposals for splitting along different directions, should go into comments below.  I expect this to be somewhat controversial. It seems like a good idea, just right now.

ngeiswei:
I think it's a good idea.

The main con is that it's a lot of work, but I suppose it could be done iteratively, taking opencog apart piece by piece. The task could be distributed across the maintainers of each part, for instance I would be in charge of moving the pattern miner and pln to their own repositories, @misgeatgit would be in charge of moving attention allocation, etc.

Worth considering at the very least...

ngeiswei:
[...]

PR #3586, opened 2019-07-16
Move pattern miner to its own repo

ngeiswei:
Continuing with https://github.com/opencog/opencog/issues/3391

Miner repo is ready https://github.com/opencog/miner

Beside having to build and install the pattern miner from the miner repo, this should not introduce any backward compatibility issues.


ngeiswei:
I'm merging as I do not expect disagreement on that one.

PR #3587, opened 2019-07-16
Remove an old and apparently unused library

linas:
From what I can tell, there are no users of this code. 
Nor does it appear to do anything special or interesting...


linas:
Failing with the same-old `The specified paths did not match any files in /ws` error. Can't someone fix this?

Merging.

PR #3589, opened 2019-07-16
Avoid test failures if URE is not installed.

linas:
Also remove the long-dead viterbi test.


linas:
All tests pass, but still fails with `The specified paths did not match any files in /ws` .. merging anyway.

PR #3593, opened 2019-07-16
Move spacetime code over to it's own directory

linas:
It is now located in https://github.com/opencog/spacetime


linas:
This is per enhancement req #3391

linas:
Everything passes, except for `The specified paths did not match any files in /ws`

Merging, as I do not believe this is a contentious change.

ngeiswei:
Your spacetime repo has much less commits than my split repos, I assume what you did is to move the spacetime code in a single directory then use
```
git filter-branch --subdirectory-filter <SPACETIME_DIR>
```
Right?

I thought of doing that but I didn't like the idea of introducing commits moving the files, but maybe it's a stupid reason.

linas:
Yes, more or less, I used the instructions at https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository but then I did have to move directories.  Then I realized I forgot about the tests dir so added that separately.  It might have been possible to grab both at once, but I did not try.

These instructions: http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history I did not look at, but maybe I should have.

> much less commits

Yes, I like that, but can't give a strong reason for why.  Just a bunch of little reasons: the history is clearer. Who contributed is clearer.

linas:
@ngeiswei It's not too late to redo URE with `filter-branch` -- now is the time to do it; it won't be possible later. I would encourage this, its not hard. Along the way rename the github repo RUE to make room, and so that you can copy the final adjusted makefiles into place. (which, for me, was the hard part).

ngeiswei:
@linas , I tried, but it squashes the git history to 1 commit, replacing the moving commit into a commit creating all the files.

Makes me wonder if what you did might have deleted some git history pertaining to the files of interest.

Dagiopia:
as it currently exists, history of spacetime isn't preserved. Git blame shows @linas because of commit:  https://github.com/opencog/spacetime/commit/929337b9dc7896aec50d2b1e88bd3bd106402c20

linas:
argh.... you are right. I am now reading https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history and I see several interesting comments. One that stands out is that its a design flaw in git ....

linas:
So, after reading the stackoverflow article, I can answer this question:

> deleted some git history

Apparently not, since it is visible in (for example) `git log --follow OctoMapNode.cc`  and so the issue is that github does not use the `--follow` flag when it displays the blame and history of a file.

However, when I try `git log --follow PointMemorySCM.cc` it works .. but only up to an earlier move.  The fancier 
`git log --follow --find-copies-harder PointMemorySCM.cc` does not break past this barrier.

linas:
> deleted some git history

Apparently so ... `git log --follow PointMemorySCM.cc` works only up to the earlier move because ... `man git-filter-branch` states: 
```
       --subdirectory-filter <directory>
           Only look at the history which touches the given subdirectory. The
           result will contain that directory (and only that) as its project
           root. Implies the section called “Remap to ancestor”.
```

and so all the history of the file, while it was in `spacetime` was preserved. The history of what happened, before it got moved to `spacetime` is lost. 

@ngeiswei -- we should create a policy for this. Either we decide that using `git filter-branch --subdirectory-filter` is OK (and the user is expected to go back to the other repo, if they want more) or we decide that its not OK, and that instead, we should fork the whole repo, and delete unwanted files. 

What we can't fix is github itself, forcing it to use `git log --follow` instead of plain `git log` for a file.  The other thing we can't fix is git itself, which should have tracked the file moves, but didn't.

It is not too late to redo the git repos I just created ... spacetime, and the others ... they can be redone now, but if we wait, it will become impossible...  I don't think I care much one way or the other.  Other than that I'm a bit lazy ...

ngeiswei:
Thanks for the clarification, @linas . I don't care much one way or the other either, I suppose the author of the respective repos should speak up.

linas:
> the authors of the respective repos should speak up.

The authors of dimensional reduction haven't been active in a decade, literally.  spacetime has been re-written 4 times, the last one by @yantrabuddhi  who hasn't been active in 3+ years.  All these directories suffer from being abandon-ware -- created, original author gone, zero users. We need to find a mode of creating things that people find useful ...

PR #3594, opened 2019-07-16
Remove another file that was marked obsolete many years ago.

linas:
Seems to be unused (no one should be using it by now).


linas:
The CircleCI seems to be running tests from some cached directory; the errors reference line numbers and code that simply does not exist any longer. I believe that merging this should resolve those issues ... I I'll just merge.

This was supposed to be a trivial removal of an obsolete, unused file, and it has snowballed into something else.

PR #3595, opened 2019-07-17
Move files to new repo: opencog/pattern-index

linas:
See https://github.com/opencog/pattern-index/


linas:
Build is failing, but not due to this pull req. I think I know the failure cause, and will fix in #3594

linas:

PR #3596, opened 2019-07-17
Remove unused eigenvalue solver.

linas:
This code was never used.  There is no point in carrying it around.


linas:

PR #3605, opened 2019-08-05
update Atom.createAtom calls to new api

noskill:
api was changed in https://github.com/opencog/atomspace/pull/2313


opencog/attentionbank/cython/opencog/bank.pyx
```diff
@@ -9,7 +9,7 @@ cdef vector_to_set(vector[cHandle] handle_vector, AtomSpace atomspace):
     atoms = set()
     while it != handle_vector.end():
         c_handle = deref(it)
-        atoms.add(Atom.createAtom(c_handle, atomspace))
+        atoms.add(Atom.createAtom(c_handle))
```
vsbogd: After this change `atomspace` can be removed from `vector_to_set` signature

linas:
thanks

PR #3606, opened 2019-08-30
fix #3550: atomspace object _as gets a new unique UUID get_uuid()

### LINKED ISSUE #3550: opencog::attentionbank(...) should not rely that objects compared by references are equal
opencog::attentionbank(AtomSpace* asp) method in AttentionBank has the following check to compare if two AtomSpaces are equal:

https://github.com/opencog/opencog/blob/802f0ec84e90db55389ee8ace40be6004a1ec9d1/opencog/attentionbank/bank/AttentionBank.cc#L249

This can fail if one AtomSpace is deleted and immediately after that the second one is created.
It can happen the the new AtomSpace has the same memory address as the previous one.

I encountered it in the Python test that that can or can't fail depending with each address the new AtomSpace is created:
(note that the #3549 pull request is required to run the test)
```python

import unittest
from unittest import TestCase

from opencog.type_constructors import *
from opencog.utilities import initialize_opencog, finalize_opencog
from opencog.bank import AttentionBank, af_bindlink


class AttentionBankTest(TestCase):

    def setUp(self):
        self.atomspace = AtomSpace()
        initialize_opencog(self.atomspace)

    def tearDown(self):
        finalize_opencog()
        del self.atomspace

    def test_get_handles_by_av_different_as1(self):
        attention_bank = AttentionBank(self.atomspace)

        node1 = ConceptNode("node1")
        attention_bank.set_sti(node1, 2.0)

        atoms = attention_bank.get_atoms_by_av(1.0, 4.0)
        self.assertEqual(1, len(atoms))

    def test_get_handles_by_av_different_as2(self):
        attention_bank = AttentionBank(self.atomspace)

        node2 = ConceptNode("node2")
        attention_bank.set_sti(node2, 2.0)

        atoms = attention_bank.get_atoms_by_av(1.0, 4.0)
        self.assertEqual(1, len(atoms))


if __name__ == '__main__':
    unittest.main()
```

chandramoulirajagopalan:
Nullpointer Atomspace instance (_as) has now a unique UUID to prevent same object reference.

### DIFF
--- a/opencog/attentionbank/bank/AttentionBank.cc
+++ b/opencog/attentionbank/bank/AttentionBank.cc
@@ -238,6 +238,7 @@ AttentionBank& opencog::attentionbank(AtomSpace* asp)
 {
     static AttentionBank* _instance = nullptr;
     static AtomSpace* _as = nullptr;
+    _as->get_uuid();
 
     // Protect setting and getting against thread races.
     // This is probably not needed.


vsbogd:
@chandramoulirajagopalan, have you tried to run this code? It should lead to segfault.

PR #3607, opened 2019-08-31
fix (3550): attentionbank objects

chandramoulirajagopalan:
Atomspace instance _as has now a unique UUID to prevent same object reference.

### DIFF
--- a/opencog/attentionbank/bank/AttentionBank.cc
+++ b/opencog/attentionbank/bank/AttentionBank.cc
@@ -238,6 +238,7 @@ AttentionBank& opencog::attentionbank(AtomSpace* asp)
 {
     static AttentionBank* _instance = nullptr;
     static AtomSpace* _as = nullptr;
+    _as->get_uuid();
 
     // Protect setting and getting against thread races.
     // This is probably not needed.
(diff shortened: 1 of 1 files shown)


opencog/attentionbank/bank/AttentionBank.cc
```diff
@@ -237,7 +237,8 @@ double AttentionBank::getNormalisedZeroToOneSTI(AttentionValuePtr av,
 AttentionBank& opencog::attentionbank(AtomSpace* asp)
 {
     static AttentionBank* _instance = nullptr;
-    static AtomSpace* _as = nullptr;
+    static AtomSpace* _as;
+    _as->get_uuid();
```
stellarspot: For the first `opencog::attentionbank(AtomSpace* asp)` call the `_as` pointer is initialiazed in some garbage  `_as->get_uuid()` has undefined behavior.

The idea of the fix was to compare `asp` and `_as` by their uuid if they both are not null.

stellarspot:
Could you add the test for the fix as well?

linas:
This makes no sense. Simply calling `get_uuid()` does nothing at all.

PR #3608, opened 2019-09-04
Rm server

linas:
Split out cogserver into it's own git repo. See issue #3391


linas:
@AmeBel and @vsbogd -- please review -- this is a rather large change, and might be disruptive to the various projects you work on.  Merge if you feel your ready; else I'll merge ... sometime later.

README.md
```diff
@@ -105,10 +106,18 @@ be built and run.
...
-> Library for generating UUID's
-> Used by various internal subsystems.
-> `sudo apt-get install uuid-dev`
+###### cogserver
+> OpenCog CogServer Network Server.
+> http://github.com/opencog/cogserver
+> It uses exactly the same build procedure as this package. Be sure
+  to `sudo make install` at the end.
+
+###### attention
```
vsbogd: It is more related to previous work, but it this comment make me check and I see that `CMakeLists.txt` doesn't check that `attention` and `spacetime` server are installed despite here they are described as dependencies. Is it correct?

vsbogd:
General comments:
- `CircleCI` config need to be added to `opencog/cogserver`
- if `cogserver` should be a part of `opencog-deps` docker image then its installation should be added into `opencog/ocpkg` tool

vsbogd:
updated previous comment

linas:
I just now added a circleci config to the cogserver repo. However, I don't know how to (don't have permissions to) turn it on (I'm, umm out of town and don't have my passwords with me) so maybe @AmeBel can turn on circleci testing?

I'll try to update ocpkg shortly.

linas:
I looked at ocpkg. I'm lost. I wanted to use the URE as an example for what to do, but ocpkg does not handle URE...

vsbogd:
As `cogserver` was a part of `opencog`, may be it makes sense just add it as additional line after https://github.com/opencog/ocpkg/blob/9547e79fb0c55e6449dc65800c5ec64c2ab8d2b7/ocpkg#L862

PR #3610, opened 2019-09-11
Remove attention-bank

linas:
The ECAN Economica Attention Allocation code is now in it's own git repo; remove it from this one.  This is pursuant to issue #3391 - split up opencog repo into distinct modular pieces.


linas:
This change affects openspi the most, I think, as the current primary user of attention allocation. Thus, asking for review from @amebel, @leungmanhin, @misgeatgit

(I realized that, for language-learning, I need something a lot like attention allocation, but maybe different in various ways; this split should allow the independent development of generic attention-allocation-type algorithms.)

linas:
I week has gone by; I'm merging.

PR #3613, opened 2019-09-23
Fix typo in README.md

che-shr-cat:
fix typo


vsbogd:
@che-shr-cat, build fails as you used old master to add commit. Please create branch from `master` or rebase to master to make sure that build is green next time.

PR #3617, opened 2019-10-11
Check if AttentionBank is installed, if not disable OpenPsi and NLP

ngeiswei:


CMakeLists.txt
```diff
@@ -474,13 +474,13 @@ IF(HAVE_URE)
...
+IF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_BANK)
 	SET(HAVE_OPENPSI 1)
-ENDIF(HAVE_ATOMSPACE AND HAVE_GUILE)
+ENDIF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_BANK)
 
-IF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_LINK_GRAMMAR AND HAVE_UUID)
+IF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_LINK_GRAMMAR AND HAVE_UUID AND HAVE_BANK)
 	SET(HAVE_NLP 1)
-ENDIF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_LINK_GRAMMAR AND HAVE_UUID)
+ENDIF(HAVE_ATOMSPACE AND HAVE_GUILE AND HAVE_LINK_GRAMMAR AND HAVE_UUID AND HAVE_BANK)
```
linas: Most of NLP builds fine without attentionbank. Ghost however, needs openpsi. And I thought it was already handled separately and stubbed out. Hmmm. We need a `HAVE_GHOST` then, if that's not the case.

linas: #3618 should fix this.

PR #3622, opened 2019-11-27
Remove PLN

ngeiswei:
PLN has been moved here https://github.com/opencog/pln


ngeiswei:
Everything seems well, so I'm merging.

PR #3625, opened 2019-12-26
Update the API to the Pattern Matcher

linas:
See opencog/atomspace#2453 for the matching update


linas:
Merging, this cannot pass without pull req opencog/atomspace#2453

PR #3631, opened 2020-03-03
fix build issue for lojban code

AmeBel:


lib/lojban.conf
```diff
@@ -25,10 +25,10 @@ LOG_TO_STDOUT         = false
 
 # Cogserver in OSX will automatically change .so extension to .dylib
 # if .so exists.
-MODULES               = opencog/cogserver/server/libbuiltinreqs.so,
-                        opencog/cogserver/shell/libscheme-shell.so,
-                        opencog/cogserver/shell/libpy-shell.so,
-                        opencog/nlp/lojban/libLojbanModule.so
+MODULES               = opencog/modules/libbuiltinreqs.so,
+                        opencog/modules/libscheme-shell.so,
+                        opencog/modules/libpy-shell.so,
```
linas: fyi, these first three are auto-loaded and do not need to be specified.

PR #3634, opened 2020-04-27
Port to python version 3

linas:
according to  opencog/atomspace#2550


linas:
Merging. See #3636 -- simply changing the README causes the same collection of unit tests to fail.

PR #3635, opened 2020-04-27
Avoid parallel test.

linas:
This is a sanity check, to understand why circle-ci is failing
on the related pull req #3634 -- its a python change that seems
to make scheme code fail...


linas:
Heh. Unit tests fail anyway.

PR #3636, opened 2020-04-27
Change to README.

linas:
Check to see if the unit tests pass or fail ...


linas:
Merging. Simply changing the README causes unit tests to fail. Foo.

PR #3641, opened 2020-06-02
Look for boost in CMakeLists.txt

ngeiswei:
Otherwise the following error occurs:

```
[ 16%] Built target nlp_atom_types
[ 17%] Linking CXX shared library libnlp-types.so
/usr/bin/ld: cannot find -lBoost::filesystem
/usr/bin/ld: cannot find -lBoost::system
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,21 @@ ELSE (PLN_FOUND)
 ENDIF (PLN_FOUND)
 
 # ----------------------------------------------------------
+# Check for boost. We need dynamic-linked, threaded libs by default.
+SET(Boost_USE_STATIC_LIBS OFF)
+SET(Boost_USE_MULTITHREADED ON)
+SET(MIN_BOOST 1.46)
+
+# Required boost packages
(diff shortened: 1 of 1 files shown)


ngeiswei:
I find it odd that something as elementary as that was missing, it's somewhat suspicious. So, please, someone with more familiarity with the opencog repo, have a look at it. Thanks.

linas:
Yeah, that's in crazy-land. first of all, you can't build the atomspace python bindings without these two boost libraries, so they should already be installed when you built the atomspace.  .. unless you are not building the python bindings.

Next, the atom types should not depend on anything, excpt for the base atom types. ... but due to crazy circular dependencies, maybe something that needs boost got linked in.
```
ldd -r ./opencog/nlp/types/libnlp-types.so
```
shows all kinds of crap linked in. Sigh...

linas:
I'd rather use pull req #3642 to fix this.  Let me know if it works for you!? (it should; I can't see how it wouldn't..)

linas:
I'm going to reject this because I think its due to an unclean build. (and because #3642 moots such unclean builds.)

PR #3642, opened 2020-06-03
Use a minimalistic linkage.

linas:
Alternative fix to pull req #3641

### DIFF
--- a/opencog/nlp/types/CMakeLists.txt
+++ b/opencog/nlp/types/CMakeLists.txt
@@ -18,7 +18,7 @@ ADD_LIBRARY (nlp-types SHARED
 )
 
 TARGET_LINK_LIBRARIES(nlp-types
-	${ATOMSPACE_LIBRARIES}
+	${ATOMSPACE_atomtypes_LIBRARY}
 )
 
 # Without this, parallel make will race and crap up the generated files.


ngeiswei:
Unfortunately it doesn't completely fix it, now I'm having this error

```
[ 27%] Linking CXX shared library libwsd.so
/usr/bin/ld: cannot find -lBoost::filesystem
/usr/bin/ld: cannot find -lBoost::system
```

I naively tried to play with replacing `ATOMSPACE_LIBRARIES` by `ATOMSPACE_atomtypes_LIBRARY` here and there but I could fix it.

linas:
OK, so nothing in opencog uses boost ... you can verify that ` grep -r boost *` doesn't return anything.

So the dependency on boost is coming from something else, which was improperly linked.  You can find the culprit by saying: 
```
ldd -r /usr/local/lib/opencog/*.so |grep undefined
```
which sadly shows far too many. So, next we need;
```
ldd -r /usr/local/lib/opencog/*.so |grep undefined |grep boost
```
which tells me that `/usr/local/lib/opencog/libasmoses_exec.so` forgot to link to boost ... and also `libasfeature_selection.so` and others  .. so this is a bug in linking as-moses ...

ngeiswei:
Thanks for your explanation @linas, I'm not proud to say, but I'm still not completely comfortable with linking issues.

> which tells me that /usr/local/lib/opencog/libasmoses_exec.so forgot to link to boost ... and also libasfeature_selection.so and others .. so this is a bug in linking as-moses ...

OK, but opencog does not rely on asmoses so that can't explain this problem. Also, it's odd that I'm the only one having this problem.

I'm not gonna look any further into that issue because I basically don't use opencog.

If that's of any value, I noticed that if I remove

```
		${ATOMSPACE_smob_LIBRARY}
```

from

https://github.com/opencog/opencog/blob/master/opencog/nlp/wsd/CMakeLists.txt#L37

then the problem goes away.

linas:
> odd that I'm the only one

I think you have libraries that were built at different times, and are no longer compatible with one-another. 

> ${ATOMSPACE_smob_LIBR

Ah, interesting.  So saying this:
```
ldd -r /usr/local/lib/opencog/libsmob.so
```
does reveal that it is linking to `libboost_filesystem.so` .. but that is indirect, that is because smob is linking to the cython/python bindings.  Cython is the ONLY user of boost_filesystem in the atomspace.  So, I think what is happening for you is that once upon a time, long ago, you built the atomspace python bindings, and `make install` ed them. They are now sitting in ` /usr/local/lib/opencog/libPythonEval.so` and the date-stamp on that file is probably from years ago. You've since removed these boost components, but did not remove  `/usr/local/lib/opencog/libPythonEval.so` and so now things slide down-hill...

ngeiswei:
> I think you have libraries that were built at different times, and are no longer compatible with one-another.

Hmm, I don't think so, I'm purging opencog from my system almost every time I compile it, with the help of that script

https://github.com/ngeiswei/ocbld/blob/68fb99fc3e9541ec0ac2b5102da960f7bbb09ab0/ocbld#L104-L117

Maybe it's missing some places but I couldn't find anything suspect by grepping my `/usr` folder.

PR #3643, opened 2020-06-12
The pattern engine API has changed

linas:
It changes in opencog/atomspace#2685 and so this is needed.


linas:
Merging, because this blocks successful unit tests on opencog/atomspace#2685

PR #3645, opened 2020-06-18
Fix package dependencies

vsbogd:


linas:
FWIW, our debian packager is @mhatta but I'm guessing maybe he does not use this infrastructure?

vsbogd:
@mhatta, what manner of packaging do you use?

PR #3646, opened 2020-06-24
Fix restapi for new flask

manisabri:
flask.ext is depricated. A minor change to CMake file so it find boost libraries 
reg issue **https://github.com/opencog/external-tools/issues/130**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,11 +21,6 @@ ENDIF(CMAKE_VERSION VERSION_GREATER 3.0.2)
 
 PROJECT(opencog)
 
-FIND_PACKAGE(Boost 1.71 REQUIRED date_time program_options thread filesystem system unit_test_framework)
-IF(${Boost_FOUND})
-  INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
-ENDIF()
-
 # ----------------------------------------------------------
 # User-modifiable options. Feel free to change these!
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,11 @@ ENDIF(CMAKE_VERSION VERSION_GREATER 3.0.2)
 
 PROJECT(opencog)
 
+FIND_PACKAGE(Boost 1.71 REQUIRED date_time program_options thread filesystem system unit_test_framework)
+IF(${Boost_FOUND})
+  INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
+ENDIF()
+
 # ----------------------------------------------------------
 # User-modifiable options. Feel free to change these!
 #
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ ENDIF(CMAKE_VERSION VERSION_GREATER 3.0.2)
 
 PROJECT(opencog)
 
+FIND_PACKAGE(Boost 1.71 REQUIRED date_time program_options thread filesystem system unit_test_framework)
+IF(${Boost_FOUND})
+  INCLUDE_DIRECTORIES(${Boost_INCLUDE_DIR})
+ENDIF()
```
linas, line 27: Why did you make this change? It's causing unit tests to fail, and does not appear to be needed for anything...

linas:
I'll merge this if you undo the CMake issue .... then, shortly after merging, I will attempt to split out the the atomspace-publisher code, and it's unit tests into it's own git repo. Splitting it up is a part of a modularization effort described in #3391 

... if you have anything to add or say, or would want to do with moving the code, let me know ...

CMakeLists.txt
manisabri: I couldn't compile without this On Ubuntu 20.04. I can add it manually later for my selft later

manisabri:
sounds good, I can do the atomspace-publisher split and make it a proper python package that works out of the box after build

linas:
I'll do the split, and then hand it over to you for cleanup?

manisabri:
that's great

linas:
The module was misplaced, I will move it over as well, it will take about an hour...

linas:
Hmmm. the only technical reason for it to be a cogserver module is because we did not yet have python and scheme modules at the time .... it might be worth converting from cogserver to either/both of those two...

manisabri:
On the other hand they all are providing services over network so they might belong to each other

manisabri:
Have you considered having a single AtomSpace with python, guile and network bindings in a way that when you can manage a single AtomSpace with all of those but the core is still C++? 
As far as I can see guile, python and cogserver create their own AtomSpaces with no connection between them

linas:
OK, the new code is in https://github.com/opencog/atomspace-restful -- hack on it any way you wish...

> As far as I can see guile, python and cogserver create their own AtomSpaces with no connection between them

Not true, they are supposed to actually all share the same atomspace. If they don't, that is a bug, but there are actually unit tests  to make sure that they do. I'm pretty sure it works correctly, people use his all the time  ... 

There might be a bug with RESful, though ...

linas:
The code for where they all share the same atomspace is tucked away in the initialization routines for each of these. The python code passes around a raw C++ pointer to the atomspace; the scheme code just defines a cogserver module so that you say `(use-modules (opencog))` and that starts the atommspace and `(use-modules (opencog cogserver))` and that starts the cogserver using the existing atomspace.

For restful, it would make sense to have `(use-modules (opencog restful))` as a guile module, and `import opencog.restful` for a python module...

PR #3648, opened 2020-06-25
Remove archaic, unused python code.

linas:
It has been moved to it's own git repo, here:
https://github.com/opencog/python-attic/

This is a part of the ongoing work for issue #3391


linas:
I'm just going to merge this because I assume that no one will object. This code has not been touched in 5 or 8 years, now. The new repo keeps the code available, but it is no longer here, cluttering things up and confusing the newcomers.

PR #3654, opened 2020-08-04
Update scheme openpsi examples

AmeBel:


linas:
You deleted `basic,scm` -- does this example not work any more?

the ping-pong example seems much too complicated -- a simpler, more basic example is needed ...

AmeBel:
Yes the `basic.scm` doesn't work.
I will take the ping-rules and create a `ping.scm`

linas:
Thanks. Anyway, I'm not sure, but if you are waiting for me to merge, don't; this is your stuff ... merge whenever...

Issue #3662, opened 2020-10-13
LgConnector is always interpreted as a node in Guile


leungmanhin:
As seen in Guile shell:

```Scheme
scheme@(guile-user)> (use-modules (opencog) (opencog nlp))
scheme@(guile-user)> (LgConnector)
$1 = (LgConnectorNode)

scheme@(guile-user)> (LgConnector (LgConnectorNode "Wa") (LgConnDirNode "-"))
$2 = (LgConnectorNode
  (LgConnectorNode "Wa")
  (LgConnDirNode "-"))

scheme@(guile-user)> (cog-get-atoms 'LgConnector)
$3 = ((LgConnectorNode
  (LgConnectorNode "Wa")
  (LgConnDirNode "-"))
 (LgConnectorNode)
 (LgConnectorNode "Wa")
)
scheme@(guile-user)> (cog-get-atoms 'LgConnectorNode)
$4 = ((LgConnectorNode
  (LgConnectorNode "Wa")
  (LgConnDirNode "-"))
 (LgConnectorNode)
 (LgConnectorNode "Wa")
)
```

linas:
Whoops. There should not be both a `LgConnectorNode` and an `LgConnectorLink` ...

linas:
This appears to be a very old bug (at least 6 years old!?). I am changing every occurence of `LgConnectorNode` to `LgConnNode`, so that it matches the naming scheme for all the other `LgConn` types.

linas:
fixed in #3669

leungmanhin:
Thanks @linas

linas:
BTW, since you seem to be paying attention, I may as well say it here: there is a major change to how opencog is built: you now need to git clone, build and install https://githib.com/opencog/lg-atomese first. It splits out the link-grammar part.  Let others know, if they care.

All that is left in the main opencog repo are the chatbots, and pretty much nothing else.  Other than that, everything should work (or not work, as the case may be) just as before.

PR #3669, opened 2021-07-06
resolve LgConnectorNode naming conflict; issue #3662

linas:
Appearantly, there was a long-standing atom naming conflict,
where there was both an LgConnectorLink and a LgConnectorNode.
Appearently, this used to work ... and recently broke, both for
unknown reasons. Rather than tracking down the breakage, it seems
like the correct solution is to fix the actual naming conflict.
    
This is especially the case, as this code seems mostly unused, and
therefore unlikely to break anyone.


linas:
This is failing in relex. I've tweaked relex with a matching update, but I guess that circleci will continue to fail until it pulls the latest...

PR #3670, opened 2021-07-07
Move link-grammar wrapper code into it's own git repo

linas:
Part of the work for #3391 -- I was finding that I was using link-grammar a lot, and needed it to be independent of the rest of the software stack.


linas:
I'm gong to merge this immediately, because:
1) The unit tests pass! Horray!
2) I don't think anyone is using this code, anyway, so no one will be impacted.

PR #3673, opened 2021-09-29
Update bldtargets.dox

Sahilbawa777:
sir i love your project its like the best I have ever seen as a programmer you are great and the best thing is the that you made it as open source program. So i appreciate you brother for this job


linas:
Thank you!  However, there are some problems with this. The `make` command must be spelled lower-case; as there does not exist any `Make` command. Likewise, the targets are lower-case, and cannot be capitalized.  Therefore, I am going to close this pull req without merging it.

PR #3679, opened 2022-06-16
Disable two of three ghost unit tests.

linas:
These test fail, and I can't figure out why. So just disable them.

@AmeBel  - fyi.

### DIFF
--- a/tests/ghost/CMakeLists.txt
+++ b/tests/ghost/CMakeLists.txt
@@ -20,13 +20,14 @@ IF(HAVE_GUILE)
 ENDIF()
 
 IF("${RELEX_REACHABLE}" STREQUAL "#t")
-  ADD_GUILE_TEST(GhostSyntaxUTest test-ghost-syntax.scm)
-  SET_PROPERTY(TEST GhostSyntaxUTest
-        APPEND PROPERTY ENVIRONMENT
-        "PYTHONPATH=${PYTHON_ROOT}:${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_SOURCE_DIR}/tests/cython/agents:${PYTHON_ROOT}"
-  )
+# These tests are failing and I don't know why.
+#  ADD_GUILE_TEST(GhostSyntaxUTest test-ghost-syntax.scm)
+#  SET_PROPERTY(TEST GhostSyntaxUTest
+#        APPEND PROPERTY ENVIRONMENT
+#        "PYTHONPATH=${PYTHON_ROOT}:${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_SOURCE_DIR}/tests/cython/agents:${PYTHON_ROOT}"
+#  )
   ADD_GUILE_TEST(GhostProcedureUTest test-ghost-procedures.scm)
-  ADD_GUILE_TEST(GhostUTest test-ghost.scm)
+#  ADD_GUILE_TEST(GhostUTest test-ghost.scm)
   
 ELSEIF("${RELEX_REACHABLE}" STREQUAL "config-error")
   message(FATAL_ERROR "Issue with relex utilities or the file ${RELEX_UTILS}")
--- a/tests/ghost/CMakeLists.txt
+++ b/tests/ghost/CMakeLists.txt
@@ -26,7 +26,7 @@ IF("${RELEX_REACHABLE}" STREQUAL "#t")
 #        APPEND PROPERTY ENVIRONMENT
 #        "PYTHONPATH=${PYTHON_ROOT}:${PROJECT_BINARY_DIR}/opencog/cython:${PROJECT_SOURCE_DIR}/tests/cython/agents:${PYTHON_ROOT}"
 #  )
-  ADD_GUILE_TEST(GhostProcedureUTest test-ghost-procedures.scm)
+#  ADD_GUILE_TEST(GhostProcedureUTest test-ghost-procedures.scm)
 #  ADD_GUILE_TEST(GhostUTest test-ghost.scm)
   
 ELSEIF("${RELEX_REACHABLE}" STREQUAL "config-error")


AmeBel:
ok

linas:
@AmeBel I assume it's an "easy" fix; Almost all of the error messages say "Did you forget to link a goal to the rule?" -- I'm 90% sure that its some file-loading issue.

Does anything anywhere use ghost for anything anymore, or is this all  dead project?  More generally, is the entire repo a dead project? Is there any scenario in which anyone will ever want to revive anything in this repo?

linas:
The last questions would be aimed at @ngeiswei @leungmanhin @bgoertzel @vsbogd -- repeating the question: Is there any scenario where anything at all in the opencog repo would ever be revived again, used in some way? Or is this particular repo dead?

This question can be expanded to the attention (ecan) repo and the spacetime repo.  Both of those also seem dead, with no forward path that I can see.

AmeBel:
On ghost, it isn't being used for anything at the moment. With regards to the whole repo, I don't have a full view of what everyone else is doing, so can't say if it is dead or not.

ngeiswei:
> I believe this is a deep and fundamental mistake.

I was agnostic about that till I got my hands on MeTTa, and since then I don't really want to go back to good old school Atomese.  For me it comes down to flexibility.  MeTTa is a non deterministic functional programming language, unifying reasoner and interpreter, all that completely user definable.  The best way to demonstrate its power is via examples, you can find some there https://github.com/trueagi-io/hyperon-experimental/tree/main/python/tests/scripts.  On my side, I've also started to port PLN to MeTTa, it's in a pre-embryonic stage for now but it should grow over the coming weeks https://github.com/trueagi-io/hyperon-pln

I still think OpenCog Classic as a lot of potential but I perceive it as more rigid and thus less nice to work with as MeTTa (even at this early stage, MeTTa still has a long way to go, especially in terms of tooling, library, dsl, etc).

There's a lot more to say, but amongst the things I like about it, is that it has the potential to become a general purpose programming language, especially suited for AI, but not only.

ngeiswei:
This presentation of Alexey gives a good idea of what MeTTa can currently do https://singularitynet.zoom.us/rec/play/oXVRUQnooDng5TcsVBrjO03dlDPjvi9LIbM4vRqqloodrU5I83S4p9JWpvwoDQJGO35JLW1Km_VP8e2g.gJsEKZiwoHpZZWyO?startTime=1650636238000&_x_zm_rtaid=LIXwnLEfTsyYwDzVhciE8w.1653485694598.b4a3ff6db2468bd91e3804c4c08a580f&_x_zm_rhtaid=361

Another feature that's not mentioned in the video (it hasn't been implemented yet in that prototype) is about interpretation/inference control (how to manipulate the interpreter stack within MeTTa, etc).  It's another part that makes MeTTa very powerful, as it has the potential to unify inference control, attentional focus, etc, in one elegant system.

linas:
Hi @ngeiswei Yeah, my apologies for being so sharp-tongued.  I'm just upset & panicking, because activity on opencog has dropped to zero, and I have oceans of code that would be impossible to port to MeTTa.  Yes, I understand that sometimes trying something completely different is a good thing. So .. sure.  I don't understand why there wasn't any effort to .. keep things more joined up.

Perhaps it's a natural divergence. Everything I do is about data structures, and extracting relationships.  I don't need a unifier, or a reasoner or a "programming language", or inference control or attentional focus. I understand why those are interesting, but .. nothing that I work on requires any of that kind of stuff. So .. well, your building something different, I guess.

Which leads to the next question: should the git repos for PLN, URE, and attention the pattern miner and spacetime be be mothballed? I keep having to make changes inside of them, (why, just today!!) to make the code compile and unit tests pass, and its a waste of time if no one is interested in that stuff. I'd like to shed that baggage, and just stick to the core components.  Would make my life easier...

This would be a question for @bgoertzel and .. I'm not sure who else cares, at this point.  

Can these unmaintained components be mothballed?

ngeiswei:
I completely understand your sentiment, @linas.

Regarding mothballing URE, PLN and a few others.  I feel it's probably premature, on the other hand, I don't want to artificially increase your workload.  Can we revisit that decision in say a month?  The process should involve creating tags and updating READMEs, mostly work on my part I suppose.  Then I would probably prefer to treat them as dormant with a possibility to be re-activated, till they eventually get archived for good.

Also, something that we should be open to are the possibilities of interoperability between Classic and Hyperon.  I don't have a clear idea about that, or a feel of whether that would worth the effort, but I suppose it will get clearer as more applications run on Hyperon.

linas:
A decision can be put off for years. I'm overstating the case: the atomspace API's change only very rarely, and when they do change, keeping everything else in sync is usually quite easy, if a bit tedious and annoying. The part that is hard to deal with is when there are opaque, hard-to-debug failures. Those are painful. That's the pain I want to avoid.

Re: Classic & Hyperon -- given what you've written, it appears that paths are rapidly diverging. The stuff that you're mentioning seems to have little or no relevance to what I'm working on, and the opposite also appears to be true.  So the question is: will there ever be Hyperon users that will want to do things that the AtomSpae is good at, and vice-versa -- will AtomSpace users (namely, me) ever need anything that Hyperon provides?  This is the question to explore. 

Let me repeat what I'm doing, and what I need:
* I'm collecting statistics in large databases. My runs take CPU-months of compute.
* The above needs a very full and complete set of statistics and statistical analysis tools (which I've got, thanks to the "matrix" API)  This seems to be something that Hyperon is not interested in ... 
* The databases are larger than what fits in RAM, so I need to have selective access. I've got that with the `StorageNode` API. This seems to be something Hyperon is not interested in.
* For practical reasons, I need network-distributed computing. I've got that with the `CogStorageNode`, and I recently polished up a write-through proxy, useful for doing scatter-gather, map-reduce type computing over a cluster. Again, this is something Hyperon seems disinterested in.
* AtomSpace frames aka views, slices, versions, change-sets, snapshots: as the result of computations flow in, I need the ability to treat them as small deltas or changes to the atomspace contents.  This is so that changesets can be kept or abandoned, expanded or collapsed.  Vaguely like your BIT in URE, but in a general setting, unrelated to any particular rules, unrelated to inference. Again, this is not a concept that appears in Hyperon.
* Generic Values: when I count things, I need to keep different counts in different places. I need the statistics system to work with all of these, so that I can compute correlations, averages, gradients, marginals, dot-products, matrix products, whatever.  Again, hyperon doesn't seem to have any infrastructure for this ...
* System management. I *don't* have this and its a major pain in the neck. Managing the big compute jobs by hand sucks. It's like doing laundry for weeks, where you have to check on the machine every few hours or every day. Ugh. Like watching paint dry. Boring AF. I really really really needs a job management solution...
* Parsing and generation.  This is the jigsaw-piece idea. As you're aware, both backwards and forwards chaining is like the problem of assembling jigsaw pieces into a completed "picture". In the URE, the "jigsaw pieces" were called "rules".   Rather than using the URE or inventing some new kind of URE to perform assembly, I'm using link-grammar for assembling together deductive chains of inferences. Mostly because its really fast and much more general than the URE (and LG has built-in inference control.)  This is something that Hyperon seems to be all about, but ... we seem to be very disconnected as the the theory and ideas of how this would work.

So, for interop with Hyperon, it seems like the last bit is the most promising.  If I give Hyperon a bunch of jigsaw pieces, can it assemble them?   How fast? What is the format for specifying the jigsaw pieces? As you noticed with the URE and the PLN rules, they were insanely over-complicated and verbose.  From what I can tell, with Hyperon, you are trying to figure out how to create much simpler, much less verbose "rules" (or whatever you call them). I call them jigsaw pieces, and they are as minimal as I can imagine them being.

Is there a way to collaborate with Hyperon to have a common jigsaw-piece format? Something that you can use  for inference, and something I can use for generation? 

Again, from where I sit: it seems like parsing, inference and generation are all ***exactly the same thing***. I don't know if you share this idea. I do know you are focused on inference. Can your concept of inference be broadened enought to also do parsing and generation, in general? (I think it can, but its up to you to see this and do this) If you do this, can we agree on some generic format or infrastructure?  If I gave yu a collection of LG jigsaw pieces, would Hyperon be able to assemble them? If not, why not?

Tag @bgoertzel and @vsbogd

ngeiswei:
> If I give Hyperon a bunch of jigsaw pieces, can it assemble them?

Yes, in theory.  In practice it depends on the rapidly evolving state of the code.

> How fast?

I don't know, not yet.

> What is the format for specifying the jigsaw pieces?

It's highly user programmable, but I guess it could look like `(= conclusion premise_1 ... premise_n)`.

>  Can your concept of inference be broadened enough to also do parsing and generation, in general?

I believe so.

Regarding deprecating URE, PLN, miner, spacetime, attention, I would prefer to finalize that when I have Hyperon counterparts.  On the other hand, such delay shouldn't incur work on your part.  So, soon, I could move the current versions of cogutils, atomspace, ure, pln, etc, to singnet, so that I have a state of OpenCog Classic that is readily available and working.  Then update some READMEs and tags on opencog so they can safely be mothballed.  It's not a lot of work, but since it is not high priority either, I just want to give myself a few weeks of margin to fine the best time to work on it.  Does that sound OK with you?

linas:
> Does that sound OK with you?

I can wait months, or a year or more. Just right now, nothing is getting in the way.

> `(= conclusion premise_1 ... premise_n)`

That's a lambda, not a jigsaw. Lambda's are a special-case of jigsaws.  Lambdas have inputs ('premise`) and a singular output (`conclusion`).  Lambdas are always heterosexual (outputs can only plug into premises; you can never join together two premises, or attach a conclusion to another conclusion.)  Jigsaws have mating rules -- they can be mono-sexual, trisexual, whatever, arbitrarily specified. Unlike lambdas, jigsaws do not have a directional flow, unless the mating rules impose a directional flow. Jigsaws have associated statistical information: certain attachments are preferred over other attachments, in a probabilistic fashion.

linas:
To illustrate the jigsaw idea in greater detail:

For example, a mono-sexual typed jigsaw looks like this:
```
(connector_1, connector_2, ..., connector_n)
```
Each connector is of type `A, B, C, ....` and the connection rules are that any connector of type `A` can attach to any other connector of type `A`, etc. 

A heterosexual connector has types `A+, A-, B+, B-, ...` with the rule that `A+` can only attach to `A-` and never to anything else.  Link-grammar is hetero-sexual (mostly, its actually more complicated... see below)

Trisexual connectors can be designated with the connector types `Aa, Ab, Ac, Ba, Bb, ...` with the rule that `Aa` can only attach to `Ab` or to `Ac` but never to `Aa`, or to anything else.

A different kind of trisexual arises when allowing wild-card types. Then, for example, the rule can be that `Aa` can only attach to `Aa` (and never `Ab` or anything else), unless there is a wildcard: so that `A*` can attach to any kind of `Ax` for any `x`.

LG connectors are of the form  `dCAPS*****...+/-`  where `d` is a dependency marker (like a lambda input-output marker, or wildcard)  and uppercase `CAPS` is the principal type (must always mate exactly), the `****...` are an infinite string of position-dependent N-sexual connectors (or wildcards in any position) and `+/-` is another hetero-sexual marker (denoting left/right)

For video, sexuality includes left/right, above/below, infront/behind, acute/obtuse angle, so that the "grammar" of the 2D image or 3D object has a positional dependence. The 2D-image or 3D-object analog of @akolonin 's "transition freedom" is just the statement that certain connectors in certain directions are very loose or weak: e.g. the connectors to the background of the image, where the background can be "anything".  This is just the image-segmentation analog of the text segmentation problem. 

To recap: there are typed connectors to connect "nearby" things; they have complex mating rules, they have numerical weights or numerical binding strengths to represent the affinity or preferential attachment of any given connection. 

The above is a description of the de-facto framework I'm working in.

ngeiswei:
Thanks for the description, @linas, I think I understand better what you've meant for a while now, and I like the various "sexual" modes it enables.  So it seems you're using LG as a reasoning engine, right?  But then my question is, how expressive is LG?  Essentially, is it Turing Complete?  Quickly going over the LG documentation, I stumbled upon "Link grammar appears to be isomorphic to categorial grammars", and a quick look at https://en.wikipedia.org/wiki/Categorial_grammar#Relation_to_context-free_grammars seems to suggest that it is equivalent to a context free grammar, thus not Turing Complete.  If that's the case I worry that this limitation is too strong for the kind of reasoning we need.

For instance the URE (and MeTTa, at least eventually), in spite of its hetero-sexuality, is Turing Complete (thanks to its syntactic unification engine).  Anyway, I should probably carefully read the categorial grammar page first, but that is my initial impression.

linas:
Hi @ngeiswei Some short answers:

> you're using LG as a reasoning engine, right?

No. I am using LG as just one small part of a general research program into AGI.  The jigsaw concept is more general and deeper than LG, it is a generalization. LG is a theory of human natural language; it is not a reasoning engine.  Jigsaws are the basic components out of which  "sheaves" are made. https://en.wikipedia.org/wiki/Sheaf_(mathematics)

> Turing complete.

Every Turing complete programming language can be represented as a tree, the abstract syntax tree (AST). https://en.wikipedia.org/wiki/Abstract_syntax_tree  Every AST is a kind of graph. (a kind of directed, acyclic graph, DAG) https://en.wikipedia.org/wiki/Directed_acyclic_graph  Every graph can be composed out of jigsaws with sexual connectors (sorry, there is no single well-written wikipedia article on that, but it is "well-known".)

One can go in the opposite direction: jigaws can be used to create sheaves, some of which (but not all of which) are graphs. Some graphs, but not all, are DAG's. Some DAG's but not all are AST's. Some AST's, but not all, represent computations. Some computations, but not all are Turing complete.  

In short, yes, jigsaws can represent Turing-complete languages. The Turing-complete languages are a narrow special case out of a much much broader class of things that are computationally interesting.

linas:
At any rate, I see that unification is a fundamental concern to you, and I will try to take some time in the coming months to put together a cleaner story regarding unification. For my work, unification has so far been irrelevant, so I've been ignoring it.

linas:
@ngeiswei My prior explanation was too long. The short version is: any lambda expression is a special cse of a jigsaw. Any URE rule is a special case of a jigsaw. Roughly speaking, the act of joining together connectors is unification.

ngeiswei:
I can see that, I like the non-straight sexuality that jigsaw puzzles generally offer.

